PR #39, opened 2021-08-21
align arrange with `Ecto.Query.order_by/3`

cigrainger:
This changes `DataFrame.arrange/2` to match the `Ecto.Query.order_by/3` API as suggested by @josevalim.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -567,7 +567,7 @@ defmodule Explorer.DataFrame do
     A single column name will sort ascending by that column:
 
       iex> df = Explorer.DataFrame.from_map(%{a: ["b", "c", "a"], b: [1, 2, 3]})
-      iex> Explorer.DataFrame.arrange(df, ["a"])
+      iex> Explorer.DataFrame.arrange(df, "a")
       #Explorer.DataFrame<
         [rows: 3, columns: 2]
         a string ["a", "b", "c"]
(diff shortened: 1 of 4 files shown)


lib/explorer/data_frame.ex
```diff
@@ -631,32 +605,33 @@ defmodule Explorer.DataFrame do
   @spec arrange(
           df :: DataFrame.t(),
           columns ::
-            [String.t() | {String.t(), :asc | :desc}]
-            | Keyword.t()
+            String.t() | [String.t() | {:asc | :desc, String.t()}] | Keyword.t()
```
polvalente, line 608:
```suggestion
            String.t() | [String.t() | {:asc | :desc, String.t()}]
```

You don't need the keyword() spec due to  it being equivalent to `[{atom(), term()}]`. Also, only strings are valid as values in this context

polvalente: Also, I think that Keyword.t() doesn't exist

polvalente (approved):
Just some minor comments. If you make the suggested changes I can approve it again :)

lib/explorer/data_frame.ex
cigrainger: Thanks! `Keyword.t()` does exist: https://hexdocs.pm/elixir/1.12/Keyword.html#t:t/0. But you're right it's not necessary anymore!

cigrainger: Only string should be valid, so that's good.

polvalente: Nice! I though only keyword() was valid :)

polvalente (approved):

PR #47, opened 2021-09-05
fix the introduction notebook

michalslaski:
Update the introduction notebook to correctly use the `DataFrame.arrange` function.


cigrainger:
Looks like I missed a spot. Thank you!

PR #50, opened 2021-09-06
Changes when CI runs

sobolevn:
Right now it was only working for pull requests, but this is not exactly right:
1. It can happen that on merge our CI will fail, even the PR was fine. It happens when PR is stale for some time and is not properly rebased. New code can be pushed in the meantime
2. When you push directly to master
3. Without `workflow_dispatch` you cannot manually trigger the CI, which is sometimes super helpful

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,9 @@
-on: pull_request
+on:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


cigrainger:
Getting this right was on my todo list. This looks right to me. :) Thanks!

PR #57, opened 2021-09-08
Add and/or operations for series

### LINKED ISSUE #56: Clarification / Question: Dataframe.filter AND/OR mask?
I see in the docs that one can filter using a mask:

```elixir
df = Explorer.DataFrame.from_map(%{a: ["a", "b", "c"], b: [1, 2, 3]})
Explorer.DataFrame.filter(df, Explorer.Series.greater(df["b"], 1))

#Explorer.DataFrame<
[rows: 2, columns: 2]
a string ["b", "c"]
b integer [2, 3]
>
```

How does one create an AND/OR mask? I could compute them manually and use a list, but is there a better way?
Sort of what I'm looking for:

```elixir
df = Explorer.DataFrame.from_map(%{a: ["a", "b", "c"], b: [1, 2, 3]})
mask = Explorer.Series.or(Explorer.Series.equal(df["b"], 1), Explorer.Series.equal(df["b"], 3))

Explorer.DataFrame.filter(df, mask)

#Explorer.DataFrame<
[rows: 2, columns: 2]
a string ["b", "c"]
b integer [1, 3]
>
```

drowzy:
Adds `Explorer.Series.and_/2` & `Explorer.Series.or_/2` by using `bitor/and` on the series.

closes #56

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -20,6 +20,7 @@ defmodule Explorer.Series do
   """
 
   alias __MODULE__, as: Series
+  alias Kernel, as: K
 
   import Explorer.Shared, only: [impl!: 1]
   import Nx.Defn.Kernel, only: [keyword!: 2]
@@ -790,14 +791,14 @@ defmodule Explorer.Series do
   """
   @spec add(left :: Series.t(), right :: Series.t() | number()) :: Series.t()
   def add(%Series{dtype: left_dtype} = left, %Series{dtype: right_dtype} = right)
-      when Kernel.and(left_dtype in [:integer, :float], right_dtype in [:integer, :float]),
+      when K.and(left_dtype in [:integer, :float], right_dtype in [:integer, :float]),
       do: apply_impl(left, :add, [right])
 
   def add(%Series{dtype: left_dtype}, %Series{dtype: right_dtype}),
     do: dtype_mismatch_error("add/2", left_dtype, right_dtype)
 
   def add(%Series{dtype: dtype} = left, right)
-      when Kernel.and(dtype in [:integer, :float], is_number(right)),
+      when K.and(dtype in [:integer, :float], is_number(right)),
       do: apply_impl(left, :add, [right])
 
   def add(%Series{dtype: dtype}, _), do: dtype_error("add/2", dtype, [:integer, :float])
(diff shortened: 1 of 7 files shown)


josevalim:
Btw, Elixir will be fine with invoking `Explorer.Series.and(foo, bar)`. So we can keep the name as `and`/`or`, but you will need to declare it as `def left and right do...end`. :)

drowzy:
> Btw, Elixir will be fine with invoking `Explorer.Series.and(foo, bar)`. So we can keep the name as `and`/`or`, but you will need to declare it as `def left and right do...end`. :)

Hah! Did not know that! Thank you :)!

Though I'm running into to trouble on `Kernel` being [imported](https://github.com/elixir-nx/explorer/blob/main/lib/explorer/series.ex#L26) and `length` being excluded to avoid the conflict with the locally defined `length/1`. However now I'm getting a conflict on `and` instead. 
Since `and` is used in guards so excluding it fails to compile. Any suggestions how to resolve that?

cigrainger:
Thanks for this! Re: the kernel conflicts, one way I would approach it (and I'm not sure if it's the right way!) is to pattern match and then `defdelegate` after the exhausted pattern matches. So with `length` as an example:

```elixir
  @spec length(series :: Series.t()) :: integer()
  def length(%Series{} = series), do: apply_impl(series, :length)
  defdelegate length(other), to: Kernel
```

josevalim:
@cigrainger unfortunately that won't work for and/or if used in guards because guards require specific constructs.

@drowzy I would try this at the top of the file:

    import Kernel, except: [and: 2, or: 2]
    alias Kernel, as: K

Now every time there is a `and`/`or` from Kernel, you rewrite `left and right` to `K.and(left, right)`. It is up to @cigrainger though to say if he wants to leave with this.

However, I would use `def left and right do` only on the API definition. In the backend, I would call it `binary_and` and `binary_or`, so we don't have to handle the ambiguity in the backend itself (only on the API facade). WDYT?

drowzy:
@josevalim Thanks for the advice :)! The API definition is update to use `and`/`or` and the behaviour and backend uses `binary_and`/`binary_or`.

lib/explorer/series.ex
```diff
@@ -99,9 +100,7 @@ defmodule Explorer.Series do
         [1.0, nil, 2.5, 3.1]
       >
 
-    Mixing data types will raise an ArgumentError.
-
-      iex> Explorer.Series.from_list([1, 2.9])
+  Mixing data types will raise an ArgumentError. iex> Explorer.Series.from_list([1, 2.9])
```
cigrainger, line 103:
```suggestion
  Mixing data types will raise an ArgumentError. 
  
    iex> Explorer.Series.from_list([1, 2.9])
```

cigrainger (commented):
Just one thing to fix. The readability of relatively complex guards is definitely taking a hit, but I think it's worth it for the functionality.

PR #60, opened 2021-09-20
add nil check masks to series

cigrainger:
See: tin. I wasn't sure about naming on this one. But basically it adds the ability to mask for nil/not-nil values. Very useful when doing something like:
```elixir
iex> [1, 2, nil, 3] |> Series.from_list() |> Series.filter(&Series.not_nil?/1)
#Explorer.Series<
  integer[3]
  [1, 2, 3]
>
```


cigrainger:
@josevalim just a ping to get your opinion on naming here.

josevalim:
:+1: for the name, especially as those are not guards.

josevalim (approved):

Issue #58, opened 2021-09-13
Settle on strings vs. atoms for column names


cigrainger:
> One more thing that came up, most noticeable in the dtypes and names test is the inconsistency of referring to columns sometimes as atoms and sometimes as strings. with_columns and names both takes the column names as strings, while dtypes expects a keyword list where they're atoms. Should I add another commit to settle on one or the other? I could also make either work, but that gives me flashbacks to Rails where controller params keys could be either strings or atoms, which caused a lot of confusion, and which Phoenix rightly (IMO) fixed by standardizing on just strings.

On this, I think the entire API is a bit split on this. I've tried to make both work in many cases where I felt it was more ergonomic, but I'm feeling the same as you that they should be standardised. I think the obvious way to do that is to settle on strings. Let's leave this the way it is right now and we can revisit this in a broader context.

_Originally posted by @cigrainger in https://github.com/elixir-nx/explorer/issues/48#issuecomment-917788882_

josevalim:
They definitely have to be strings because I can see people using dataframes to work on external data and you don't want to convert to atoms. On the other hand, for things like dtypes, having them as atoms is definitely more convenient. So I would go with supporting both but normalizing all of them to strings on the facade API instead of the backend, as this is shared logic across all backends and all shared logic goes in the facade.

cigrainger:
This is pretty much the approach I've taken for column names -- support both but normalize on the facade API. There are a couple places where I think strings aren't supported and I'll try to root those out (the dtypes/schema approach above definitely is one of those). @losvedir what do you think about everything accepting strings for column names but some places _also_ accepting atoms where it may be more convenient or readable (e.g. keyword lists as arguments)?

losvedir:
> @losvedir what do you think about everything accepting strings for column names but some places also accepting atoms where it may be more convenient or readable (e.g. keyword lists as arguments)?

Sounds good to me! As long as the whole interface accepts one or the other then I'm happy. And it makes sense to me for that to be strings not only because that works more easily with different backends per what José said, but also because that's what the library drives you towards in the core interaction with DataFrames: you do `df["my_col"]` and not `df[:my_col]` or `df.my_col`. From what I've come across I've only run into `dtypes` not supporting strings, and that's an easy fix.

[...]

PR #62, opened 2021-09-21
use strings (not atoms) for column names in dtype specification

cigrainger:
See #58

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -54,7 +54,7 @@ defmodule Explorer.DataFrame do
   ## Options
 
     * `delimiter` - A single character used to separate fields within a record. (default: `","`)
-    * `dtypes` - A keyword list of `[column_name: dtype]`. If `nil`, dtypes are imputed from the first 1000 rows. (default: `nil`)
+    * `dtypes` - A list of `{"column_name", dtype}` tuples. Uses column names as read, not as defined in options. If `nil`, dtypes are imputed from the first 1000 rows. (default: `nil`)
     * `header?` - Does the file have a header of column names as the first row or not? (default: `true`)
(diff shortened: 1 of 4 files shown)


lib/explorer/data_frame.ex
```diff
@@ -54,7 +54,7 @@ defmodule Explorer.DataFrame do
   ## Options
 
     * `delimiter` - A single character used to separate fields within a record. (default: `","`)
-    * `dtypes` - A keyword list of `[column_name: dtype]`. If `nil`, dtypes are imputed from the first 1000 rows. (default: `nil`)
+    * `dtypes` - A list of `{"column_name", dtype}` tuples. Uses column names as read, not as defined in options. If `nil`, dtypes are imputed from the first 1000 rows. (default: `nil`)
```
josevalim, line 57: Just to throw out there, cause I am not sure it is a good idea, you could also make this be: `dtypes: [string: "foo", string: "bar", integer: "baz"]`.

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -33,7 +33,7 @@ defmodule Explorer.PolarsBackend.DataFrame do
     dtypes =
       if dtypes do
         Enum.map(dtypes, fn {colname, dtype} ->
-          {Atom.to_string(colname), Shared.internal_from_dtype(dtype)}
+          {colname, Shared.internal_from_dtype(dtype)}
```
josevalim, line 36: It is probably worth adding some validation here (check the types and raise invalid dtype otherwise).

lib/explorer/data_frame.ex
dustinfarris: I personally like the 2-tuples better

PR #73, opened 2021-10-12
add shell.nix

cigrainger:
I'm using NixOS full time at the minute. The changes to `.gitignore` are necessary. The `shell.nix` is not but I figured I'd put it up in case there are any other Nix-curious folks like me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,8 @@ explorer-*.tar
 
 # For the Mac users
 .DS_Store
+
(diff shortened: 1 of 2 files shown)


josevalim:
Theoretically, the Nix files and .DS_Store for Mac (as well as any other OS or editor specific config) should go to your global gitignore, this way you don't to ignore them per project: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c :)

Issue #70, opened 2021-10-08
Improve type casting when creating a Series or DataFrame


dgreiss:
These 2 examples currently throw an error:

- `Explorer.Series.from_list([1, 2.0])`
- `Explorer.Series.from_list([nil, nil])`

Would it be possible for a mix of integers and floats down cast to floats, similar to how it is done when adding a two series: 
```elixir
s = Series.from_list([1, 2, 3])
s1 = Series.from_list([1.0, 2.0, 3.0])
Series.add(s, s1)
```
For the list of `nil`s, would it be possible to be casted to a default type. In R, a list of `NA` is casted to a boolean type `tibble::tibble(a = c(NA))`

Thanks!

dgreiss:
Closed by #77

PR #77, opened 2021-10-30
Type casting mixed numeric types to float

dgreiss:
First attempt at solving #70

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -109,6 +109,7 @@ defmodule Explorer.Series do
   def from_list(list, opts \\ []) do
     backend = backend_from_options!(opts)
     type = check_types(list)
+    {list, type} = cast_numerics(list, type)
     backend.from_list(list, type)
   end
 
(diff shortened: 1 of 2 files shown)


lib/explorer/series.ex
```diff
@@ -100,15 +100,24 @@ defmodule Explorer.Series do
...
 
-    iex> Explorer.Series.from_list([1, 2.9])
-    ** (ArgumentError) cannot make a series from mismatched types: type of 2.9 does not match inferred dtype integer
+   iex> Explorer.Series.from_list([1, 2.0])
+   #Explorer.Series<
+     float[2]
+     [1.0, 2.0]
+   >
+
+  Mixing data non-numeric types will raise an ArgumentError.
```
cigrainger, line 111:
```suggestion
  Mixing non-numeric data types will raise an ArgumentError.
```

lib/explorer/series.ex
```diff
@@ -1457,23 +1466,45 @@ defmodule Explorer.Series do
...
   end
 
+  defp cast_numerics(list, type) when type == :numeric do
+    data =
+      Enum.map(list, fn item ->
+        case item do
+          nil -> nil
+          _ -> item / 1
+        end
+      end)
```
cigrainger, line 1501:
```suggestion
      Enum.map(list, fn 
        nil -> nil
        item -> item / 1
      end)
```

notebooks/exploring_explorer.livemd
```diff
@@ -126,15 +126,21 @@ And you can fill in those missing values using one of the following strategies:
 Series.fill_missing(s, :forward)
 ```
 
+In the case of mixed numeric types (i.e. integers and floats), `Series` will downcast to a float:
+
+```elixir
+Series.from_list([1, 2.0])
+```
+
 However, `Series` must all be of the same `dtype` or else you'll get an `ArgumentError`.
```
cigrainger, line 135:
```suggestion
In all other cases, `Series` must all be of the same `dtype` or else you'll get an `ArgumentError`.
```

cigrainger (commented):
Looking good. Just a few small suggestions.

PR #78, opened 2021-11-03
expose Series.transform/2 as an escape hatch

cigrainger:
I realised I never exposed this to the public-facing API. I'm not wedded to the name `map` here. Could be `apply`? Am I missing any obvious gotchas?

### DIFF
--- a/lib/explorer/backend/series.ex
+++ b/lib/explorer/backend/series.ex
@@ -106,5 +106,5 @@ defmodule Explorer.Backend.Series do
 
   # Escape hatch
 
-  @callback map(s, fun) :: s
+  @callback transform(s, fun) :: s
 end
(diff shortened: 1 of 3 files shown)


josevalim:
I think the issue with `map/2` is that people would expect it to be a very natural and common operation, while I understand this isn't really the case here as we are rebuilding the whole series? In that case, I think we need to have a more "expensive" name, such as rebuild or transform. One idea is also to call it `new/2`: i.e. you are building a series from an old series according to the given function. WDYT?

cigrainger:
Yes, we're rebuilding the whole series in order to apply arbitrary elixir functions. I think `transform/2` captures it perfectly. It's expensive-sounding but captures that we're just applying a function. `new/2` does make a lot of sense when my brain is already in this space, but I think if I put myself in the head of someone coming to the library that it would be confusing -- I'd think it was what to reach for instead of `from_list/1` or similar.

josevalim:
Sounds good. Another option if you want to be more explicit is like `rebuild` or something, but `transform` does the job too.

PR #79, opened 2021-11-03
expose Series.concat/2

### LINKED ISSUE #75: Implement `Series.concatenate`

cigrainger:
Pretty straightforward. As usual looking for naming feedback and any gotchas.

I've got two specific questions:
- Should we downcast to float if we have a mix of integer and float series (as in #77)?
- Perhaps the API should be closer to `DataFrame.concat_rows` and permit a list?

I'm mulling those over at the minute. I think the answer is yes to both.

Closes #75 

Edit: Okay this is ready now! I've applied the above.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -207,8 +207,18 @@ defmodule Explorer.Series do
         string[3]
         ["1", "2", "3"]
       >
+
+    `cast/2` will return the series as a no-op if you try to cast to the same dtype.
+
+      iex> s = Explorer.Series.from_list([1, 2, 3])
+      iex> Explorer.Series.cast(s, :integer)
+      #Explorer.Series<
+        integer[3]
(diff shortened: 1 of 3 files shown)


josevalim:
Agreed on yes to both. :)

cigrainger:
I think downcasting is going to require implementing `Series.cast/2`, which I've been meaning to add for a bit.

lib/explorer/series.ex
```diff
@@ -207,8 +207,18 @@ defmodule Explorer.Series do
         string[3]
         ["1", "2", "3"]
       >
+
+    `cast/2` will return the series as a no-op if you try to cast to the same dtype.
```
josevalim, line 211:
```suggestion
  `cast/2` will return the series as a no-op if you try to cast to the same dtype.
```

josevalim (approved):

Issue #16, opened 2021-07-26
Add more datasets


cigrainger:
We'd love to have more datasets available. Toy datasets in `explorer` itself are very welcome. And if there are ideas for linking up to and consuming datasets from https://github.com/elixir-nx/scidata that's fantastic too.

cigrainger:
I think three is a good number to close this at, but always happy to take PRs for more!

PR #81, opened 2021-11-18
feat: adds wine dataset

amferraboli:
Issue #16 points out the willing of adding more datasets to the lib.

This PR adds the Wine Dataset, a classic toy dataset for ML classification problem. The data and features were collected from https://archive.ics.uci.edu/ml/datasets/Wine and modified to a suitable .csv format.


lib/explorer/datasets.ex
```diff
@@ -12,4 +12,24 @@ defmodule Explorer.Datasets do
...
+
+    Original Owners:
+    Forina, M. et al, PARVUS -
+    An Extendible Package for Data Exploration, Classification and Correlation.
+    Institute of Pharmaceutical and Food Analysis and Technologies, Via Brigata Salerno,
+    16147 Genoa, Italy.
+
+    Wine. (1991). UCI Machine Learning Repository.
+  """
+  def wine_data,
```
cigrainger, line 33:
```suggestion
  def wine,
```

priv/datasets/wine_data.csv
```diff
@@ -0,0 +1,179 @@
+class,alcohol,malic_acid,ash,alcalinity_of_ash,magnesium,total_phenols,flavanoids,nonflavanoid_phenols,proanthocyanins,color_intensity,hue,od280/od315_of_diluted_wines,proline
```
cigrainger, line 1: I'm a bit apprehensive of a `/` in a column name. Might be the R user in me, but it feels destined to break _something_ (even though it does appear to work fine). Perhaps `od280_or_od315_of_diluted_wines`?

cigrainger (commented):
Fantastic! Thank you so much for contributing this. Just a few minor things.

cigrainger:
@amferraboli sorry to change things out from under you! With #84 the dataset will live in `datasets/wine.csv` and the function for loading it will match the updated `Explorer.Datasets.fossil_fuels/0`. If you need any help or have any questions don't hesitate to ask!

lib/explorer/datasets.ex
amferraboli: Thanks for the suggestion, adopted!

priv/datasets/wine_data.csv
amferraboli: After a little research I found out it is actually a ratio to measure protein concentration. So I ended up with `od280_over_od315_of_diluted_wines`. Let me know if you agree.

amferraboli:
@cigrainger thanks for the update. Just made the necessary adjustments to allow merging in the new folder structure and to maintain the loading dataset and naming patterns.

PR #86, opened 2021-11-22
Remove repetition in dataset module

### LINKED ISSUE #85: Proposal: Dynamically generate dataset functions at compile time
Right now, in order to add a dataset, one needs to:

1.  Add the data in `datasets`
2.  Add a function for that dataset in the `Explorer.Datasets` module.

The Problem with this is that we are essentially writing the same code repeatedly:

```elixir
def fossil_fuels,
    do: @datasets_dir |> Path.join("fossil_fuels.csv") |> DataFrame.read_csv!()
```

```elixir
def wine,
    do: @datasets_dir |> Path.join("wine.csv") |> DataFrame.read_csv!()
```

In essence:

```elixir
def unquote(dataset) do
  @datasets_dir
  |> Path.join(unquote(dataset) <> ".csv")
  |> Dataframe.read_csv!()
end
```

Note: I'm pretty new to macros, and I haven't tested this code..

I feel that this is unnecessary repetition which can easily be automated.

--------------------

I propose that we dynamically generate dataset functions at compile time.

To be more specific, I would automate it like this:

Place the data in `datasets/:name/data.csv`.
Place the docs in `datasets/:name/docs.md`.

Iterate over all dirs in `datasets` and write the above function.

Benjamin-Philip:
Closes #85.


cigrainger (approved):
LGTM

PR #90, opened 2021-12-23
turned on polars decompress support for csv

ricpruss:
Added a test for read_csv! of a gzip compressed CSV. 

Added one line to the Cargo.toml of polars to turn on support for the decompression feature.

This should close https://github.com/elixir-nx/explorer/issues/89.


cigrainger:
Fantastic! Thanks so much.

PR #97, opened 2022-01-11
Remove .tool-versions from versioning

philss:
This makes easier to develop when you are using different versions of
Elixir and OTP.


cigrainger:
Good idea 👍

PR #98, opened 2022-01-14
Update Polars to v0.18.0

philss:
This is a minor update but since Polars' API is unstable, it also breaks
some of Explorer's APIs. Most notable are the functions from
`Explorer.Series` module (and the Series behaviour).

The changes are related to `Explorer.Series.rolling_` functions.
They have changed to accept `min_periods` (int) and `center` (bool).
The `ignore_nil?` option is gone.

For future reference, this is the diff of commits between Polars'
versions in this update:

https://github.com/pola-rs/polars/compare/2055cf205e992c5eb0809ada55e067ef4c293b0d...rust-polars-0.18.0


philss:
@cigrainger would you mind to take a look? I made the decision to propagate the API changes to Explorer. Do you think this is fine?

lib/explorer/series.ex
```diff
@@ -662,25 +662,17 @@ defmodule Explorer.Series do
...
-
       iex> s = Explorer.Series.from_list([true, false, true])
       iex> Explorer.Series.quantile(s, 0.5)
-      ** (ArgumentError) Explorer.Series.quantile/2 not implemented for dtype :boolean. Valid dtypes are [:integer, :float, :date, :datetime].
+      ** (ArgumentError) Explorer.Series.quantile/2 not implemented for dtype :boolean. Valid dtypes are [:integer, :float].
   """
   @spec quantile(series :: Series.t(), quantile :: float()) :: any()
   def quantile(%Series{dtype: dtype} = series, quantile)
-      when dtype in [:integer, :float, :date, :datetime],
+      when dtype in [:integer, :float],
```
josevalim, line 671: Should we still support date and datetime by casting and then converting back to date/datetime? Do you know what is the rationale for them to no longer support date/datetime?

josevalim: In any case, if we are no longer supporting this, the docs have to be updated. :)

lib/explorer/series.ex
```diff
@@ -1274,18 +1266,18 @@ defmodule Explorer.Series do
...
       #Explorer.Series<
-        integer[10]
-        [nil, 5, 7, 11, 13, 17, 19, 23, 25, 29]
+        float[10]
+        [1.0, 5.0, 8.0, 11.0, 14.0, 17.0, 20.0, 23.0, 26.0, 29.0]
       >
   """
-  def rolling_sum(series, window_size, weights \\ nil, ignore_nil? \\ true),
-    do: apply_impl(series, :rolling_sum, [window_size, weights, ignore_nil?])
+  def rolling_sum(series, window_size, weights \\ nil, min_periods \\ 1, center \\ false),
```
josevalim, line 1279: I am wondering if we should change the API to keyword lists. Does dplyr have a similar API? If so, how does it look?

cigrainger:
@philss At first blush this looks great. I had a branch I was working on to do much the same. I'll be able to review properly in the morning.

lib/explorer/series.ex
philss: I agree with that :+1: 
But maybe we should do in another PR :thinking: 

And dplyr doesn't seems to have rolling aggregates: https://dplyr.tidyverse.org/articles/window-functions.html?q=rolling#types-of-window-functions
But in general function options looks like keyword lists.

lib/explorer/series.ex
philss: I couldn't find the reason why they don't work anymore :/

The only problem I see trying to casting to integers and converting is that maybe the value is not a valid datetime / date anymore. I could return `nil` for invalid cases. WDYT?

@cigrainger thoughts about this? I will update the docs if you agree to remove support for this.

josevalim: The datetime and date are contiguous, so we should be fine!

cigrainger: This is where I was getting stuck as well. I couldn't find a rationale for dropping it but there's been a fair amount of reworking. `polars` is definitely moving fast and breaking things. I'm fine with dropping this for now and revisiting it at a later date.

lib/explorer/series.ex
cigrainger: Agreed it should probably be in another PR and also agreed that this should be keyword lists. And yeah, no rolling aggregates in dplyr -- they suggest using [RcppRoll](https://cran.r-project.org/web/packages/RcppRoll/RcppRoll.pdf).

test/explorer/data_frame_test.exs
```diff
@@ -259,9 +259,12 @@ defmodule Explorer.DataFrameTest do
...
     @tag :tmp_dir
-    test "can write parquet to file", %{df: df} do
-      {:ok, "test.parquet"} = DF.write_parquet(df, "test.parquet")
-      {:ok, parquet_df} = DF.read_parquet("test.parquet")
+    test "can write parquet to file", %{df: df, tmp_dir: tmp_dir} do
+      parquet_path = Path.join(tmp_dir, "test.parquet")
+
+      assert {:ok, ^parquet_path} = DF.write_parquet(df, parquet_path)
+      assert {:ok, parquet_df} = DF.read_parquet(parquet_path)
+
```
cigrainger, line 267: Thanks for this touchup!

cigrainger (approved):
Everything else looks perfect. :+1:

lib/explorer/series.ex
philss: I will let the documentation of the additional options to another PR as well :)

lib/explorer/series.ex
philss: I was able to make the casting work. It should have the same behavior as before :)

josevalim (approved):

cigrainger (approved):

PR #99, opened 2022-01-17
Change optional params to keyword lists

philss:
This commit changes the `Explorer.Series.rolling_` functions to accept a
keyword lists instead of optional parameters. This makes these functions
easier to use.


cigrainger (approved):
LGTM! Thanks for this!

PR #100, opened 2022-01-18
Update "Exploring Explorer" notebook for Livebook v0.4

philss:
This change removes the warning about deprecated way of reading
filenames. The new form now is by using Kino.Input.

This change also removes some markdown code blocks for options
descriptions. The idea is to read them as markdown, which is more
pleasant.

### DIFF
--- a/notebooks/exploring_explorer.livemd
+++ b/notebooks/exploring_explorer.livemd
@@ -4,11 +4,10 @@
 
 Connect to a `Mix` project with `Explorer` installed or:
 
-<!-- livebook:{"force_markdown":true} -->
-
 ```elixir
 Mix.install([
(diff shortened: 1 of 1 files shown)


notebooks/exploring_explorer.livemd
```diff
@@ -4,19 +4,17 @@
 
 Connect to a `Mix` project with `Explorer` installed or:
 
-<!-- livebook:{"force_markdown":true} -->
```
philss, line 7: I couldn't understand the intention here. I removed in order to run the Elixir cell. Is that OK?

cigrainger: Definitely okay. I'm not sure why I had set it to markdown in the first place.

cigrainger (approved):
Brilliant! <3

PR #102, opened 2022-01-24
feature: Adds the iris dataset

kimjoaoun:
#16 asks for more datasets, so here they are!
 
I downloaded the file from [UCL's ML Repository](https://archive.ics.uci.edu/ml/datasets/iris), loaded it in R, added the column names and exported it in CSV.


cigrainger (approved):
Looks great. Thanks so much!

PR #106, opened 2022-01-27
More examples for Series functions

kimjoaoun:
It adds examples for: `Series.(divide|reverse|sort|pow|take_every|cum_sum)`.

Sending examples while trying to understand how the Rust-Elixir interop works :)

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -338,6 +338,24 @@ defmodule Explorer.Series do
 
   @doc """
   Takes every *n*th value in this series, returned as a new series.
+
+  ## Examples
+
+      iex> s1 = 1..10 |> Enum.to_list() |> Explorer.Series.from_list()
+      iex> s1 |> Explorer.Series.take_every(2)
+      #Explorer.Series<
+        integer[5]
+        [1, 3, 5, 7, 9]
+      >
+
+    If *n* is bigger than the length of the series, the result is a new Series with only the first value of supplied one.
+
+      iex> s2 = 1..10 |> Enum.to_list() |> Explorer.Series.from_list()
+      iex> s2 |> Explorer.Series.take_every(20)
+      #Explorer.Series<
+        integer[1]
+        [1]
+      >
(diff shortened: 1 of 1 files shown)


lib/explorer/series.ex
```diff
@@ -869,6 +903,16 @@ defmodule Explorer.Series do
 
     * `:integer`
     * `:float`
+
+  ## Examples
```
cigrainger, line 907: I think there needs to be a line under this.

lib/explorer/series.ex
```diff
@@ -891,6 +935,14 @@ defmodule Explorer.Series do
 
     * `:integer`
     * `:float`
+
+  ## Example
```
cigrainger, line 939: An empty line under this too.

lib/explorer/series.ex
```diff
@@ -1215,6 +1267,16 @@ defmodule Explorer.Series do
 
   @doc """
   Sorts the series.
+
+  ## Example
```
cigrainger, line 1271: Even if there's only one example, let's keep calling it 'Examples'.

lib/explorer/series.ex
```diff
@@ -338,6 +338,24 @@ defmodule Explorer.Series do
...
+  ## Examples
+
+      iex> s = 1..10 |> Enum.to_list() |> Explorer.Series.from_list()
+      iex> s |> Explorer.Series.take_every(2)
+      #Explorer.Series<
+        integer[5]
+        [1, 3, 5, 7, 9]
+      >
+
+    If *n* is bigger than the length of the series, the result is a new Series with only the first value of supplied one.
```
cigrainger, line 351:
```suggestion
    If *n* is bigger than the length of the series, the result is a new series with only the first value of the supplied series.
```

cigrainger (commented):
Thanks so much for this. It really makes a difference. Just a few suggestions.

cigrainger (approved):

PR #108, opened 2022-01-28
Explorer.Series.pow integer support

### LINKED ISSUE #105: Explorer.Series.pow does not work when supplied integers as exponent
`Explorer.Series.pow` returns an error when the user supplies an integer as exponent. It only works when exponent is a float, and I think because of that all values in the Series are casted to floating point numbers, which is not a desirable behaviour.

When exponent is a integer:

```Ex
iex> s1 = [8, 16, 32] |> Explorer.Series.from_list()
#Explorer.Series<
  integer[3] 
  [8, 16, 32]
>
iex> Explorer.Series.pow(s1, 2)
** (ArgumentError) argument error
    (explorer 0.1.0-dev) Explorer.PolarsBackend.Native.s_pow(shape: (3,)
Series: '' [i64]
[
        8
        16
        32
], 1)
    (explorer 0.1.0-dev) lib/explorer/polars_backend/shared.ex:14: Explorer.PolarsBackend.Shared.apply_native/3
```

When exponent is a float:

```Ex
iex> s1 = [8, 16, 32] |> Explorer.Series.from_list()
iex> Explorer.Series.pow(s1, 2.0)
#Explorer.Series<
  float[3]
  [64.0, 256.0, 1024.0]
>
```

kimjoaoun:
Adds support for Integers on Series.pow()

The version in this PR supports only positive integers. 

This pull request, when completed, closes #105.

### DIFF
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -235,9 +235,11 @@ defmodule Explorer.PolarsBackend.Series do
   def divide(left, right) when is_number(right), do: divide(left, scalar_rhs(right, left))
 
   @impl true
-  def pow(left, exponent) when is_number(exponent),
+  def pow(left, exponent) when is_float(exponent),
     do: Shared.apply_native(left, :s_pow, [exponent])
 
+  def pow(left, exponent) when is_integer(exponent),
+    do: Shared.apply_native(left, :s_int_pow, [exponent])
   # Comparisons
 
   @impl true
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -235,9 +235,7 @@ defmodule Explorer.PolarsBackend.Series do
   def divide(left, right) when is_number(right), do: divide(left, scalar_rhs(right, left))
 
   @impl true
-  def pow(left, exponent) when is_float(exponent) do
-        Shared.apply_native(left, :s_pow, [exponent])
-  end
+  def pow(left, exponent) when is_float(exponent), do: Shared.apply_native(left, :s_pow, [exponent])
 
   def pow(left, exponent) when is_integer(exponent) and exponent >= 0 do
     cond do
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -238,7 +238,7 @@ defmodule Explorer.PolarsBackend.Series do
   def pow(left, exponent) when is_float(exponent),
     do: Shared.apply_native(left, :s_pow, [exponent])
 
-  def pow(left, exponent) when is_integer(exponent),
+  def pow(left, exponent) when is_integer(exponent) and exponent >= 0,
     do: Shared.apply_native(left, :s_int_pow, [exponent])
   # Comparisons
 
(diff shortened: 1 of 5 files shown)


josevalim (approved):

kimjoaoun:
Currently working on adding support for negative integers.

kimjoaoun:
Marked as ready for review since I don't think It'll be necessary to add support do negative integers, since not even Elixir's `Integer.pow/2` supports them.
You can use Explorer.Series.pow with a negative floating point number, as expected.

josevalim:
Yeah, the issue with supporting negative integers is that they return floats, so you have an api signature that may return integer or floats and that can be confusing. So if you have negative integers, you need to cast them to floats before.

lib/explorer/polars_backend/series.ex
```diff
@@ -235,9 +235,11 @@ defmodule Explorer.PolarsBackend.Series do
   def divide(left, right) when is_number(right), do: divide(left, scalar_rhs(right, left))
 
   @impl true
-  def pow(left, exponent) when is_number(exponent),
+  def pow(left, exponent) when is_float(exponent),
     do: Shared.apply_native(left, :s_pow, [exponent])
 
+  def pow(left, exponent) when is_integer(exponent),
```
josevalim, line 241:
```suggestion
  def pow(left, exponent) when is_integer(exponent) and exponent >= 0,
```

kimjoaoun: I think this change would result in a exception if the user try to evaluate the following:

```Ex
iex> s = [2, 4, 6] |> Explorer.Series.from_list()
iex> Explorer.Series.pow(s, -3.0)
```

native/explorer/src/series.rs
```diff
@@ -708,6 +708,13 @@ pub fn s_pow(data: ExSeries, exponent: f64) -> Result<ExSeries, ExplorerError> {
     Ok(ExSeries::new(s.pow(exponent)?))
 }
 
+#[rustler::nif]
+pub fn s_int_pow(data: ExSeries, exponent: i64) -> Result<ExSeries, ExplorerError> {
```
cigrainger, line 712: You should actually be able to just use `u32` in the signature here for `exponent` and `rustler` should automatically marshall correctly.

cigrainger (commented):
Fantastic! Thanks for doing this so quickly. Might be worth updating `s_pow` to use the same pattern as Ritchie indicated `Series::pow` will be deprecated and it's just one less thing to do down the road. But perfectly fine not to as well :).

kimjoaoun:
@cigrainger It's my pleasure to help! I'm learning a lot in these contributions, it is being an amazing experience :')

I updated `s_pow` but I'm not an avid Rust programmer, so I don't know if I did so in the best implementation possible.

cigrainger (approved):
It looks good to me! Thanks again, this is great.

cigrainger:
Okay, one last request and then I think it's ready to merge -- could you please add an example with a float series and float exponent? Also, what happens if you have a float series and int exponent?

kimjoaoun:
> Okay, one last request and then I think it's ready to merge -- could you please add an example with a float series and float exponent? Also, what happens if you have a float series and int exponent?

I added the docs. Before these recent commits if you tried to run code with a float series and a int exponent it would break, but now I fixed this case.

cigrainger (approved):
Fantastic! Thanks again.

PR #110, opened 2022-02-03
Tabular Printing for DataFrames

### LINKED ISSUE #22: Add tabular print method
Useful influences:
https://github.com/alexhallam/tv
https://pillar.r-lib.org/dev/index.html

kimjoaoun:
When completed, this PR closes #22.

I don't know if the code is in the right place, nor if it is the right way to implement this feature, but hey, it works!

But it is not ready yet, I'm trying to find a way to give the "second row" (the one with column `<types>`) a different color to differentiate it from the ones with data values.

```
iex(15)> Explorer.Datasets.iris |> Explorer.Inspect.df_pretty_inspect()  
+--------------+-------------+--------------+-------------+-------------+
| sepal_length | sepal_width | petal_length | petal_width | species     |
+==============+=============+==============+=============+=============+
| <float>      | <float>     | <float>      | <float>     | <string>    |
+--------------+-------------+--------------+-------------+-------------+
| 5.1          | 3.5         | 1.4          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 4.9          | 3.0         | 1.4          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 4.7          | 3.2         | 1.3          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 4.6          | 3.1         | 1.5          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 5.0          | 3.6         | 1.4          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+

:ok
```

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1659,8 +1659,7 @@ defmodule Explorer.DataFrame do
     types =
       df
       |> dtypes()
-      |> Enum.map(&Atom.to_string(&1))
-      |> Enum.map(fn x -> "<" <> x <> ">" end)
+      |> Enum.map(&"<#{Atom.to_string(&1)}>")
 
     values =
       headers
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1647,8 +1647,8 @@ defmodule Explorer.DataFrame do
 
    ## Examples
 
-      iex> df = Explorer.Datasets.iris()
-      iex> Explorer.DataFrame.table(df)
+      df = Explorer.Datasets.iris()
+      Explorer.DataFrame.table(df)
   """
   def table(df, nrow \\ 5) when nrow >= 0 do
     {rows, cols} = shape(df)
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1651,8 +1651,8 @@ defmodule Explorer.DataFrame do
        Explorer.DataFrame.table(df)
   """
   def table(df, nrow \\ 5) when nrow >= 0 do
-    {rows, cols} = Explorer.DataFrame.shape(df)
-    headers = Explorer.DataFrame.names(df)
+    {rows, cols} = shape(df)
+    headers = names(df)
 
     df = slice(df, 0, nrow)
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1666,7 +1666,7 @@ defmodule Explorer.DataFrame do
       |> Enum.map(&Series.to_list(df[&1]))
       |> Enum.zip_with(& &1)
 
-    name_type = Enum.zip_with(headers, types, fn x, y -> x<>y end)
+    name_type = Enum.zip_with(headers, types, fn x, y -> x <> y end)
 
     TableRex.Table.new()
     |> TableRex.Table.put_title("Explorer DataFrame: [rows: #{rows}, columns: #{cols}]")
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1659,17 +1659,19 @@ defmodule Explorer.DataFrame do
     types =
       df
       |> dtypes()
-      |> Enum.map(&"<#{Atom.to_string(&1)}>")
+      |> Enum.map(&"\n<#{Atom.to_string(&1)}>")
 
     values =
       headers
       |> Enum.map(&Series.to_list(df[&1]))
       |> Enum.zip_with(& &1)
 
+    name_type = Enum.zip_with(headers, types, fn x, y -> x<>y end)
+
     TableRex.Table.new()
     |> TableRex.Table.put_title("Explorer DataFrame: [rows: #{rows}, columns: #{cols}]")
-    |> TableRex.Table.put_header(headers)
-    |> TableRex.Table.add_row(types)
+    |> TableRex.Table.put_header(name_type)
+    |> TableRex.Table.put_header_meta(0..cols, align: :center)
     |> TableRex.Table.add_rows(values)
     |> TableRex.Table.render!(
       header_separator_symbol: "=",
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1642,6 +1642,35 @@ defmodule Explorer.DataFrame do
        when is_atom(colname) and is_map(acc) and is_list(value),
        do: mutate_reducer({Atom.to_string(colname), value}, acc, df)
 
+  def table(df, nrow \\ 5) when nrow >= 0 do
+    df = Explorer.DataFrame.slice(df, 0, nrow)
+
+    types =
+      df
+      |> Explorer.DataFrame.dtypes()
+      |> Enum.map(&Atom.to_string(&1))
+      |> Enum.map(fn x -> "<" <> x <> ">" end)
+
+    headers = Explorer.DataFrame.names(df)
+
+    values =
+      headers
+      |> Enum.map(&Explorer.DataFrame.pull(df, &1))
+      |> Enum.map(&Explorer.Series.to_list(&1))
+      |> Enum.zip()
+      |> Enum.map(&Tuple.to_list(&1))
+
+    TableRex.Table.new()
+    |> TableRex.Table.put_header(headers)
+    |> TableRex.Table.add_row(types)
+    |> TableRex.Table.add_rows(values)
+    |> TableRex.Table.render!(
+      header_separator_symbol: "=",
+      horizontal_style: :all
+    )
+    |> IO.puts()
+  end
+
   # Helpers
 
   defp backend_from_options!(opts) do
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1654,7 +1654,7 @@ defmodule Explorer.DataFrame do
     {rows, cols} = Explorer.DataFrame.shape(df)
     headers = Explorer.DataFrame.names(df)
 
-    df = Explorer.DataFrame.slice(df, 0, nrow)
+    df = slice(df, 0, nrow)
 
     types =
       df
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1643,13 +1643,13 @@ defmodule Explorer.DataFrame do
        do: mutate_reducer({Atom.to_string(colname), value}, acc, df)
 
   @doc """
-  Display the DataFrame in a tabular fashion.
+   Display the DataFrame in a tabular fashion.
 
-  ## Examples
+   ## Examples
 
-      df = Explorer.Datasets.fossil_fuels()
-      Explorer.DataFrame.table(df)
- """
+       df = Explorer.Datasets.fossil_fuels()
+       Explorer.DataFrame.table(df)
+  """
   def table(df, nrow \\ 5) when nrow >= 0 do
     df = Explorer.DataFrame.slice(df, 0, nrow)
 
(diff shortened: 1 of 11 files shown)


cigrainger:
One thing I'll be loathe to lose in a default inspect is the number of rows, columns, and groups. That is the second line here:

```
#Explorer.DataFrame<
  [rows: 150, columns: 5]
```

I wonder if there's a way to include that info with `table_rex`?

kimjoaoun:
> One thing I'll be loathe to lose in a default inspect is the number of rows, columns, and groups. That is the second line here:
> 
> ```
> #Explorer.DataFrame<
>   [rows: 150, columns: 5]
> ```
> 
> I wonder if there's a way to include that info with `table_rex`?

I can add these things in the title field, but it wouldn't look great.

Table Rex's API looks really flexible, we can find a way to put this infos somewhere.

lib/explorer/polars_backend/native.ex
```diff
@@ -69,6 +69,7 @@ defmodule Explorer.PolarsBackend.Native do
   def df_n_chunks(_df), do: err()
   def df_new(_cols), do: err()
   def df_pivot_wider(_df, _id_cols, _pivot_column, _values_columns), do: err()
+  def df_pretty_inspect(_df, _nrow), do: err()
```
josevalim, line 72: Is this a data frame callback? Instead shouldn't we have the whole printing in Elixir instead instead of native?

lib/explorer/inspect.ex
```diff
@@ -59,6 +59,33 @@ defmodule Explorer.Inspect do
...
+
+    TableRex.Table.new()
+      |> TableRex.Table.put_header(headers)
+      |> TableRex.Table.add_row(types)
+      |> TableRex.Table.add_rows(values)
+      |> TableRex.Table.render!(
+          header_separator_symbol: "=",
+          horizontal_style: :all)
+      |> IO.puts
+  end
```
josevalim, line 87: Btw, it is probably too early, but don't forget to run `mix format` to get formatted code later. :)

lib/explorer/polars_backend/native.ex
kimjoaoun: For some reason I thought that we needed this callback, I'll remove it.
Sorry 😳

josevalim: No worries at all. Also note that Explorer.Inspect is private so we may want a public API for the inspect function. We may also want to filter the table somehow... printing the whole table can be too much? Should we automatically limit it 100 elements and have a `limit: ` flag?

josevalim: No worries at all. Also note that Explorer.Inspect is private so we may want a public API for the inspect function.

kimjoaoun: Well, I'll need some help on how to do that. I'm new on that Elixir thing 😅

josevalim: @kimjoaoun I will be glad to help! The first question is what the public API should be and once @cigrainger lets us know his preference we will get it done. :)

cigrainger: @josevalim do you mean re: how to inspect? I think the table view as the default `Inspect` implementation and then `Explorer.DataFrame.glimpse()` for the current (non-tabular) inspect implementation.

kimjoaoun: Understood. So the current `Inspect` would become a normal function and the tabular goes as the `Inspect`.
Wouldn't this break all the doctests?

cigrainger: Exactly. And yes, yes it would. :grimacing:

cigrainger: I'm also happy to flip the other direction and add the tabular version as a normal function. That would be `Explorer.DataFrame.table()` or similar. We had had some discussions in which we figured more people would be comfortable with a table by default. I prefer the `glimpse` inspired version for a number of reasons. Perhaps we should start by adding the table as a normal function and see how we go?

kimjoaoun: Well, I see no problem in doing all in this PR, but it'll be a long one with a lot of changes.
So the roadmap of this PR now is:

- [ ] Adding `nrow` and `ncol` to table printing.
- [ ] Replacing the current Inspect with the table printer
- [ ] Fixing all tests.

josevalim: A table function sounds like a great starting point!

kimjoaoun: Ok!
I think I'll need to start a new PR, since I deleted my fork, but I forgot doing so I'm not be able to do changes in this PR.

cigrainger: I agree, let's start with the table function in this PR and then reassess.

kimjoaoun:
What do you think of this layout? It is the one implemented in the last commit.

```elixir
iex(7)> Explorer.Datasets.iris |> Explorer.DataFrame.table(1)
+-----------------------------------------------------------------------+
|              Explorer DataFrame: [rows: 150, columns: 5]              |
+--------------+-------------+--------------+-------------+-------------+
| sepal_length | sepal_width | petal_length | petal_width | species     |
+==============+=============+==============+=============+=============+
| <float>      | <float>     | <float>      | <float>     | <string>    |
+--------------+-------------+--------------+-------------+-------------+
| 5.1          | 3.5         | 1.4          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
```

cigrainger:
I really like it. Giving myself a moment to think it through.

lib/explorer/data_frame.ex
```diff
@@ -1642,6 +1642,45 @@ defmodule Explorer.DataFrame do
...
+    {rows, cols} = Explorer.DataFrame.shape(df)
+    headers = Explorer.DataFrame.names(df)
+
+    df = Explorer.DataFrame.slice(df, 0, nrow)
+
+    types =
+      df
+      |> Explorer.DataFrame.dtypes()
+      |> Enum.map(&Atom.to_string(&1))
+      |> Enum.map(fn x -> "<" <> x <> ">" end)
```
cigrainger, line 1663:
```suggestion
    types =
      df
      |> Explorer.DataFrame.dtypes()
      |> Enum.map(&"<#{Atom.to_string(&1)}>")
```

lib/explorer/data_frame.ex
```diff
@@ -1642,6 +1642,45 @@ defmodule Explorer.DataFrame do
...
+
+    values =
+      headers
+      |> Enum.map(&Explorer.DataFrame.pull(df, &1))
+      |> Enum.map(&Explorer.Series.to_list(&1))
+      |> Enum.zip()
+      |> Enum.map(&Tuple.to_list(&1))
```
cigrainger, line 1670:
```suggestion
    values =
      headers
      |> Enum.map(&Explorer.Series.to_list(df[&1]))
      |> Enum.zip_with(&Tuple.to_list(&1))
```

cigrainger: Just a bit of code golf ;)

lib/explorer/data_frame.ex
```diff
@@ -1642,6 +1642,45 @@ defmodule Explorer.DataFrame do
...
+  """
+  def table(df, nrow \\ 5) when nrow >= 0 do
+    {rows, cols} = Explorer.DataFrame.shape(df)
+    headers = Explorer.DataFrame.names(df)
+
+    df = Explorer.DataFrame.slice(df, 0, nrow)
+
+    types =
+      df
+      |> Explorer.DataFrame.dtypes()
```
cigrainger, line 1661:
```suggestion
      |> dtypes()
```

lib/explorer/data_frame.ex
```diff
@@ -1642,6 +1642,45 @@ defmodule Explorer.DataFrame do
...
+   ## Examples
+
+       df = Explorer.Datasets.fossil_fuels()
+       Explorer.DataFrame.table(df)
+  """
+  def table(df, nrow \\ 5) when nrow >= 0 do
+    {rows, cols} = Explorer.DataFrame.shape(df)
+    headers = Explorer.DataFrame.names(df)
+
+    df = Explorer.DataFrame.slice(df, 0, nrow)
```
cigrainger, line 1657:
```suggestion
    df = slice(df, 0, nrow)
```

lib/explorer/data_frame.ex
```diff
@@ -1642,6 +1642,45 @@ defmodule Explorer.DataFrame do
...
+   Display the DataFrame in a tabular fashion.
+
+   ## Examples
+
+       df = Explorer.Datasets.fossil_fuels()
+       Explorer.DataFrame.table(df)
+  """
+  def table(df, nrow \\ 5) when nrow >= 0 do
+    {rows, cols} = Explorer.DataFrame.shape(df)
+    headers = Explorer.DataFrame.names(df)
```
cigrainger, line 1655:
```suggestion
    {rows, cols} = shape(df)
    headers = names(df)
```

lib/explorer/data_frame.ex
kimjoaoun: wow, i didn't knew that we could do `df["col_name"]` to get a Series, cool!

kimjoaoun:
New aesthetic changes :)
Now the type of the column is in the same cell as the column name.

Output below:

```
+-----------------------------------------------------------------------+
|              Explorer DataFrame: [rows: 150, columns: 5]              |
+--------------+-------------+--------------+-------------+-------------+
| sepal_length | sepal_width | petal_length | petal_width |   species   |
|   <float>    |   <float>   |   <float>    |   <float>   |  <string>   |
+==============+=============+==============+=============+=============+
| 5.1          | 3.5         | 1.4          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 4.9          | 3.0         | 1.4          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 4.7          | 3.2         | 1.3          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 4.6          | 3.1         | 1.5          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
| 5.0          | 3.6         | 1.4          | 0.2         | Iris-setosa |
+--------------+-------------+--------------+-------------+-------------+
```

lib/explorer/data_frame.ex
```diff
@@ -1642,6 +1642,44 @@ defmodule Explorer.DataFrame do
        when is_atom(colname) and is_map(acc) and is_list(value),
        do: mutate_reducer({Atom.to_string(colname), value}, acc, df)
 
+  @doc """
+   Display the DataFrame in a tabular fashion.
+
+   ## Examples
+
+       df = Explorer.Datasets.fossil_fuels()
+       Explorer.DataFrame.table(df)
```
cigrainger, line 1651: I think `Explorer.Datasets.iris/0` is sufficiently small that it would make sense to put it here and make the doctest functional.

cigrainger (commented):
Looking great. Just the one point about the doctest and then I think it's ready to go!

lib/explorer/data_frame.ex
kimjoaoun: @cigrainger I'm having problems at making the doctest work.
Can you do this part?

cigrainger: Sure!

cigrainger: Oh... I'm sorry I was being silly. Doctests only accept valid Elixir syntax or opaque data starting with `#`. Perfectly fine to leave it without the functional doctest. I'll add a normal test.

PR #112, opened 2022-02-08
improving code coverage and documentation

guigaoliveira:
Hi,

I'm studying Explorer code as a hobby and I decided to improve test coverage and consequently documentation. Few changes were also made based on suggestions from dialyzer and credo (I installed locally), but I can revert if necessary. I also ran "mix format". I think I also found a strange bug in the `Explorer.DataFrame.pivot_wider/4` function and will report it in an issue.


cigrainger (approved):
Hey fantastic set of catches! Thanks so much for your contribution. Definitely looking forward to the issue you found.

PR #114, opened 2022-02-08
Parse dates and limit infer schema row number exposed

nyo16:
This PR exposing parse_dates and infer_schema_length options that polar-rs lib supports.


cigrainger (approved):
Fantastic. Thank you! No changes needed.

Issue #91, opened 2021-12-23
Reading data from a web URL


kimjoaoun:
- Proposal: Reading a dataset (csv?) from the internet with `read_csv`.

I think it would be great if we had a way to read data from a URL when using `Explorer.read_csv()`.

In R's [library readr](https://github.com/tidyverse/readr) package (which uses [lib vroom](https://github.com/r-lib/vroom) as a backend) you can supply a string that starts with "http", "https", "ftp" or even "ftps" and it will connect to that address and download the data from the link. I think it downloads the data in a temporary location, loads it in `R` and then deletes it.
I think it would be great to have a similar functionality in Explorer, because it is specially useful when giving lectures/classes.

- Implementation idea:

It could be implemented with pattern matching, by matching the beginning of the string with those protocols identifiers and downloading it in a temporary location, loading it in Elixir using the already existing `Explorer.read_csv()`, and then deleting the file.
I could even try to help implementing if necessary.

Something like the example below would be perfect:

```Ex
url = "https://raw.githubusercontent.com/dadosgovbr/catalogos-dados-brasil/master/dados/catalogos.csv"

df =
  url
  |> Explorer.read_csv()
```

Edit: It is also a functionality available in Python's Pandas.

cigrainger:
Definitely a great idea. I don't think we'll need to download to a temporary location -- we can just stream the file into memory. If you're keen to have a swing at it, feel free. I won't be able to get to it till after the holidays. But I'll have a go then otherwise. :+1:

kimjoaoun:
I'll try to do it :)

kimjoaoun:
Ok @cigrainger I've just realised that might be too complex for me. You are free to do it :)

cigrainger:
No prob. Thanks!

kimjoaoun:
Oh, I've just realised that [scidata](https://github.com/elixir-nx/scidata/blob/master/lib/scidata/utils.ex)'s `get!()` implements something similar. It might be useful taking a look at it.

srowley:
I would be willing to take a stab at this. I am inclined to use an HTTP client library like [Req](https://github.com/wojtekmach/req) so that you could pass options for authentication, etc., but if adding a dependency is undesirable could use `:httpc.request`, more or less copying the scidata approach. Could also make Req optional I guess and use `:httpc.request` if it isn't loaded.

WDYT?

srowley:
[...]

PR #116, opened 2022-02-10
Reading data from url example

srowley:
Here is a working example of a solution to address #91. I tried a few things to DRY out the Rust side but got stuck because I don't really know Rust at all. Welcome suggestions for improvement.


cigrainger:
Looking good! I should be able to give this a proper review tomorrow morning (AEST). Seems like we could expose `Explorer.DataFrame.from_binary/2` to complement `Explorer.DataFrame.to_binary/1`, calling `read_csv_from_memory` directly.

cigrainger:
@josevalim totally agree. The fetching part is somewhat trivial and keeping it out of the API keeps it more flexible.

josevalim:
I have created #167 to discuss the APIs here: #116.

josevalim:
Closing in favor of #186.

PR #120, opened 2022-02-16
don't assume Series.transform/2 returns the same dtype

### LINKED ISSUE #119: DataFrame cast string to Time
I am kinda stuck with a simple problem. Since I can't find the solution in the documentation and [SO](https://stackoverflow.com/questions/71062847/elixir-explorer-dataframe-cast-string-to-time?noredirect=1#comment125638945_71062847) is quiet, maybe this is the place? Even if only this issue serves as documentation for someone else, it might be worth it?

So my simplified scenario:

I try to cast an Explorer.Series row from a string to a Time type.

```elixir
df = Explorer.DataFrame.from_map(%{a: ["00:30:00", "01:00:00", "05:30:00"]})
    
transform_duration = fn duration ->
  time = Time.from_iso8601!(duration)
  time.hour + (time.minute / 60) + (time.second / 60*60)
end

DataFrame.mutate(df, a: Series.transform(s, fn x -> transform_duration.(x) end))
```

The DataFrame.mutate function won't allow me to transform to a different type. Should I combine this Series.transform with a Series.cast?

Maybe my whole approach ain't good? It ain't working in any way I'v tried to grasp the idea.

cigrainger:
As on the tin. This fixes #119.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -22,7 +22,7 @@ defmodule Explorer.Series do
   alias __MODULE__, as: Series
   alias Kernel, as: K
 
-  import Explorer.Shared, only: [impl!: 1]
+  import Explorer.Shared, only: [impl!: 1, check_types: 1, cast_numerics: 2]
   import Nx.Defn.Kernel, only: [keyword!: 2]
   import Kernel, except: [length: 1, and: 2]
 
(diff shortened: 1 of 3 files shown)


lib/explorer/series.ex
```diff
@@ -1833,17 +1840,24 @@ defmodule Explorer.Series do
     apply(impl, fun, [series | args])
   end
 
-  defp check_types(list) do
-    {last_item, type, types_match?} =
-      Enum.reduce_while(list, {nil, nil, true}, &check_types_reducer/2)
+  def check_types(list) do
```
josevalim, line 1843: Do you want to `@doc false` those functions maybe? They are not public API, right? Or perhaps move to a module you can share them?

cigrainger: Good point!

cigrainger: How's that now?

josevalim (approved):

PR #122, opened 2022-02-18
Add workflow for building precompiled NIFs

philss:
This is going to be used to precompile the Explorer NIF in order to be
reused later.

This PR only includes the workflow and adjustments to the explorer crate. The workflow is going to run on each tag push and on each push to `main`.

It's building for the most common targets (including the ones used by Raspbarry PI/Nerves). In the future we should be able to adjust more or less targets in [RustlerPrecompiled](https://github.com/philss/rustler_precompiled).

PS: I should remove the current branch name from the list after we decide to merge or not.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,176 @@
+name: Build precompiled NIFs
+
+env:
+  NIF_DIRECTORY: "native/explorer"
+
+on:
+  push:
+    branches:
+      - main
+      - ps-add-precompilation-support
+    tags:
+      - '*'
+
+defaults:
+  run:
+    # Sets the working dir for "run" scripts.
+    # Note that this won't change the directory for actions (tasks with "uses").
+    working-directory: "./native/explorer"
+
+jobs:
+  build_release:
+    name: NIF ${{ matrix.job.nif }} - ${{ matrix.job.target }} (${{ matrix.job.os }})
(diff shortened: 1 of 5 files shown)


cigrainger:
:tada: I was just adding `floki` to a project today and looking jealously at the precompiled `html5ever` NIF. I'll give this a proper once over. Thanks @philss!

philss:
@cigrainger it's comming to explorer too! :smiley: 

Although I don't know yet how to use it without having a new release. I will investigate this today/tomorrow.

philss:
> I think we could have an application env var that forces the build. By default it is false but we can set it to true in config/config.exs (since config/config.exs is not loaded by dependencies).

I'm working on making this easier here: https://github.com/philss/rustler_precompiled/pull/6
Also we decided to always force the build if the version is a pre-release (like `0.1.0-dev`). So we avoid using an outdated (precompiled) NIF in development.

philss:
Hey guys, I open another PR with the actual `RustlerPrecompiled` config pointing to this branch at my fork => https://github.com/philss/explorer/pull/1
Do you think I should close this one and open everything in one PR?

josevalim:
Everything in one PR sounds good to me!

cigrainger:
Agreed, everything in one is perfect.

philss:
So the PR is now here => https://github.com/elixir-nx/explorer/pull/123

PR #123, opened 2022-02-21
Add `rustler_precompiled` and workflow for precompiling NIFs

philss:
This PR contains the changes necessary to use precompiled NIFs.

It will build on every push to the `main` branch and push of tags. But with tags we are also automatically creating a new release that contains all the precompiled NIFs for the most common architectures and operational systems.
After the release, the Hex package can be published as described in the [Precompilation Guide](https://hexdocs.pm/rustler_precompiled/precompilation_guide.html) (only for project maintainers).

The NIF will be downloaded based on the target system (normally the same of the host system, but can be different for example when building for [Nerves](https://www.nerves-project.org/)) and will be verified with a checksum that was generated in the release of the Hex package.

This shouldn't change much of the development flow, so if you are using a pre-release it won't affect you. Once we have a new release, the users of the Hex package will be able to use Explorer without having to install the Rust toolchain.

Even with a released version, it is still possible to force the compilation by setting a system env var:

    EXPLORER_BUILD=true mix compile --force

Or by setting an application env:

    config :rustler_precompiled, :force_build, explorer: true

For more details about `rustler_precompiled`, please [check the documentation](https://hexdocs.pm/rustler_precompiled).

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -58,11 +58,10 @@ defmodule Explorer.MixProject do
         "lib",
         "native",
         "datasets",
-        # "notebooks", should we include this one?
         "checksum-*.exs",
         "mix.exs",
         "README.md",
-        # "CHANGELOG.md",
+        "CHANGELOG.md",
         "LICENSE"
       ],
       licenses: ["MIT"],
--- a/mix.exs
+++ b/mix.exs
@@ -10,6 +10,7 @@ defmodule Explorer.MixProject do
       name: "Explorer",
       version: @version,
       elixir: "~> 1.12",
+      package: package(),
       deps: deps(),
       docs: docs()
     ]
@@ -26,7 +27,7 @@ defmodule Explorer.MixProject do
     [
       {:ex_doc, "~> 0.24", only: :dev, runtime: false},
       {:nx, "~> 0.1.0"},
-      {:rustler, "~> 0.23.0"},
+      {:rustler_precompiled, "~> 0.2"},
       {:table_rex, "~> 3.1.1"}
     ]
   end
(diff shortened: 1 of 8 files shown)


cigrainger (approved):
Amazing. I think we should put something in the README re: getting started, but I'm happy to do that in a followup PR.

Thanks @philss! Are you comfortable for this to be merged?

mix.exs
```diff
@@ -50,4 +51,22 @@ defmodule Explorer.MixProject do
...
     ]
   end
+
+  defp package do
+    [
+      files: [
+        "lib",
+        "native",
+        "datasets",
+        # "notebooks", should we include this one?
```
josevalim, line 61: Notebooks are not necessary.

```suggestion
```

mix.exs
```diff
@@ -50,4 +51,22 @@ defmodule Explorer.MixProject do
...
+        "checksum-*.exs",
+        "mix.exs",
+        "README.md",
+        # "CHANGELOG.md",
```
josevalim, line 65:
```suggestion
        "CHANGELOG.md",
```

mix.exs
```diff
@@ -50,4 +51,21 @@ defmodule Explorer.MixProject do
...
+  defp package do
+    [
+      files: [
+        "lib",
+        "native",
+        "datasets",
+        "checksum-*.exs",
+        "mix.exs",
+        "README.md",
+        "CHANGELOG.md",
```
josevalim, line 64:
```suggestion
```

josevalim (approved):
Beautiful work @philss! This is good to go!

@cigrainger for the README, we only need to note that the release package does not require Rust installed. The `EXPLORER_BUILD` flag that only makes sense once the package is published. Everything else should just work™!

philss:
> I think we should put something in the README re: getting started, but I'm happy to do that in a followup PR.

Oh, please go ahead then. I think the only thing to mention is the ways to force the build when people are using a released version, like José said.

I think it's ready to go! :D
 But let me know if you need help when releasing the package for the first time. The only detail is the checksum file that depends on the "release build". The [Release flow](https://hexdocs.pm/rustler_precompiled/precompilation_guide.html#the-release-flow) explains that in depth.

Thanks, guys!

dustinfarris:
I'm having trouble using this — explorer is still compiling polars

dustinfarris:
possible to get a pre-release then? 😅

PR #125, opened 2022-02-23
Always build NIF on CI

philss:
This change makes sure that we always build the NIF on CI. Even if we
are releasing an official version, we should build from source.
This will prevent the build to fail in those occasions, where the
precompiled NIFs won't be available yet - they take 30 minutes to build.

This change also includes a restriction for running the "release"
workflow - that builds the precompiled NIFs - only if the "native" code
changed. This is because we almost don't depend on Elixir code for that
workflow.

This is related to the previous PR #123

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
 env:
   ELIXIR_VERSION: 1.13
   OTP_VERSION: 24.2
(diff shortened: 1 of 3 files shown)


josevalim (approved):

cigrainger (approved):
👍

PR #127, opened 2022-03-07
Creates a DataFrame from a list of maps

### LINKED ISSUE #59: create data frame from records
We are trying out explorer in our application to do some last-mile data shaping.

We pull records from the db and it would be nice if DataFrame had a function to ingest those.

Currently doing this:

```ex
data = [%{id: 1, name: "John"}, %{id: 2, name: "Jane"}]

df =
  data
  |> Enum.zip_with(
    fn [{key, _value} | _rest] = zipped_column_enum -> 
      %{key => Enum.map(zipped_column_enum, fn {_key, value} -> value end)} 
    end
  )
  |> Enum.reduce(&Map.merge/2)
  |> Explorer.DataFrame.from_map()
```

kimjoaoun:
This pull request adds a function called `Explorer.DataFrame.from_list_of_maps/1`, which accepts a list of maps and transforms it in a DataFrame. Each `map` corresponds to a row in the `DataFrame` and each entry in a map is mapped to a column. All maps must have the same length and the same key names.

If approved, this pull request closes #59.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -193,6 +193,28 @@ defmodule Explorer.DataFrame do
   def to_map(df, convert_series? \\ true), do: apply_impl(df, :to_map, [convert_series?])
 
   @doc """
+  Converts a list of maps to a dataframe.
+
+  ## Examples
+
+      iex> l = [%{id: 1, name: "José"}, %{id: 2, name: "Christopher"}, %{id: 3, name: "Cristine"}]
+      iex> Explorer.DataFrame.from_list(l)
+      #Explorer.DataFrame<
+        [rows: 3, columns: 2]
+        id integer [1, 2, 3]
+        name string ["José", "Christopher", "Cristine"]
+      >
+  """
+  def from_list([h | t]) do
+    for _record <- t, reduce: record_to_df(h) do
+      acc -> t |> record_to_df() |> then(&DataFrame.concat_rows(acc, &1))
+    end
+  end
+
+  defp wrap_values(map), do: Map.new(map, fn {k, v} -> {k, List.wrap(v)} end) # OK
+  defp record_to_df(map), do: map |> wrap_values() |> DataFrame.from_map() # OK
+
+  @doc """
   Writes a dataframe to a binary representation of a delimited file.
 
   ## Options
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -213,7 +213,7 @@ defmodule Explorer.DataFrame do
     end)
   end
 
-  defp wrap_values(map) do
+  defp wrap_values(map) when is_map(map) do
     Map.new(map, fn {key, value} -> {key, List.wrap(value)} end)
   end
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -198,14 +198,14 @@ defmodule Explorer.DataFrame do
   ## Examples
 
       iex> l = [%{id: 1, name: "José"}, %{id: 2, name: "Christopher"}, %{id: 3, name: "Cristine"}]
-      iex> Explorer.DataFrame.from_list(l)
+      iex> Explorer.DataFrame.from_list_of_maps(l)
       #Explorer.DataFrame<
         [rows: 3, columns: 2]
         id integer [1, 2, 3]
         name string ["José", "Christopher", "Cristine"]
       >
   """
-  def from_list([head | tail]) do
+  def from_list_of_maps([head | tail]) do
     Enum.reduce(tail, record_to_df(head), fn record, acc ->
       record
       |> record_to_df()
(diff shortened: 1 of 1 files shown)


josevalim:
Should we call it from_list instead? Much simpler IMO. Another option is to rename from_list/from_map to from_rows/from_series respectively.

kimjoaoun:
> Should we call it from_list instead? Much simpler IMO. Another option is to rename from_list/from_map to from_rows/from_series respectively.

I'll change the name. I've just realised that this naming discussion happend on #59 and `from_list` won.

lib/explorer/data_frame.ex
```diff
@@ -192,6 +192,33 @@ defmodule Explorer.DataFrame do
...
+
+      iex> l = [%{id: 1, name: "José"}, %{id: 2, name: "Christopher"}, %{id: 3, name: "Cristine"}]
+      iex> Explorer.DataFrame.from_list(l)
+      #Explorer.DataFrame<
+        [rows: 3, columns: 2]
+        id integer [1, 2, 3]
+        name string ["José", "Christopher", "Cristine"]
+      >
+  """
+  def from_list([head | tail]) do
```
guigaoliveira, line 208:
```suggestion
  def from_list([head | tail]) when is_map(head) do
```

lib/explorer/data_frame.ex
```diff
@@ -192,6 +192,33 @@ defmodule Explorer.DataFrame do
...
+    Enum.reduce(tail, record_to_df(head), fn record, acc ->
+      record
+      |> record_to_df()
+      |> then(&DataFrame.concat_rows(acc, &1))
+    end)
+  end
+
+  defp wrap_values(map) do
```
guigaoliveira, line 216:
```suggestion
  defp wrap_values(map) when is_map(map) do
```

cigrainger (approved):
Fantastic. My only question is how does it perform? Is it reasonably fast? That was my only worry when I suggested reducing with `concat_rows`.

kimjoaoun:
Hmm, I haven't check that, @cigrainger. Do you have any suggestion on how to test the performance?

cigrainger:
> Hmm, I haven't check that, @cigrainger. Do you have any suggestion on how to test the performance?

Honestly, I haven't dug into this very much. I'm not sure how observer plays with NIFs, for example, and whether we can get a sense of actual memory usage (which is mostly what I'm worried about). `benchee` should give you a sense of the amount of time it takes. I'd be interested to see what scaling looks like at some numbers of rows. Say 100, 1_000, 1_000_000.

I'm not overly concerned because this should be entirely obviated by lazy eval, but in the meantime it would be nice know we're not going to completely blow up anyone's machine by creating incrementally larger dataframes for every row they want to append before the GC kicks in.

josevalim:
It may also be more efficient to create a map from the list completely in Elixir, so benchmarking is a good idea. I would pick one of the existing datasets in Explorer and use them for the benchmarking.

cigrainger:
@kimjoaoun Sorry I haven't gotten to benchmarking this yet, but I will soon. :pray:

kimjoaoun:
No problem, @cigrainger!

philss:
I took the opportunity to compare this function with `from_map/1` and the results were quite interesting:

```
Operating System: Linux
CPU Information: AMD Ryzen 9 5950X 16-Core Processor
Number of Available Cores: 32
Available memory: 15.57 GB
Elixir 1.13.1
Erlang 24.1.5

warmup: 2 s
time: 10 s
memory time: 2 s
parallel: 1
inputs: none specified
Estimated total run time: 28 s

Benchmarking from_list...
Benchmarking from_map...

Name                ips        average  deviation         median         99th %
from_map       118.15 K        8.46 μs   ±982.62%        5.09 μs       13.28 μs
from_list       13.54 K       73.86 μs    ±34.00%       64.56 μs      167.95 μs

Comparison: 
from_map       118.15 K
from_list       13.54 K - 8.73x slower +65.39 μs

Memory usage statistics:

Name         Memory usage
from_map             4 KB
from_list        44.92 KB - 11.23x memory usage +40.92 KB

**All measurements for memory usage were the same**
```
I think the memory usage is not accurate because the BEAM doesn't have access to the entire usage of the NIF. But I assume it will use more memory anyway.

The [source code is here](https://gist.github.com/philss/1211fb82d108bd21cffb217ed5dc968c). You also can:

 [Run in Livebook](https://livebook.dev/run?url=https%3A%2F%2Fgist.githubusercontent.com%2Fphilss%2F1211fb82d108bd21cffb217ed5dc968c%2Fraw%2Fb6a93f465b7f526e71bd0cab5512e507bf053b8e%2Fbench-from-list.md.livemd)

cigrainger:
Thanks @philss! This is very good to know. One little issue is that the comparison should include creating the `example_map` from the list of maps. The more I think about this the more obvious it is that (at least until we're on lazy mode and probably even then) we should be leaning on Rust for this. There's `DataFrame::from_rows` that we can use. And Rustler will handle hashmaps (not sure about keyword lists yet, but I don't see why not). @kimjoaoun Sorry but I'm going to take this a completely different direction. PR incoming. 

Just for completeness, if I include the creation of the `example_map` in the benchmark function, then it yields this:
```

kimjoaoun:
No problem at all, @cigrainger! Can you tag me in the incoming PR? I'd like to see the implementation!

cigrainger:
Closing in favour of #137.

PR #130, opened 2022-03-16
bump polars

### LINKED ISSUE #115: inconsistency when running Explorer.DataFrame.pivot_wider/4
columns "a" and "b" change order when executing: 
```elixir
iex(1)>  df = Explorer.DataFrame.from_map(%{id: [1, 1], variable: ["a", "b"], value: [1, 2]})
#Explorer.DataFrame<
  [rows: 2, columns: 3]
  id integer [1, 1]
  value integer [1, 2]
  variable string ["a", "b"]
>
iex(2)> Explorer.DataFrame.pivot_wider(df, "variable", "value")  
#Explorer.DataFrame<
  [rows: 1, columns: 3]
  id integer [1]
  a integer [1]
  b integer [2]
>
iex(3)> Explorer.DataFrame.pivot_wider(df, "variable", "value")
#Explorer.DataFrame<
  [rows: 1, columns: 3]
  id integer [1]
  b integer [2]
  a integer [1]
>
iex(4)> Explorer.DataFrame.pivot_wider(df, "variable", "value")
#Explorer.DataFrame<
  [rows: 1, columns: 3]
  id integer [1]
  b integer [2]
  a integer [1]
>
iex(5)> Explorer.DataFrame.pivot_wider(df, "variable", "value")
#Explorer.DataFrame<
  [rows: 1, columns: 3]
  id integer [1]
  a integer [1]
  b integer [2]
>
```

cigrainger:
Bumps `polars` to `0.20.0` and fixes some uncaught bugs in the process :grimacing:!

- Polars was treating literal nulls as empty strings when reading CSV strings
- Rolling min/mean seemed to be adding a 0 value before the first actual value
- I'm not sure what was going on with the dtypes assertion on read_csv but there's no way the Rust-side code for generating the schema was right
- Fixes #115! Appears there's some more stability now.

Delving into lazy eval now so I don't plan on bumping `polars` again until that's done :).

This switches from `shell.nix` to `flake.nix`. I understand the desire to avoid including dev environment info in the repo, but Nix flakes actually act in a packaging capacity as well and I intend to make Explorer available as a flake for others. Flakes also must be checked in to version control.


lib/explorer/data_frame.ex
```diff
@@ -1293,6 +1293,17 @@ defmodule Explorer.DataFrame do
 
   * `id_cols` - A set of columns that uniquely identifies each observation. Defaults to all columns in data except for the columns specified in `names_from` and `values_from`. Typically used when you have redundant variables, i.e. variables whose values are perfectly correlated with existing variables. May accept a filter callback or list of column names.
   * `names_prefix` - String added to the start of every variable name. This is particularly useful if `names_from` is a numeric vector and you want to create syntactic variable names.
+
+  ## Examples
+
+      iex> df = Explorer.DataFrame.from_map(%{id: [1, 1], variable: ["a", "b"], value: [1, 2]})
```
josevalim, line 1299: Somewhat unrelated: another reason to call this function `from_series` is that we could receive any enumerable as a series. This could be beneficial because keywords preserve key ordering but maps do not.

josevalim (approved):

lib/explorer/data_frame.ex
cigrainger: Oh good point. No reason not to take a keyword list here. I'll raise an issue and make the change.

PR #133, opened 2022-03-17
Rename `DF.from_map/1` to `DF.from_columns/1`

philss:
This closes https://github.com/elixir-nx/explorer/issues/131

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -34,7 +34,7 @@ defmodule Explorer.Backend.DataFrame do
 
   # Conversion
 
-  @callback from_map(map()) :: df
+  @callback from_columns(map() | Keyword.t()) :: df
   @callback to_map(df, convert_series? :: boolean()) :: map()
   @callback to_binary(df, header? :: boolean(), delimiter :: String.t()) :: String.t()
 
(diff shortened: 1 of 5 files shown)


cigrainger:
Thanks so much for this @philss! I'm actually liking @josevalim's suggestion to go with `from_columns` and from `from_rows`. I love the keyword list support. Could we go with `from_columns`?

philss:
@cigrainger sure! But I think [he mentioned](https://github.com/elixir-nx/explorer/issues/131#issuecomment-1068853036) that if we are going to support series, it could be `from_series`.

> Another thing to consider is if we also want to support from_series(%{a: some_actual_series, b: another_actual_series}). If we don't, then perhaps it should be called from_cols or from_columns.

@josevalim can you confirm?

Anyway, I think I prefer `from_columns`/`from_cols` :grimacing:

josevalim:
Yeah, if we support series I would go with `from_series` but I am fine either way!

cigrainger:
The reason I'm not keen on `from_series` is that I think it gets confusing because we support lists, but also because one might think we're taking a single series and making it into a dataframe. E.g. if we were to support list columns and you wanted to transpose one into a dataframe.

PR #134, opened 2022-03-17
Fix type casting for numeric series

philss:
When creating a series with a mix of integers and floats, we should have
a series of floats in the end.

This change fixes a small bug when a list like this would give an error:
`[1, 2.0, 3]`.


cigrainger (approved):
🤦 Good catch!

PR #136, opened 2022-03-19
`DF.concat_rows` can cast numeric cols if they differ

philss:
This change makes the behaviour of `Dataframe.concat_rows` be similar to
the one of `Series.from_list` that accepts both integers and floats in
the same series.

Closes https://github.com/elixir-nx/explorer/issues/135.


philss:
It's not beautiful but solves the problem :grimacing:

philss:
I'm closing because I want to simplify this as a suggestion from @josevalim

PR #137, opened 2022-03-20
add `DataFrame.from_rows/2`

### LINKED ISSUE #59: create data frame from records

cigrainger:
WIP. ~Last hurdle is figuring out how to coerce atom keys to string keys without iterating over the whole list of rows Elixir-side. @philss any ideas?~ Figured it out! I'm able to pass a term and then match on the TermType. A bit convoluted, but it works!

I also think the `ExAnyValue` approach may permit us to have a single `s_from_list` without the macro implementation. And I think remove the iteration over the full list Elixir-side. @jonatanklosko @josevalim I believe that came up during our call.

Closes #59.

cc @kimjoaoun

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -35,6 +35,7 @@ defmodule Explorer.Backend.DataFrame do
   # Conversion
 
   @callback from_columns(map() | Keyword.t()) :: df
+  @callback from_rows(list(map() | Keyword.t())) :: df
   @callback to_map(df, convert_series? :: boolean()) :: map()
   @callback to_binary(df, header? :: boolean(), delimiter :: String.t()) :: String.t()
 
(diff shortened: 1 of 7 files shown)


josevalim:
@cigrainger does it mean we can have any value as a column now and they get the type term/any or something like that?!

cigrainger:
@josevalim it means we can pretty arbitrarily support any value that Polars supports without checking/coercing on the Elixir side. The enum with its supported type variants is [here](https://pola-rs.github.io/polars/polars/prelude/enum.AnyValue.html).

josevalim (approved):

PR #138, opened 2022-03-21
interop improvements

cigrainger:
This started off by making things a touch more idiomatic with implementing `From` instead of using ad hoc conversion methods.

Then I found a couple features on Polars that we aren't using so I removed those (hopefully improving compile times and size of the .so). 

Then I got stuck in removing dead code (esp from the dataframe api) in preparation for moving to lazy. Silly to have left things in from the original fork from `ExPolars`.


marknuzz:
Why do the dataframes need to be stored in Rust? Why not a more mature language with more established interop tooling?

josevalim:
Despite their relative young age, the integration between Elixir and Rust is pretty solid. Plus many recent data efforts, such as Apache Arrow, have been centered around Rust.

In any case, backends are pluggable! So you are welcome to explore alternatives and we can discuss their inclusion based on their technical merits. :)

PR #141, opened 2022-03-21
lazy by default

### LINKED ISSUE #54: Implement lazy by default
In a functional language with immutable data, memory management is important. The current implementation utilises `polars`'s eager mode and computes new dataframes for every function. Because the dataframes are represented as a [ResourceArc](https://docs.rs/rustler/0.22.0/rustler/resource/struct.ResourceArc.html), they are only dropped from memory when the GC runs. This can be pretty heavy on memory, to say the least. The most efficient approach would be to treat dataframes as _lazy by default_ with 'peeking' for inspect. In [R, for example](https://www.r-bloggers.com/2018/07/about-lazy-evaluation/), function arguments are only evaluated when they are _needed_ to show output.

An additional benefit to lazy by default is the opportunity to optimise queries. Why evaluate every function call when you can build up a query that may be executed in a more efficient way all together?

Polars has [`polars_lazy`](https://docs.rs/polars-lazy/0.15.1/polars_lazy/index.html) which permits exactly this. Making this shift will then permit the use of lazy evaluation for other backends -- esp. [Datafusion/Ballista](https://github.com/apache/arrow-datafusion) and Ecto.

For Explorer, we'll need to do a bit of exploration (pun absolutely intended) for how we can achieve this while maintaining the flexibility of pluggable backends. And when looking to a pure Elixir backend we should consider whether it's unnecessarily onerous compared to the benefits.

I'd really love ideas and feedback for making Explorer lazy by default. Is there a good peeking mechanism in other libraries? For example, something I'm going to be exploring is how `tibble`s in R minimise computation for `print` and `head`.

cigrainger:
> THIS IS A WORK IN PROGRESS (including this description)

## What?

Shifts the execution model for Polars to [lazy](https://docs.rs/polars/0.20.0/polars/docs/lazy/index.html).

## Why?

Mostly: memory. Since Elixir is a functional language with immutable data, we're currently _duplicating_ the Polars dataframes every time we apply an operation! Rustler releases the unused values, but the GC is not super fast. It's easy to apply many operations in quick succession and multiply the data in memory many times.

Also: speed. Using the lazy execution engine permits Polars to optimize queries.

## How?

<hr>
When complete, closes #54 (finally). And then we can start work on cutting the first Hex release (which will bring Rustler precompiled!).


cigrainger:
Closing this in favour of an approach which treats lazy as a separate backend.

PR #142, opened 2022-03-21
`DF.concat_rows` casts numeric cols if they differ

philss:
This change makes the behaviour of `Dataframe.concat_rows` be similar to
the one of `Series.from_list` that accepts both integers and floats in
the same series.

Closes https://github.com/elixir-nx/explorer/issues/135.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1546,7 +1546,8 @@ defmodule Explorer.DataFrame do
   @doc """
   Combine two or more dataframes row-wise (stack).
 
-  Column names and dtypes must match exactly.
+  Column names and dtypes must match. The only exception is for numeric
+  columns that can be mixed together, and casted automatically to float columns.
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


cigrainger (approved):
LGTM! Thanks @philss. :heart:

philss:
@cigrainger :purple_heart: 
Please don't merge it yet. José may want to take a look :)

lib/explorer/data_frame.ex
```diff
@@ -1558,15 +1559,79 @@ defmodule Explorer.DataFrame do
...
+      iex> Explorer.DataFrame.concat_rows([df1, df2])
+      #Explorer.DataFrame<
+        [rows: 6, columns: 2]
+        x float [1.0, 2.0, 3.0, 4.2, 5.3, "..."]
+        y string ["a", "b", "c", "d", "e", "..."]
+      >
   """
-  def concat_rows([%DataFrame{} = h | t] = dfs) do
-    key = Map.new(Enum.zip(names(h), dtypes(h)))
+  def concat_rows([%DataFrame{} | _t] = dfs) do
```
josevalim, line 1572: I think we can simplify it a bit more. The trick to realize here is that, because all dataframes must have the same name and the compatible types, you can start with the initial shape and only do a single pass to compute the final shape. Something like:

    def compute_changed_types_concat_rows([head | tail]) do
      types = Map.new(Enum.zip(names(head), dtypes(head)))

      Enum.reduce(tail, %{}, fn df, changed_types ->
        if num_columns(df) != map_size(types) do
          raise "columns do not match"
        end

        Enum.reduce(Enum.zip(names(df), dtypes(df)), changed_types, fn {name, type}, changed_types ->
          cond do
            not Map.has_key?(changed_types, name) -> raise "columns do not match"
            types_are_equal?(...) -> changed_types
            types_are_compatible?(...) -> Map.put(changed_types, name, :float)
            true -> raise "incompatible types"
        end)
      end)

josevalim:

PR #143, opened 2022-03-28
Use built-in Elixir modules for date & datetime

philss:
This is an attempt to remove the rust analyzer warning about the map key
with the name of the struct (`__struct__`).

### DIFF
--- a/native/explorer/src/datatypes.rs
+++ b/native/explorer/src/datatypes.rs
@@ -1,7 +1,7 @@
 use chrono::prelude::*;
 use polars::prelude::*;
 use rustler::resource::ResourceArc;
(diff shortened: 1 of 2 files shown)


cigrainger:
Does this work?? I just assumed it wouldn't. I'm not sure why. That's fantastic.

philss:
Yeah, I didn't know either. I was experimenting and it worked in the first try :tada: 
The only restriction is to keep the same field names.

philss:
PS: I also fixed the warnings related to the `derive(NifStruct)`: https://github.com/rusterlium/rustler/pull/441
But we should wait for a new version of Rustler to be released :)

cigrainger:
Amazing. So the only thing I'm worried about with this is if the elixir structs were to change it would mean a major version from Explorer too. But that's not too big of a deal as long as it's documented.

PR #146, opened 2022-03-30
Provide column name in error for `DF.from_columns/2`

philss:
Closes https://github.com/elixir-nx/explorer/issues/69


cigrainger (approved):
Planning on moving this Rust-side, but this absolutely gets us the value in the meantime. Fantastic! Thanks @philss <3

PR #148, opened 2022-03-31
Add "schedule" annotation to NIF functions

philss:
This is needed to classify most of the NIF functions as
possible "dirty".
For smaller sets, those functions should not take more
than 1 millisecond to perform. But for larger datasets,
it's probably going to take more time.

This follows the recommendations from:
https://www.erlang.org/doc/man/erl_nif.html#lengthy_work

Also see https://github.com/rusterlium/rustler/issues/402

### DIFF
--- a/native/explorer/src/dataframe.rs
+++ b/native/explorer/src/dataframe.rs
@@ -32,7 +32,7 @@ macro_rules! df_read_read {
     };
 }
 
-#[rustler::nif]
+#[rustler::nif(schedule = "DirtyIo")]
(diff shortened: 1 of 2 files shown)


cigrainger (approved):
I've been wondering about this and hadn't seen any problems so have been putting it off. Thanks for doing this!

josevalim:
@cigrainger btw, once the backend becomes lazy, you only need those annotations in the functions that do IO operations and the ones that compute. The regular functions do not need to be tagged. :)

PR #151, opened 2022-04-06
Allow a series of `nils` to be created

philss:
This makes the behavior similar to what we see in `py-polars` and
`nodejs-polars`, where a series with only `nil` values can be created
with the dtype of `float`.

This closes https://github.com/elixir-nx/explorer/issues/150

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -30,6 +30,8 @@ defmodule Explorer.Series do
   @type dtype :: :float | :integer | :boolean | :string | :date | :datetime
   @type t :: %Series{data: data, dtype: dtype}
 
+  @valid_dtypes [:float, :integer, :boolean, :string, :date, :datetime]
+
   @enforce_keys [:data, :dtype]
   defstruct [:data, :dtype]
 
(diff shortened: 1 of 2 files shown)


lib/explorer/series.ex
```diff
@@ -108,15 +108,19 @@ defmodule Explorer.Series do
         [1.0, 2.0]
       >
 
+  Trying to create a "nil" series will result in a series of floats.
+  If this is not desired, you can always use `cast/2` to change the type.
+
+      iex> Explorer.Series.from_list([nil, nil])
```
polvalente, line 114: Can't there be an option that already accepts the target type?

polvalente (approved):

lib/explorer/series.ex
philss: Yeah, I think this is a good idea :+1: 
The `PolarsBackend.Series` accept a `dtype` option. We could accept in the main function as well.

cigrainger: Yep totally agree @polvalente!

cigrainger (approved):
Looks great! Thanks @philss!

philss:
I'm going to try to add the option in this PR.

lib/explorer/series.ex
philss: I found this to be harder than I thought :grimacing: 
We would first have to cast the values from series to the desired type before passing to the backend. Or we would have to cast in the backend. (this is the case for setting the `dtype` of series with nils and not nil).

What I think it's easier to do now is to control only the case where the list is "all nil" and we want to specify a dtype for the column. WDYT?

polvalente: Perhaps if the target type is passed, and it's different than the inferred one, we could use `cast/2` internally, for ease of use

philss: Oh, good idea! I changed to do something like that. Please check again.

cigrainger (approved):
Looks great. Thanks @philss!

philss:
@cigrainger yeah, I think so :smiley:

PR #154, opened 2022-04-07
Change `DF.to_map/2` to use string keys by default

philss:
We want to avoid the inconsistency of having the columns of our
dataframes as strings and the map keys as atoms. This way we always have
column names as strings, and if the user wants he/she can configure to
use atoms.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -300,14 +300,27 @@ defmodule Explorer.DataFrame do
 
   By default, the constituent series of the dataframe are converted to Elixir lists.
 
+  ## Options
+
+    * `:convert_series` - Should the series should be converted to a list? (default: `true`)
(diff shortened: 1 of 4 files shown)


philss:
cc @josevalim

lib/explorer/data_frame.ex
```diff
@@ -300,14 +300,27 @@ defmodule Explorer.DataFrame do
 
   By default, the constituent series of the dataframe are converted to Elixir lists.
 
+  ## Options
+
+    * `:convert_series` - Should the series should be converted to a list? (default: `true`)
```
josevalim, line 305:
```suggestion
    * `:convert_series` - Convert the series to lists (default: `true`)
```

josevalim (approved):

cigrainger:
Love it. Thanks @philss!

PR #162, opened 2022-04-13
Use microseconds instead of milliseconds

philss:
This change introduces the usage of microseconds for datetime series in order
to preserve more precision.

Closes https://github.com/elixir-nx/explorer/issues/156


cigrainger (approved):
Looks good! Thanks @philss! <3

PR #163, opened 2022-04-19
[WIP] Better error messages

### LINKED ISSUE #158: `Series.equal` behaviour for arbitrarily sized Series
Hey folks,

I were doing some tests yesterday and realized that - different than I would expect - making comparisons using `Series.equal` for any RHS in which the length is **not** 1 or **exactly the same length** as LHS would raise an error. 

Some context:

## From documentation
```elixir
#Same sized Series -> OK
iex > s1 = Explorer.Series.from_list([1, 2, 3])
iex > s2 = Explorer.Series.from_list([1, 2, 4])
iex > Explorer.Series.equal(s1, s2)
#Explorer.Series<
  boolean[3]
  [true, true, false]
>

#RHS with length 1 -> OK
iex > s = Explorer.Series.from_list([1, 2, 3])
iex > Explorer.Series.equal(s, 1)
#Explorer.Series<
  boolean[3]
  [true, false, false]
>
```
 But if I try to compare with an arbitrarily sized Series, it won't work:
 
 ```elixir
iex > s = Series.from_list([1,2,3])
iex > Series.equal(s, Series.from_list([2,3]))
#Error
    (explorer 0.1.0-dev) Explorer.PolarsBackend.Native.s_eq(shape: (3,)
Series: '' [i64]
[
	1
	2
	3
], shape: (2,)
Series: '' [i64]
[
	2
	3
])
    (explorer 0.1.0-dev) lib/explorer/polars_backend/shared.ex:14: Explorer.PolarsBackend.Shared.apply_native/3
 ```

In R, the comparison occur, but raises a warning.

halian-vilela:
> This is a first approach to handle error messages raised by comparison functions. I've made a first attempt only with `Series.equal/2` because I realized that the job may be much more complex than I've pictured it in the start. So I'll submit for some code review before applying the approach to the other functions.

## What?

As discussed in #158, we need better error messages for comparison functions when applied to Series of some specific lengths.

## Why?

**Developer Experience:** current error messages raised are not clear enough on what's really wrong. Changes proposed here will give a better overview on what's happening when:
1. One of the Series is empty
2. Length of the Series do not match each other

## How?

I've tried to match (as most as I could) the coding style already present in the surrounding functions, so my approach was:
1. To create a new helper function (`dtype_length_error`) to raise the error, which deals with the function that was called and the arguments (it solves the discussion about showing both LHS and RHS in the error message). This function has a clause to deal with empty Series and mismatches
2. To create another helper function (`is_valid_length`) to check the length of the arguments and, by adding it to the first clause, making sure that all the other clauses will never have to deal with empty or mismatched Series. 
3. To add a bunch of new tests to `series_test.exs` to check if all of the above would work (I've added two more examples to the doctest session, but some other cases didn't make much sense as docs, so I opted to add them directly to the dedicated file)

- - - 

**Next steps:**

If the approach is correct, the plan would be to create more tests to check **all** of the element-wise functions and apply the lenght-check + error-raising  to all of those that breaks when there is an empty or mismatched pair. 

When completed it closes #158 and a lot of related similar issues.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -2013,9 +2013,9 @@ defmodule Explorer.Series do
   end
 
   defp dtype_length_error(function, left, right, :length_mismatch) do
-    raise(
-      ArgumentError,
-      "Explorer.Series.#{function} could not compare. One of the series must be length 1 or length of left argument (#{length(left)}) must match the length of right argument (#{length(right)})"
-    )
+    raise ArgumentError,
+          "Explorer.Series.#{function} could not compare. " <>
+            "One of the series must be length 1 or the length of both series should match, "
+            "got #{length(left)} and #{length(right)}"
   end
 end
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1214,6 +1214,20 @@ defmodule Explorer.Series do
         [true, true, false]
       >
 
+      iex> s1 = Explorer.Series.from_list([1, 2, 3])
+      iex> s2 = Explorer.Series.from_list([1])
+      iex> Explorer.Series.equal(s1, s2)
+      #Explorer.Series<
+        boolean[3]
+        [true, false, false]
+      >
+
+      iex> s1 = Explorer.Series.from_list([1, 2, 3])
+      iex> s2 = Explorer.Series.from_list([1,2])
+      iex> Explorer.Series.equal(s1, s2)
+      ** (ArgumentError) Explorer.Series.equal/2 could not compare. One of the series must be length 1 or length of left argument (3) must match the length of right argument (2)
+
+
       iex> s = Explorer.Series.from_list([1, 2, 3])
       iex> Explorer.Series.equal(s, 1)
       #Explorer.Series<
@@ -1253,8 +1267,17 @@ defmodule Explorer.Series do
           left :: Series.t(),
           right :: Series.t() | number() | Date.t() | NaiveDateTime.t() | boolean() | String.t()
         ) :: Series.t()
-  def equal(%Series{dtype: dtype} = left, %Series{dtype: dtype} = right),
-    do: apply_impl(left, :eq, [right])
+
+  def equal(%Series{dtype: _left_dtype} = left, %Series{dtype: _right_dtype} = right) do
+    case is_valid_length(left, right) do
+      {false, nil} -> dtype_length_error("equal/2", left, right, nil)
+      false -> dtype_length_error("equal/2", left, right)
+      true -> apply_impl(left, :eq, [right])
+    end
+  end
+
+  # def equal(%Series{dtype: dtype} = left, %Series{dtype: dtype} = right),
+  #   do: apply_impl(left, :eq, [right])
 
   def equal(%Series{dtype: dtype} = left, right)
       when K.and(dtype in [:integer, :float], is_number(right)),
@@ -1271,6 +1294,15 @@ defmodule Explorer.Series do
 
   def equal(%Series{dtype: :boolean} = left, right) when is_boolean(right),
     do: apply_impl(left, :eq, [right])
+
+  defp is_valid_length(left, right) do
+    case {Series.length(left), Series.length(right)} do
+      {l, r} when K.or(l == 0, r == 0) -> {false, nil}
+      {l, r} when K.or(l == 1, r == 1) -> true
+      {l, r} when l != r -> false
+      {_, _} -> true
+    end
+  end
 
   @doc """
   Returns boolean mask of `left != right`, element-wise.
@@ -1974,4 +2006,20 @@ defmodule Explorer.Series do
         "cannot invoke Explorer.Series.#{function} with mismatched dtypes: #{left_dtype} and " <>
           "#{right_dtype}."
       )
+
+  defp dtype_length_error(function, left, right, zero_length \\ "")
+
+  defp dtype_length_error(function, _left, _right, nil) do
+    raise(
+      ArgumentError,
+      "Explorer.Series.#{function} cannot compare zero length series"
+    )
+  end
+
+  defp dtype_length_error(function, left, right, _zero_length) do
+    raise(
+      ArgumentError,
+      "Explorer.Series.#{function} could not compare. One of the series must be length 1 or length of left argument (#{length(left)}) must match the length of right argument (#{length(right)})"
+    )
+  end
 end
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -2008,10 +2008,8 @@ defmodule Explorer.Series do
       )
 
   defp dtype_length_error(function, _left, _right, :zero_length) do
-    raise(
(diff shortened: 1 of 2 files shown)


lib/explorer/series.ex
```diff
@@ -1272,6 +1295,15 @@ defmodule Explorer.Series do
   def equal(%Series{dtype: :boolean} = left, right) when is_boolean(right),
     do: apply_impl(left, :eq, [right])
 
+  defp is_valid_length(left, right) do
+    case {Series.length(left), Series.length(right)} do
+      {l, r} when K.or(l == 0, r == 0) -> {false, nil}
```
josevalim, line 1300: We should actually be able to compare empty with empty, no?

josevalim: I would say even anything with empty should work too. It just returns false for everything.

halian-vilela: Hey @josevalim, thanks for your feedback!

I should have made a disclaimer about this, I agree that _empty with empty_  and _empty with anything_ should work, the problem is that Polars doesn't seem to know how to deal with it yet.

Take a look at the returns when comparing empty Series directly with the `apply_impl` (to bypass intermediary clauses and send it directly to Polars)

**Note:** `S` is alias for `Explorer.Series`

## Empty with Empty
```elixir
iex > S.apply_impl(S.from_list([]),:eq,[S.from_list([])])     
#Explorer.Series<
  boolean[0]
  []
```  
I think that returning an empty Series for this is suboptimal, shouldn't we expect that when comparing two empties we should have a truthy value? After all, they are indeed equal.

## Empty with other lengths

```elixir
iex > S.apply_impl(S.from_list([1,2,3]),:eq,[S.from_list([])])
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
                                                                     left: `3`,
                                                                                right: `0`', /Users/aurus/.cargo/registry/src/github.com-1ecc6299db9ec823/arrow2-0.10.0/src/compute/comparison/primitive.rs:18:5
    (explorer 0.1.0-dev) Explorer.PolarsBackend.Native.s_eq(shape: (3,)
Series: '' [i64]
[
        1
        2
        3
], shape: (0,)
Series: '' [f64]
[
])
    (explorer 0.1.0-dev) lib/explorer/polars_backend/shared.ex:14: Explorer.PolarsBackend.Shared.apply_native/3
``` 
It just does not handle it well...

- - - 

Because of that, I preferred to raise a user friendly error. But I do agree that all of this should just work. Is that really a Polars issue (I could even open an issue there) or am I missing something?

josevalim:
> I think that returning an empty Series for this is suboptimal, shouldn't we expect that when comparing two empties we should have a truthy value? After all, they are indeed equal.

I disagree. It is not checking if the two series are equal. It is a pairwise comparisons of the elements in the series. So an empty list is indeed correct here. Nx has a similar behaviour.

But i am fine if a comparison of anything with empty raises. But empty with empty should work. :)

halian-vilela:
> I disagree. It is not checking if the two series are equal. It is a pairwise comparisons of the elements in the series. So an empty list is indeed correct here. Nx has a similar behaviour.
> 
> But i am fine if a comparison of anything with empty raises. But empty with empty should work. :)

Nice, you got a point! It should be able to compare both empty now. After thinking a lot about that, I just saw that we also have `all_equal/2` that just behaves as I was expecting... probably that's the best way to go to compare if they are really equal.

Nonetheless, I don't think it's obvious for newcomers and suggested to add a little hint in the documentation, it should now read:

```
Returns boolean mask of `left == right`, element-wise. If both are empty it will return an empty series.
```

Are we ok with that?

## Summary of other small changes in this last commit

1. Changed the flag to raise the zero length error from `nil` to an atom `:zero_length` (it's clearer)
2. Moved the validation function closer to the other helpers
3. Changed the zero length error message to match the new scenario (things cannot be compared **with** zero length series)

lib/explorer/series.ex
```diff
@@ -1202,7 +1202,7 @@ defmodule Explorer.Series do
   # Comparisons
 
   @doc """
-  Returns boolean mask of `left == right`, element-wise.
+  Returns boolean mask of `left == right`, element-wise. If both are empty it will return an empty series.
```
josevalim, line 1205:
```suggestion
  Returns boolean mask of `left == right`, element-wise.

  If both are empty it will return an empty series.
```

lib/explorer/series.ex
```diff
@@ -1253,8 +1267,14 @@ defmodule Explorer.Series do
...
         ) :: Series.t()
-  def equal(%Series{dtype: dtype} = left, %Series{dtype: dtype} = right),
-    do: apply_impl(left, :eq, [right])
+
+  def equal(%Series{dtype: _left_dtype} = left, %Series{dtype: _right_dtype} = right) do
+    case is_valid_length(left, right) do
+      {false, :zero_length} -> dtype_length_error("equal/2", left, right, :zero_length)
+      false -> dtype_length_error("equal/2", left, right)
+      true -> apply_impl(left, :eq, [right])
+    end
```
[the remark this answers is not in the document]
halian-vilela: Nice! I've already refactored it. But it will change the validation logic, hence valid patterns will return false so that invalid could return the reason.

I'm commiting it now.

Thanks for reviewing!

lib/explorer/series.ex
```diff
@@ -1974,4 +2006,18 @@ defmodule Explorer.Series do
         "cannot invoke Explorer.Series.#{function} with mismatched dtypes: #{left_dtype} and " <>
           "#{right_dtype}."
       )
+
+  defp dtype_length_error(function, _left, _right, :zero_length) do
+    raise(
+      ArgumentError,
+      "Explorer.Series.#{function} cannot compare with zero length series"
+    )
```
josevalim, line 2014:
```suggestion
    raise ArgumentError,
          "Explorer.Series.#{function} cannot compare with zero length series"
```

lib/explorer/series.ex
```diff
@@ -1974,4 +2006,18 @@ defmodule Explorer.Series do
...
+  end
+
+  defp dtype_length_error(function, left, right, :length_mismatch) do
+    raise(
+      ArgumentError,
+      "Explorer.Series.#{function} could not compare. One of the series must be length 1 or length of left argument (#{length(left)}) must match the length of right argument (#{length(right)})"
+    )
```
josevalim, line 2021:
```suggestion
    raise ArgumentError,
          "Explorer.Series.#{function} could not compare. " <>
            "One of the series must be length 1 or the length of both series should match, "
            "got #{length(left)} and #{length(right)}"
```

josevalim (approved):
Just two final comments nitpicking the error message. Final question: should we add those same checks to `not_equal`, `less_than`, `and`, `or`, and other pairwise functions?

josevalim (approved):

cigrainger (approved):
Late to the party but I like this a lot. Definitely go ahead with the other ones and let's get some good Elixir error messages.

PR #166, opened 2022-04-20
Column selection by index

### LINKED ISSUE #161: Select DataFrame columns using a range of numbers.
### Problem

I realised that it is currently impossible to select DataFrame columns using a range of values:

```elixir
iex>  Explorer.DataFrame.select(Explorer.Datasets.iris(), [1..3])

** (Protocol.UndefinedError) protocol String.Chars not implemented for 1..2 of type Range (a struct). This protocol is implemented for the following type(s): Atom, BitString, Date, DateTime, Float, Integer, List, NaiveDateTime, Time, URI, Version, Version.Requirement
    (elixir 1.13.0) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.13.0) lib/string/chars.ex:22: String.Chars.to_string/1
    (explorer 0.1.0-dev) lib/explorer/data_frame.ex:1993: Explorer.DataFrame.maybe_raise_column_not_found/2
    (elixir 1.13.0) lib/enum.ex:937: Enum."-each/2-lists^foreach/1-0-"/2
    (explorer 0.1.0-dev) lib/explorer/data_frame.ex:571: Explorer.DataFrame.select/3
```

or...

```elixir
iex > Explorer.DataFrame.select(Explorer.Datasets.iris(), 1..3)
** (Protocol.UndefinedError) protocol String.Chars not implemented for 1..2 of type Range (a struct). This protocol is implemented for the following type(s): Atom, BitString, Date, DateTime, Float, Integer, List, NaiveDateTime, Time, URI, Version, Version.Requirement
    (elixir 1.13.0) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.13.0) lib/string/chars.ex:22: String.Chars.to_string/1
    (explorer 0.1.0-dev) lib/explorer/data_frame.ex:1993: Explorer.DataFrame.maybe_raise_column_not_found/2
    (elixir 1.13.0) lib/enum.ex:937: Enum."-each/2-lists^foreach/1-0-"/2
    (explorer 0.1.0-dev) lib/explorer/data_frame.ex:571: Explorer.DataFrame.select/3
iex(5)> Explorer.Datasets.iris |> Explorer.DataFrame.select(1..2)  
** (FunctionClauseError) no function clause matching in Explorer.DataFrame.select/3    
    
    The following arguments were given to Explorer.DataFrame.select/3:
    
        # 1
        #Explorer.DataFrame<
          [rows: 150, columns: 5]
          sepal_length float [5.1, 4.9, 4.7, 4.6, 5.0, "..."]
          sepal_width float [3.5, 3.0, 3.2, 3.1, 3.6, "..."]
          petal_length float [1.4, 1.4, 1.3, 1.5, 1.4, "..."]
          petal_width float [0.2, 0.2, 0.2, 0.2, 0.2, "..."]
          species string ["Iris-setosa", "Iris-setosa", "Iris-setosa",
           "Iris-setosa", "Iris-setosa", "..."]
        >
    
        # 2
        1..2
    
        # 3
        :keep 
    
    Attempted function clauses (showing 2 out of 2):
    
        def select(df, columns, keep_or_drop) when is_list(columns)
        def select(df, callback, keep_or_drop) when is_function(callback)
    
    (explorer 0.1.0-dev) lib/explorer/data_frame.ex:568: Explorer.DataFrame.select/3
```

### Possibility

I think that a possible implementation is to create a new `select/3` that accepts a range (`1..3`, for example) as the `columns` argument and returns a new DataFrame with the columns which the indexes fall within the range. That might be done using [Polars `select_by_range`](https://docs.rs/polars/latest/polars/frame/struct.DataFrame.html#method.select_by_range) method.

```elixir
Explorer.DataFrame.select(df, %Range{} = columns, keep_or_drop)
```
### Desired result

```elixir
iex > Explorer.Datasets.iris |> Explorer.DataFrame.select(1..3)
#Explorer.DataFrame<
  [rows: 150, columns: 3]
  sepal_length float [5.1, 4.9, 4.7, 4.6, 5.0, "..."]
  sepal_width float [3.5, 3.0, 3.2, 3.1, 3.6, "..."]
  petal_length float [1.4, 1.4, 1.3, 1.5, 1.4, "..."]
>
```

### Other ecosystems:

This feature is currently available in:
R's `dplyr`
Python's `Pandas`
Julia's `DataFrames.jl`

kimjoaoun:
- [x] Column selection by list of integers.
- [x] Column selection by range of integers.
- [x] Support for :drop argument.
- [x] Doctests for :keep
- [x] Doctests for :drop

This pull requests, if approved, closes #161.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -591,8 +591,7 @@ defmodule Explorer.DataFrame do
     columns =
       if Enum.all?(columns, &is_number/1) do
         map_with_names =
-          Enum.zip(0..length(column_names), column_names)
-          |> Enum.into(%{})
+          for {name, idx} <- Enum.with_index(column_names), into: %{}, do: {idx, name}
 
         Enum.map(columns, &Map.fetch!(map_with_names, &1))
       else
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -595,6 +595,8 @@ defmodule Explorer.DataFrame do
           |> Enum.into(%{})
 
         Enum.map(columns, &Map.fetch!(map_with_names, &1))
+      else
+        columns
       end
 
     Enum.each(columns, fn name ->
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -532,6 +532,25 @@ defmodule Explorer.DataFrame do
         a string ["a", "b", "c"]
       >
 
+
+  You can also use a range or a list of integers:
+
+      iex> df = Explorer.DataFrame.from_columns(a: ["a", "b", "c"], b: [1, 2, 3], c: [4, 5, 6])
+      iex> Explorer.DataFrame.select(df, [0, 1])
+      #Explorer.DataFrame<
+        [rows: 3, columns: 2]
+        a string ["a", "b", "c"]
+        b integer [1, 2, 3]
+      >
+
+      iex> df = Explorer.DataFrame.from_columns(a: ["a", "b", "c"], b: [1, 2, 3], c: [4, 5, 6])
+      iex> Explorer.DataFrame.select(df, 0..1)
+      #Explorer.DataFrame<
+        [rows: 3, columns: 2]
+        a string ["a", "b", "c"]
+        b integer [1, 2, 3]
+      >
+
   Or you can use a callback function that takes the dataframe's names as its first argument:
 
       iex> df = Explorer.DataFrame.from_columns(a: ["a", "b", "c"], b: [1, 2, 3])
@@ -556,6 +575,7 @@ defmodule Explorer.DataFrame do
         [rows: 3, columns: 1]
         c integer [4, 5, 6]
       >
+
   """
   @spec select(
           df :: DataFrame.t(),
@@ -579,7 +599,6 @@ defmodule Explorer.DataFrame do
 
   def select(df, columns = %Range{}, keep_or_drop) do
     range = Enum.to_list(columns)
-
     select(df, range, keep_or_drop)
   end
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -586,13 +586,20 @@ defmodule Explorer.DataFrame do
   def select(df, columns, keep_or_drop \\ :keep)
 
   def select(df, columns, keep_or_drop) when is_list(columns) do
-    if Enum.all?(columns, &is_binary/1) do
-      column_names = names(df)
+    column_names = names(df)
 
-      Enum.each(columns, fn name ->
-        maybe_raise_column_not_found(column_names, name)
-      end)
-    end
+    columns =
+      if Enum.all?(columns, &is_number/1) do
+        map_with_names =
+          Enum.zip(0..length(column_names), column_names)
+          |> Enum.into(%{})
+
+        Enum.map(columns, &Map.fetch!(map_with_names, &1))
+      end
+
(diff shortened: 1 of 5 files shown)


native/explorer/src/dataframe.rs
```diff
@@ -413,6 +413,21 @@ pub fn df_select_at_idx(data: ExDataFrame, idx: usize) -> Result<Option<ExSeries
...
 
+#[rustler::nif(schedule = "DirtyCpu")]
+pub fn df_select_at_range(
+    data: ExDataFrame,
+    range: Vec<usize>,
+) -> Result<ExDataFrame, ExplorerError> {
+    df_read!(data, df, {
+        let columns = range
+            .into_iter()
+            .map(|i| df.select_at_idx(i).unwrap().clone())
```
philss, line 424: If you use a `Range`, maybe you can select directly with [`select_by_range`](https://pola-rs.github.io/polars/polars/frame/struct.DataFrame.html#method.select_by_range). I'm not sure if will be faster or use less memory.

Another idea is to use [`get_column_names`](https://pola-rs.github.io/polars/polars/frame/struct.DataFrame.html#method.get_column_names) and select them by the indexes (the  `range` you have today), and after that use [`select`](https://pola-rs.github.io/polars/polars/frame/struct.DataFrame.html#method.select).

kimjoaoun: The problem with using `select_at_range` is that it would break the functionality of selecting with a list of values.

philss: I see, makes sense. My idea was to avoid selecting each series individually. But I'm not sure if this is a problem. Probably not :)

philss: Another idea is to solve the problem entirely in Elixir. You can select the column names based in their position and pass down the names you want to select (in the order of the range or column idxs). This can have the benefit to avoid the need of adding another NIF function. WDYT?

cigrainger: @philss I like that idea a lot. I'm not sure how all of this is going to play into lazy as there isn't a set column index in a lazy frame so it seems like leaning on the names is the best bet.

kimjoaoun: It's a great idea. I'll make the necessary changes!

lib/explorer/data_frame.ex
```diff
@@ -568,13 +588,29 @@ defmodule Explorer.DataFrame do
   def select(df, columns, keep_or_drop) when is_list(columns) do
     column_names = names(df)
 
+    columns =
+      if Enum.all?(columns, &is_number/1) do
+        map_with_names =
+          Enum.zip(0..length(column_names), column_names)
+          |> Enum.into(%{})
```
philss, line 595: Just a minor suggestion: you could use [`Enum.with_index/2`](https://hexdocs.pm/elixir/Enum.html#with_index/2) and the [`for`](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#for/1) comprehension to make this slightly shorter.

```suggestion
        map_with_names =
          for {name, idx} <- Enum.with_index(column_names), into: %{}, do: {idx, name}
```

philss (approved):
LGTM :+1: 
Just a minor suggestion.

cigrainger (approved):
Looks good to me!

As an aside: I think this might be a bit of Polars specificity creeping into the DataFrame API (e.g. we may be able to pass column indices directly to a different backend) but let's cross that bridge when we get there. For now this is great.

PR #168, opened 2022-04-25
update access api to use integers

cigrainger:
As on the tin. Thanks to @kimjoaoun we can use integers for `select` now. The obvious extension for me was to enable access like `df[[1, 2, 3]]` or `df[1..4]` or `df[4]`. This also extends `pull` to accept integers. e.g. `DataFrame.pull(df, 3)`.


kimjoaoun (approved):
LGTM :)

PR #174, opened 2022-04-26
bump rust deps

cigrainger:
This bumps rust dependencies based on the [dependabot security alerts](https://github.com/elixir-nx/explorer/security/dependabot). Not sure but I think this merits cutting 0.1.1?

Fixes https://github.com/elixir-nx/explorer/security/dependabot/1
Does not fix https://github.com/elixir-nx/explorer/security/dependabot/2
Fixes https://github.com/elixir-nx/explorer/security/dependabot/3
Fixes https://github.com/elixir-nx/explorer/security/dependabot/4

Also: adds a changelog.

Edit: added a couple doc fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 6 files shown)


josevalim (commented):
A new release doesn't hurt :) your call

PR #178, opened 2022-04-27
Implement the Table.Reader protocol

### LINKED ISSUE #169: Complement `from_rows` and `from_columns` with `to_rows` and `to_columns`
I think instead of `to_map` we should have `to_rows` (which would yield a list of maps) and `to_columns` which would yield a map of lists. That way the functions have symmetry.

jonatanklosko:
Closes #169.

Integrates Explorer with [`table`](https://hexdocs.pm/table/Table.html), which allows a `DataFrame` to be consumed as rows or as columns.


josevalim (approved):

cigrainger (approved):
Looks good to me!

PR #183, opened 2022-04-28
make nx an optional dependency

### LINKED ISSUE #177: Make Nx an optional dependency
The `keyword!` function is now part of Elixir as `Keyword.validate!`.

For the tensor operations (to_tensor and from_tensor), we can add `@compile {:no_warn_undefined, Nx}`.

cigrainger:
This closes #177.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -32,7 +32,7 @@ defmodule Explorer.MixProject do
   defp deps do
     [
       {:ex_doc, "~> 0.24", only: :docs, runtime: false},
-      {:nx, "~> 0.1.0"},
+      {:nx, "~> 0.1.0", only: :test},
       {:rustler_precompiled, "~> 0.3"},
       {:table, "~> 0.1.0"},
       {:table_rex, "~> 3.1.1"}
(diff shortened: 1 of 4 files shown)


mix.exs
```diff
@@ -32,7 +32,7 @@ defmodule Explorer.MixProject do
   defp deps do
     [
       {:ex_doc, "~> 0.24", only: :docs, runtime: false},
-      {:nx, "~> 0.1.0"},
+      {:nx, "~> 0.1.0", only: :test},
```
josevalim, line 35:
```suggestion
      {:nx, "~> 0.1.0", optional: true},
```

polvalente: Should be 0.2.0 actually, right?

cigrainger: TIL!

josevalim:
```suggestion
      {:nx, "~> 0.1", optional: true},
```

It is unlikely the API used by Explorer will ever actually change/break.

josevalim:
@cigrainger you also need to change the mix.exs to say it requires Elixir ~> 1.13. CI already seems to use 1.13.

josevalim (approved):

PR #184, opened 2022-04-28
Make Rustler dependency optional

philss:
This will make the installation faster for users that don't want or
don't need to compile the NIF from source.

In case the user needs, he/she can add the dependency as optional
in their application as well.

For further details, see https://hexdocs.pm/rustler_precompiled/RustlerPrecompiled.html


cigrainger (approved):
Oh that's fantastic! 😁

PR #185, opened 2022-04-28
rename to_map to to_columns and add to_rows

### LINKED ISSUE #169: Complement `from_rows` and `from_columns` with `to_rows` and `to_columns`

cigrainger:
Closes #169. I'm not sure if `to_rows` is the most efficient possible implementation, but row-wise access or iteration is discouraged in `polars` and presents a challenge for types so I went with this.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -456,6 +456,30 @@ defmodule Explorer.DataFrame do
   end
 
   @doc """
+  Converts a dataframe to a list of maps (rows).
+
+  ## Options
+
+    * `:atom_keys` - Configure if the resultant maps should have atom keys. (default: `false`)
+
+  ## Examples
+
+      iex> df = Explorer.DataFrame.from_columns(floats: [1.0, 2.0], ints: [1, nil])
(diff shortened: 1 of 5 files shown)


josevalim (approved):

lib/explorer/data_frame.ex
```diff
@@ -441,18 +441,42 @@ defmodule Explorer.DataFrame do
...
+  def to_columns(df, opts \\ []) do
     opts = Keyword.validate!(opts, convert_series: true, atom_keys: false)
 
-    apply_impl(df, :to_map, [opts[:convert_series], opts[:atom_keys]])
+    apply_impl(df, :to_columns, [opts[:convert_series], opts[:atom_keys]])
+  end
+
+  @doc """
+  Converts a dataframe to a list of maps (rows).
+
```
josevalim, line 460: Perhaps it is worth adding a disclaimer that this is potentially an expensive operation because Explorer often keeps data in a columnar format?

PR #192, opened 2022-04-30
Do fewer passes on cols while inspecting

josevalim:


cigrainger (approved):
🎉🎉🎉🎉🎉💕

cigrainger:
Oh this is probably a silly question @josevalim but how do we get rid of the quotation marks around the ellipses?

josevalim:
In `opts` there is a `:limit` field. Set it to 5 and let container_doc add the `...` for you. :)

josevalim:
Ah, you will also need to slice 6 elements.

PR #200, opened 2022-05-02
mirror nx for cumulative and window functions

### LINKED ISSUE #189: Mirror names with Nx API
There are some Series APIs that are slightly different to Nx:

* `cum_min` and friends are called `cumulative_min` in Nx
* `rolling_min` and friends are called `window_min` in Nx

Should we aim to unify those? I am open to changing Nx if desired too or staying different. :)

cigrainger:
Closes #189

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1738,7 +1738,7 @@ defmodule Explorer.Series do
   """
   def count(series), do: apply_impl(series, :count)
 
-  # Rolling
+  # Window
 
   @doc """
   Calculate the rolling sum, given a window size and optional list of weights.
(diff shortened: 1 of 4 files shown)


lib/explorer/series.ex
```diff
@@ -924,28 +924,30 @@ defmodule Explorer.Series do
...
-  def cum_max(series, reverse? \\ false)
+  @spec cumulative_max(series :: Series.t(), opts :: Keyword.t()) :: Series.t()
+  def cumulative_max(series, opts \\ [])
 
-  def cum_max(%Series{dtype: dtype} = series, reverse?)
-      when dtype in [:integer, :float, :date, :datetime],
-      do: apply_impl(series, :cum_max, [reverse?])
+  def cumulative_max(%Series{dtype: dtype} = series, opts)
+      when dtype in [:integer, :float, :date, :datetime] do
+    opts = Keyword.validate!(opts, reverse?: false)
```
josevalim, line 945: I don't think we use question marks for options (although perhaps we should have):

```suggestion
    opts = Keyword.validate!(opts, reverse: false)
```

josevalim (commented):
Just one tiny comment and ship it!!!

PR #201, opened 2022-05-02
Improve columns support and validations

philss:
This is change to reduce duplication, add flexibility and improve
validation of columns when doing operations with a data frame.

This is:

- adding some guards to better detect columns and column pairs
- adding helpers to always convert columns to an expected "string"
- removing conversions for specific functions
- adding support for selecting columns by atom names


philss:
cc @josevalim

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,110 @@ defmodule Explorer.DataFrame do
   @enforce_keys [:data, :groups]
   defstruct [:data, :groups]
 
+  @type column :: String.t() | non_neg_integer() | atom()
+  @type columns :: [column] | Range.t()
+  @type columns_pairs(other) :: [{column(), other}] | %{column() => other}
```
cigrainger, line 109:
```suggestion
  @type column_pairs(other) :: [{column(), other}] | %{column() => other}
```

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,110 @@ defmodule Explorer.DataFrame do
...
+  end
+
+  defp column_index_map(names),
+    do: for({name, idx} <- Enum.with_index(names), into: %{}, do: {idx, name})
+
+  # Normalize column names without verifying if columns exist
+  defp normalize_columns(df, columns),
+    do: for(column <- Enum.to_list(columns), do: to_column(df, column))
+
+  # Normalize column names and raise if column doesn't not exist
```
cigrainger, line 156:
```suggestion
  # Normalize column names and raise if column does not exist
```

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,110 @@ defmodule Explorer.DataFrame do
...
+  # Guards and helpers for columns
+
+  defguard is_column(column) when is_binary(column) or is_atom(column) or is_integer(column)
+  defguard is_column_pairs(columns) when is_list(columns) or is_map(columns)
+
+  defp to_column(_, column) when is_atom(column), do: Atom.to_string(column)
+  defp to_column(_, column) when is_binary(column), do: column
+  defp to_column(df, column) when is_integer(column), do: to_column!(df, column)
+
+  defp to_column!(df, column) when is_atom(column), do: to_column!(df, Atom.to_string(column))
```
cigrainger, line 122: From an idiomatic standpoint, I'm not understanding the reasoning for bang and non-bang versions of this here. Asking so I can learn :).

cigrainger (commented):
This reads so much better. :tada: Just a couple very minor things.

lib/explorer/data_frame.ex
philss: I should have added a comment about that :grimacing: 
The difference is that the bang version will raise an exception if the column is not present. But the problem here is that the version without a bang that reads an integer column also raises. I left this way because I can´t guarantee that an integer column will ever be valid to be converted. Not sure if this is going to be confusing :/

lib/explorer/data_frame.ex
```diff
@@ -423,7 +494,7 @@ defmodule Explorer.DataFrame do
       ** (ArgumentError) cannot create series "ints": cannot make a series from mismatched types - the value "wrong" does not match inferred dtype integer
   """
   @doc type: :single
-  @spec from_columns(series :: map() | Keyword.t(), opts :: Keyword.t()) :: DataFrame.t()
+  @spec from_columns(series :: columns_pairs(list()), opts :: Keyword.t()) :: DataFrame.t()
```
philss, line 497:
```suggestion
  @spec from_columns(series :: column_pairs(list()), opts :: Keyword.t()) :: DataFrame.t()
```

lib/explorer/data_frame.ex
```diff
@@ -471,7 +542,7 @@ defmodule Explorer.DataFrame do
       >
   """
   @doc type: :single
-  @spec from_rows(rows :: list(map()) | Keyword.t(), opts :: Keyword.t()) :: DataFrame.t()
+  @spec from_rows(rows :: list(columns_pairs(any())), opts :: Keyword.t()) :: DataFrame.t()
```
philss, line 545:
```suggestion
  @spec from_rows(rows :: list(column_pairs(any())), opts :: Keyword.t()) :: DataFrame.t()
```

lib/explorer/data_frame.ex
```diff
@@ -932,32 +992,14 @@ defmodule Explorer.DataFrame do
       >
   """
   @doc type: :single
-  @spec mutate(df :: DataFrame.t(), with_columns :: map() | Keyword.t()) :: DataFrame.t()
-  def mutate(df, with_columns) when is_map(with_columns) do
-    with_columns = Enum.reduce(with_columns, %{}, &mutate_reducer(&1, &2, df))
+  @spec mutate(df :: DataFrame.t(), columns :: columns_pairs(any())) ::
```
philss, line 995:
```suggestion
  @spec mutate(df :: DataFrame.t(), columns :: column_pairs(any())) ::
```

lib/explorer/data_frame.ex
```diff
@@ -1187,55 +1207,34 @@ defmodule Explorer.DataFrame do
...
-    case Keyword.keyword?(names) do
-      false ->
-        names =
-          Enum.map(names, fn
-            name when is_atom(name) -> Atom.to_string(name)
-            name -> name
-          end)
-
-        check_new_names_length(df, names)
+  @spec rename(df :: DataFrame.t(), names :: columns() | columns_pairs(column()) | function()) ::
```
philss, line 1210:
```suggestion
  @spec rename(df :: DataFrame.t(), names :: columns() | column_pairs(column()) | function()) ::
```

lib/explorer/data_frame.ex
cigrainger: Makes sense to me! 🙏

lib/explorer/data_frame.ex
```diff
@@ -1088,19 +1147,13 @@ defmodule Explorer.DataFrame do
   def distinct(df, opts) do
     opts = Keyword.validate!(opts, columns: [], keep_all?: false)
```
josevalim, line 1148: The default for `columns:` must be nil. This is to avoid a similar issue as the one we detected in `drop_nil` where empty list is "ambiguous".

lib/explorer/data_frame.ex
```diff
@@ -1120,20 +1173,16 @@ defmodule Explorer.DataFrame do
       >
   """
   @doc type: :single
-  @spec drop_nil(df :: DataFrame.t(), columns_or_column :: [String.t()] | String.t()) ::
+  @spec drop_nil(df :: DataFrame.t(), columns_or_column :: :all | column() | columns()) ::
           DataFrame.t()
-  def drop_nil(df, columns_or_column \\ [])
+  def drop_nil(df, columns_or_column \\ :all)
```
josevalim, line 1178: Remember to test this change (i.e. an empty list does not drop nothing).

lib/explorer/data_frame.ex
```diff
@@ -1187,55 +1236,37 @@ defmodule Explorer.DataFrame do
...
+  @spec rename(
+          df :: DataFrame.t(),
+          names :: column_names() | column_pairs(column_name()) | function()
+        ) ::
+          DataFrame.t()
+  def rename(df, [name | _] = names) when is_column_name(name) do
+    new_names = to_column_names(names)
 
-        apply_impl(df, :rename, [names])
+    check_new_names_length(df, new_names)
```
josevalim, line 1247:
```suggestion
    check_new_names_length!(df, new_names)
```

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,136 @@ defmodule Explorer.DataFrame do
...
+  end
+
+  defp column_index_map(names),
+    do: for({name, idx} <- Enum.with_index(names), into: %{}, do: {idx, name})
+
+  # Normalize column names without verifying if columns exist.
+  defp to_column_names(names),
+    do: for(name <- Enum.to_list(names), do: to_column_name(name))
+
+  # Normalize column names and raise if column does not exist
```
josevalim, line 165:
```suggestion
  # Normalize column names and raise if column does not exist.
```

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,136 @@ defmodule Explorer.DataFrame do
...
+    |> Enum.to_list()
+    |> Enum.map_reduce(nil, fn
+      column, maybe_map when is_integer(column) ->
+        map = maybe_map || column_index_map(existing_columns)
+
+        existing_column =
+          case Map.fetch(map, column) do
+            {:ok, column} -> column
+            :error -> raise ArgumentError, "column does not exist at index #{column}"
+          end
```
josevalim, line 179: Encapsulate this in a function to remove the duplication with `to_column_pairs`. :)

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,136 @@ defmodule Explorer.DataFrame do
...
   end
 
-  def fetch(df, column) when is_binary(column) or is_integer(column) do
+  def fetch(df, column) when is_column(column) do
     {:ok, pull(df, column)}
   end
 
   @impl true
-  def pop(df, column) when is_binary(column) or is_integer(column) do
+  def pop(df, column) when is_column(column) do
```
josevalim, line 217: We should also support ranges on pop for consistency with fetch. I think we can likely normalize all of these functions.

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,136 @@ defmodule Explorer.DataFrame do
...
+  # Normalize column names without verifying if columns exist.
+  defp to_column_names(names),
+    do: for(name <- Enum.to_list(names), do: to_column_name(name))
+
+  # Normalize column names and raise if column does not exist
+  defp to_existing_columns(df, columns) do
+    existing_columns = names(df)
+
+    columns
+    |> Enum.to_list()
```
josevalim, line 170:
```suggestion
```

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,136 @@ defmodule Explorer.DataFrame do
...
+    end)
+    |> then(fn {pairs, _} -> pairs end)
+  end
+
+  defp column_index_map(names),
+    do: for({name, idx} <- Enum.with_index(names), into: %{}, do: {idx, name})
+
+  # Normalize column names without verifying if columns exist.
+  defp to_column_names(names),
+    do: for(name <- Enum.to_list(names), do: to_column_name(name))
```
josevalim, line 163:
```suggestion
  defp to_column_names(names) when is_list(names),
    do: for(name <- names, do: to_column_name(name))
```

lib/explorer/data_frame.ex
josevalim: Although I am also thinking about adding a guard to `is_list(columns)` so we force ranges to be handled upstream.

lib/explorer/data_frame.ex
```diff
@@ -1120,20 +1173,16 @@ defmodule Explorer.DataFrame do
...
+  def drop_nil(df, column) when is_column(column), do: drop_nil(df, [column])
 
-  def drop_nil(df, columns) when is_list(columns) do
-    column_names = names(df)
-
-    Enum.each(columns, fn name ->
-      maybe_raise_column_not_found(column_names, name)
-    end)
+  def drop_nil(df, [column | _] = columns) when is_column(column) do
+    columns = to_existing_columns(df, columns)
```
josevalim, line 1185:
```suggestion
  def drop_nil(df, columns) do
    columns = to_existing_columns(df, columns)
```

lib/explorer/data_frame.ex
```diff
@@ -1365,7 +1398,8 @@ defmodule Explorer.DataFrame do
       >
   """
   @doc type: :single
-  def dummies(df, columns), do: apply_impl(df, :dummies, [columns])
+  def dummies(df, [column | _] = columns) when is_column(column),
+    do: apply_impl(df, :dummies, [to_existing_columns(df, columns)])
```
josevalim, line 1402:
```suggestion
  def dummies(df, columns),
    do: apply_impl(df, :dummies, [to_existing_columns(df, columns)])
```

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,136 @@ defmodule Explorer.DataFrame do
...
+  def fetch(df, %Range{} = columns) do
+    columns = to_existing_columns(df, columns)
+
     {:ok, select(df, columns)}
   end
 
-  def fetch(df, column) when is_binary(column) or is_integer(column) do
+  def fetch(df, column) when is_column(column) do
     {:ok, pull(df, column)}
   end
```
josevalim, line 214:
```elixir
  def fetch(df, column) when is_column(column) do
    {:ok, pull(df, column)}
  end

  def fetch(df, columns) do
    {:ok, select(df, to_existing_columns(columns)}
  end
```

The same for `pop` and `select`.

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,139 @@ defmodule Explorer.DataFrame do
...
+    end)
+    |> then(fn {pairs, _} -> pairs end)
+  end
+
+  defp fetch_column_at!(map, column) do
+    case Map.fetch(map, column) do
+      {:ok, column} -> column
+      :error -> raise ArgumentError, "column does not exist at index #{column}"
+    end
+  end
```
josevalim, line 159:
```suggestion
  defp fetch_column_at!(map, index) do
    case map do
      %{^index => column} -> column
      %{} -> raise ArgumentError, "no column exists at index #{index}"
    end
  end
```

lib/explorer/data_frame.ex
```diff
@@ -1086,20 +1148,23 @@ defmodule Explorer.DataFrame do
...
 
-    Enum.each(columns, fn name ->
-      maybe_raise_column_not_found(column_names, name)
-    end)
+        callback when is_function(callback) ->
+          Enum.filter(names(df), callback)
+
+        [] ->
+          raise ArgumentError,
+                "you must provide at least one column or omit the column option to select all columns"
```
josevalim, line 1163: I am not a big fan of raising on empty lists. I think it is a semantically valid value. Can we instead return the dataframe as is?

lib/explorer/data_frame.ex
```diff
@@ -1309,28 +1370,30 @@ defmodule Explorer.DataFrame do
...
     |> then(&rename(df, &1))
   end
 
   def rename_with(df, callback, columns) when is_function(callback) and is_function(columns) do
     case df |> names() |> Enum.filter(columns) do
-      [_ | _] = columns ->
+      [column | _] = columns when is_column(column) ->
         rename_with(df, callback, columns)
 
       [] ->
```
josevalim, line 1399: Same here, I would not raise if there is nothing to rename. Just return the dataframe as is!

lib/explorer/data_frame.ex
josevalim: Note I have slightly changed the error message. :)

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,139 @@ defmodule Explorer.DataFrame do
...
+
+  defp to_existing_columns(df, %Range{} = columns) do
+    existing_columns = names(df)
+
+    Enum.map(columns, fn idx ->
+      case Enum.at(existing_columns, idx) do
+        nil -> raise ArgumentError, "column does not exist at index #{idx}"
+        column -> column
+      end
+    end)
```
josevalim, line 201: Let's call `column_index_map(existing_columns)` and fetch those values from the index instead!

josevalim (commented):
I have dropped only two final but minor comments. Then go ahead and merge it (you should have commit access!).

lib/explorer/data_frame.ex
```diff
@@ -104,39 +104,139 @@ defmodule Explorer.DataFrame do
...
+  end
```
josevalim, line 202:
```suggestion
  defp to_existing_columns(df, %Range{} = columns) do
    Enum.slice(names(df), columns)
  end
```

And this will actually put us half way to supporting #204.

PR #202, opened 2022-05-03
Add lazy backend

### LINKED ISSUE #195: Add lazy backend
- [x] Add `Explorer.PolarsBackend.LazyDataFrame` backed by `ExLazyFrame`
- [x] Add `lazy/0` (returns the lazy version of the current backend - noop if already lazy)
- [x] Add `to_lazy/1` (returns the lazy version of the current dataframe - noop if already lazy)
- [x] Add `collect/1` (returns an eager version of the current dataframe - noop if already eager)
- [x] Raise for unsupported functions
- [x] Implement sufficient functions to inspect

cigrainger:
Closes #195 when complete.

This adds a minimal lazy backend for Polars and adds some basic functionality to build on.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -502,6 +502,22 @@ defmodule Explorer.DataFrame do
   ## Conversion
 
   @doc """
+  Converts the dataframe to the lazy version of the current backend.
+
+  If already lazy, this is a noop.
+  """
+  @spec to_lazy(df :: DataFrame.t()) :: DataFrame.t()
+  def to_lazy(df), do: Shared.apply_impl(df, :to_lazy)
+
+  @doc """
+  This collects the lazy data frame into an eager one, computing the query.
+
+  If already eager, this is a noop.
+  """
+  @spec collect(df :: DataFrame.t()) :: {:ok, DataFrame.t()} | {:error, term()}
+  def collect(df), do: Shared.apply_impl(df, :collect)
+
+  @doc """
   Creates a new dataframe.
 
   Accepts any tabular data adhering to the `Table.Reader` protocol, as well as a map or a keyword list with series.
--- a/lib/explorer/polars_backend/data_frame.ex
+++ b/lib/explorer/polars_backend/data_frame.ex
@@ -156,7 +156,7 @@ defmodule Explorer.PolarsBackend.DataFrame do
   def lazy, do: Explorer.PolarsBackend.LazyDataFrame
 
   @impl true
-  def to_lazy(df), do: df |> Shared.apply_native(:df_to_lazy) |> Shared.update_dataframe(df)
+  def to_lazy(df), do: Shared.apply_native(df, :df_to_lazy)
 
   @impl true
   def collect(df), do: df
--- a/lib/explorer/polars_backend/data_frame.ex
+++ b/lib/explorer/polars_backend/data_frame.ex
@@ -153,6 +153,9 @@ defmodule Explorer.PolarsBackend.DataFrame do
   # Conversion
 
   @impl true
+  def lazy, do: Explorer.PolarsBackend.LazyDataFrame
+
+  @impl true
   def from_tabular(tabular) do
     {columns, %{columns: keys}} = Table.to_columns_with_info(tabular)
 
--- a/lib/explorer/polars_backend/data_frame.ex
+++ b/lib/explorer/polars_backend/data_frame.ex
@@ -159,7 +159,7 @@ defmodule Explorer.PolarsBackend.DataFrame do
   def to_lazy(df), do: df |> Shared.apply_native(:df_to_lazy) |> Shared.update_dataframe(df)
 
   @impl true
-  def collect(df), do: {:ok, df}
+  def collect(df), do: df
 
   @impl true
   def from_tabular(tabular) do
--- a/lib/explorer/polars_backend/data_frame.ex
+++ b/lib/explorer/polars_backend/data_frame.ex
@@ -156,6 +156,12 @@ defmodule Explorer.PolarsBackend.DataFrame do
   def lazy, do: Explorer.PolarsBackend.LazyDataFrame
 
   @impl true
+  def to_lazy(df), do: df |> Shared.apply_native(:df_to_lazy) |> Shared.update_dataframe(df)
+
+  @impl true
+  def collect(df), do: {:ok, df}
+
+  @impl true
   def from_tabular(tabular) do
     {columns, %{columns: keys}} = Table.to_columns_with_info(tabular)
 
--- a/lib/explorer/polars_backend/lazy_data_frame.ex
+++ b/lib/explorer/polars_backend/lazy_data_frame.ex
@@ -53,9 +53,6 @@ defmodule Explorer.PolarsBackend.LazyDataFrame do
   def select(ldf, columns, :drop) when is_list(columns),
     do: Shared.apply_native(ldf, :lf_drop, [columns])
 
-  @impl true
-  def pull(ldf, name), do: Shared.apply_native(ldf, :lf_pull, [name])
-
   # Groups
 
   @impl true
--- a/lib/explorer/polars_backend/lazy_data_frame.ex
+++ b/lib/explorer/polars_backend/lazy_data_frame.ex
@@ -7,7 +7,7 @@ defmodule Explorer.PolarsBackend.LazyDataFrame do
 
   defstruct resource: nil, reference: nil
 
-  @behaviour Explorer.Backend.DataFrame
+  use Explorer.Backend.DataFrame, backend: "LazyPolars"
 
   # Conversion
 
@@ -20,12 +20,25 @@ defmodule Explorer.PolarsBackend.LazyDataFrame do
   @impl true
   def collect(ldf), do: Shared.apply_native(ldf, :lf_collect)
 
+  # Introspection
+
+  @impl true
+  def names(ldf), do: Shared.apply_native(ldf, :lf_names)
+
+  @impl true
+  def shape(ldf), do: {nil, ldf |> names() |> length()}
+
+  # Single table verbs
+
+  @impl true
+  def pull(ldf, name), do: Shared.apply_native(ldf, :lf_pull, [name])
+
   # TODO: Make the functions of non-implemented functions
   # explicit once the lazy interface is ready.
   funs =
     Explorer.Backend.DataFrame.behaviour_info(:callbacks) --
       (Explorer.Backend.DataFrame.behaviour_info(:optional_callbacks) ++
-         Module.definitions_in(__MODULE__, :def))
+         Module.definitions_in(__MODULE__, :def) ++ [{:inspect, 2}])
 
   for {fun, arity} <- funs do
     args = Macro.generate_arguments(arity, __MODULE__)
--- a/lib/explorer/polars_backend/lazy_data_frame.ex
+++ b/lib/explorer/polars_backend/lazy_data_frame.ex
@@ -33,7 +33,10 @@ defmodule Explorer.PolarsBackend.LazyDataFrame do
       |> Enum.map(&Shared.normalise_dtype/1)
 
   @impl true
-  def shape(ldf), do: {nil, ldf |> names() |> length()}
+  def shape(ldf), do: {nil, n_columns(ldf)}
+
+  @impl true
+  def n_columns(ldf), do: ldf |> names() |> length()
 
   # Single table verbs
 
--- a/lib/explorer/polars_backend/lazy_data_frame.ex
+++ b/lib/explorer/polars_backend/lazy_data_frame.ex
@@ -48,6 +48,13 @@ defmodule Explorer.PolarsBackend.LazyDataFrame do
   def tail(ldf, rows), do: Shared.apply_native(ldf, :lf_tail, [rows])
 
   @impl true
+  def select(ldf, columns, :keep) when is_list(columns),
+    do: Shared.apply_native(ldf, :lf_select, [columns])
+
+  def select(ldf, columns, :drop) when is_list(columns),
+    do: Shared.apply_native(ldf, :lf_drop, [columns])
+
+  @impl true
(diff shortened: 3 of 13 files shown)


lib/explorer/data_frame.ex
```diff
@@ -386,6 +386,22 @@ defmodule Explorer.DataFrame do
...
+  @spec to_lazy(df :: DataFrame.t()) :: DataFrame.t()
+  def to_lazy(df), do: apply_impl(df, :to_lazy)
+
+  @doc """
+  This collects the lazy data frame into an eager one, computing the query.
+
+  If already eager, this is a noop.
+  """
+  @spec collect(df :: DataFrame.t()) :: {:ok, DataFrame.t()} | {:error, term()}
+  def collect(df), do: apply_impl(df, :collect)
```
josevalim, line 403: We should definitely have a bang variant then, so it flows nicer in the API. Although I wonder if we need the tuple version in the first place? I would think any failure while collecting would be a hard exception.

lib/explorer/polars_backend/lazy_data_frame.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  @behaviour Explorer.Backend.DataFrame
+
+  # Conversion
+
+  @impl true
+  def lazy, do: __MODULE__
+
+  @impl true
+  def to_lazy(ldf), do: Shared.to_dataframe(ldf)
```
josevalim, line 19: A lazy dataframe to lazy should return itself!

```suggestion
  def to_lazy(ldf), do: Shared.to_dataframe(ldf)
```

lib/explorer/polars_backend/lazy_data_frame.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  @impl true
+  def collect(ldf) do
+    case Native.lf_collect(ldf) do
+      {:ok, df} -> Shared.to_dataframe(df)
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  funs =
```
josevalim, line 29:
```suggestion
  # TODO: Make the functions of non-implemented functions
  # explicit once the lazy interface is ready.
  funs =
```

lib/explorer/polars_backend/shared.ex
```diff
@@ -26,8 +27,9 @@ defmodule Explorer.PolarsBackend.Shared do
   def to_dataframe(df, groups \\ [])
   def to_dataframe(%DataFrame{} = df, _groups), do: df
 
-  def to_dataframe(%PolarsDataFrame{} = polars_df, groups),
-    do: %DataFrame{data: polars_df, groups: groups}
+  def to_dataframe(%module{} = polars_df, groups)
+      when module in [PolarsDataFrame, PolarsLazyFrame],
+      do: %DataFrame{data: polars_df, groups: groups}
```
josevalim, line 32: At the moment I am thinking we want don't want this change, so we are forced to think more explicitly where these boundaries cross.

cigrainger: This `to_dataframe` is saying it's making it an `Explorer.DataFrame.t()` by putting the native struct inside the data container. It's not actually making the lazy dataframe eager.

lib/explorer/data_frame.ex
cigrainger: I think you're right that it should be a hard exception. Good point.

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -160,6 +160,15 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
   # Conversion
 
+  @impl true
+  def lazy, do: Explorer.PolarsBackend.LazyDataFrame
+
+  @impl true
+  def to_lazy(df), do: df |> Shared.apply_native(:df_to_lazy) |> Shared.to_dataframe()
+
+  @impl true
+  def collect(df), do: {:ok, df}
```
josevalim, line 170:
```suggestion
  def collect(df), do: df
```

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -160,6 +160,15 @@ defmodule Explorer.PolarsBackend.DataFrame do
 
   # Conversion
 
+  @impl true
+  def lazy, do: Explorer.PolarsBackend.LazyDataFrame
+
+  @impl true
+  def to_lazy(df), do: df |> Shared.apply_native(:df_to_lazy) |> Shared.to_dataframe()
```
josevalim, line 167: This is losing the groups information, as we talked. :) I created #207 as a follow up.

native/explorer/src/lazyframe.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+#[rustler::nif(schedule = "DirtyCpu")]
+pub fn lf_dtypes(data: ExLazyFrame) -> Result<Vec<String>, ExplorerError> {
+    let lf = &data.resource.0;
+    Ok(lf
+        .schema()
+        .iter_dtypes()
+        .map(|dtype| dtype.to_string())
+        .collect())
```
josevalim, line 58: I have a question: how expensive are the names and dtypes operations? Does LazyPolars always keep it readily available? Or does it actually have to execute some of the computations?

native/explorer/src/lazyframe.rs
```diff
@@ -0,0 +1,71 @@
...
+pub fn lf_select(data: ExLazyFrame, columns: Vec<&str>) -> Result<ExLazyFrame, ExplorerError> {
+    let lf = &data.resource.0.clone().select(&[cols(columns)]);
+    Ok(ExLazyFrame::new(lf.clone()))
+}
+
+#[rustler::nif(schedule = "DirtyCpu")]
+pub fn lf_drop(data: ExLazyFrame, columns: Vec<&str>) -> Result<ExLazyFrame, ExplorerError> {
+    let lf = &data.resource.0.clone().select(&[col("*").exclude(columns)]);
+    Ok(ExLazyFrame::new(lf.clone()))
+}
```
josevalim, line 71: Btw, if those operations are very fast (it seems to be the case), they don't need to be dirty nif. I would expect most lazy operations to be fast, except the collect ones. :)

I would also expect the names/dtypes operations in regular data frames to not need to be dirty. Anything akin to a "property read" should be fine.

native/explorer/src/lazyframe.rs
cigrainger: Fairly inexpensive. They use the logical plan, which has to be built, but is fairly instantaneous.

native/explorer/src/lazyframe.rs
cigrainger: Good point -- that's a holdover from copy-paste :).

lib/explorer/polars_backend/lazy_data_frame.ex
```diff
@@ -0,0 +1,87 @@
...
+
+  @impl true
+  def select(ldf, columns, :keep) when is_list(columns),
+    do: Shared.apply_native(ldf, :lf_select, [columns])
+
+  def select(ldf, columns, :drop) when is_list(columns),
+    do: Shared.apply_native(ldf, :lf_drop, [columns])
+
+  @impl true
+  def pull(ldf, name), do: Shared.apply_native(ldf, :lf_pull, [name])
```
josevalim, line 58: I feel like we don't  want this? We always want to collect before we pull?

lib/explorer/data_frame.ex
```diff
@@ -501,6 +501,22 @@ defmodule Explorer.DataFrame do
...
+  def to_lazy(df), do: Shared.apply_impl(df, :to_lazy)
+
+  @doc """
+  This collects the lazy data frame into an eager one, computing the query.
+
+  If already eager, this is a noop.
+  """
+  @spec collect(df :: DataFrame.t()) :: {:ok, DataFrame.t()} | {:error, term()}
+  def collect(df), do: Shared.apply_impl(df, :collect)
+
```
josevalim, line 519: We also want to add `def lazy()`, that simply returns the lazy version of the current backend. :)

test/explorer/data_frame_test.exs
```diff
@@ -699,4 +699,69 @@ defmodule Explorer.DataFrameTest do
     assert Enum.to_list(columns["x"]) == [1, 2, 3]
     assert Enum.to_list(columns["y"]) == ["a", "b", "c"]
   end
+
+  describe "lazy" do
```
josevalim, line 703: I suggest moving this to a separate module because we can easily be looking at a 2kLOC in the future. :)

lib/explorer/polars_backend/lazy_data_frame.ex
cigrainger: It depends on how we define `pull`. I can be convinced either way. `pull` in `dplyr` always returns the full computed vector. We don't have to stick to `dplyr` but I can't see much value in users working with 'lazy' `Series` directly. Only in the context of DF operations.

test/explorer/data_frame_test.exs
```diff
@@ -699,4 +699,69 @@ defmodule Explorer.DataFrameTest do
     assert Enum.to_list(columns["x"]) == [1, 2, 3]
     assert Enum.to_list(columns["y"]) == ["a", "b", "c"]
   end
```
josevalim, line 701: Let's test `collect` for the regular no-op backend. :)

lib/explorer/polars_backend/lazy_data_frame.ex
cigrainger: I agree but there's no way to get the realised series from a lazy series without some very major changes (keep the lazy df resource in the series struct?).

josevalim: You could introduce a LazySeries with a different backend implementation that keeps the dataframe but I think we should not allow the operation altogether (i.e. do not implement pull and let it raise instead). Especially because even if we can memoize/cache for LazyPolars, we won't be able to do it for ExplorerSQL? So I would postpone this!

cigrainger: I would think we could memoize/cache for ExplorerSQL. dbplyr does it! But that's fine. It means either no Access or rewriting access and writing a custom inspect which mostly just copies the default implementation, but again not a huge deal. Also means that if you want to pull you need to select, then collect, then pull. Not the end of the world.

lib/explorer/data_frame.ex
cigrainger: Oh I implemented it but then didn't expose it 😬

lib/explorer/polars_backend/lazy_data_frame.ex
josevalim: > I would think we could memoize/cache for ExplorerSQL.

The problem is that there is no "easy" way to do this in Erlang/Elixir, I believe. We would need a process for coordination or some sort of special NIF. I will keep thinking about this problem though!

native/explorer/src/lazyframe.rs
josevalim: Ok, one of the things I was talking to @philss is that we could keep the computation of names and types within `Explorer.DataFrame` (like we keep `groups`). I think ExplorerSQL would need to keep track of names and types _without_ realizing the plan (which is fairly more expensive there)? So we can either:

1. Implement this in ExplorerSQL
2. Implement this in Explorer.DataFrame so it benefits everyone (including LazyPolars)

Thoughts?

josevalim: Btw, we don't need to decide it for this PR, but something we will have to sync fairly soon!

lib/explorer/data_frame.ex
cigrainger: Oh now I'm questioning myself. When exposing it (not defining in the behaviour), should it be `lazy/0` or `lazy/1`? If `lazy/0` is it just the lazy version of the _default_ backend? It would need to be `lazy/1`, taking a data frame, to return the lazy version of a given DF's backend.

native/explorer/src/lazyframe.rs
cigrainger: Let's sync on this soon and figure it out. It's not setting alarm bells ringing for me except for keeping the names/dtypes in order. Given it's an instant lookup in eager polars and very cheap in lazy we could even just call the NIF on every `update_dataframe` call, which would be easy enough.

lib/explorer/data_frame.ex
josevalim: >  If lazy/0 is it just the lazy version of the default backend? 

Yes!

My rationale was that `lazy()` returns the lazy version of the new dataframe. So we would do:

    Explorer.DataFrame.lazy().read_csv()

And then `to_lazy` is the one converting an existing dataframe to lazy. We could have both as `lazy/0` and `lazy/1` but I thought the different names could help? 🤔

josevalim: Oh shit, I just realized that this does not quite work. I need to think about it. Let's postpone lazy() then as well. :D

josevalim: Ok... one way I think the above can work is for us to have something like:

    require Explorer.Backend.DataFrame
    lazy_delegates!()

And that will define `read_csv` and friends that call `Explorer.DataFrame` passing the backend as an option.

Is this worth it? I would say yes. I think `Explorer.DataFrame.lazy().read_csv()` is a really nice API for reading lazy dataframes. If you agree, I can try to work on that on a follow up PR. :) So you can skip `lazy/0` on this one still!

josevalim (approved):

lib/explorer/data_frame.ex
cigrainger: I agree. I think that's a really nice API as well, totally. An alternative that comes to mind is to pass `lazy: true` or `:backend` as options, but I think I prefer what you suggested.

josevalim: Oh, I like the `lazy: true` option. Definitely better than backend for this use case. I feel `.lazy()` is more explicit though when reading the code (but also a bit more magic).

josevalim (approved):
I think we only need a test for inspect and ship it!!! Exciting!!!!!

PR #203, opened 2022-05-03
Normalize options on IO operations

### LINKED ISSUE #187: Normalize options on IO operations
For example, IPC has columns and projections, but CSV calls it `with_columns`. Options like with_row_count, with_n_rows (max_rows), with_columns, and with_projections are supported by Parquet and IPC too, but not exposed. My suggestion is to add those options to all relevant IO operations and make them consistent.

I also suggest to merge columns+projections into a single option, called `columns`. The option is either a list of strings, a list of atoms (to be converted into a list of strings), or a list of integers. Internally we convert it to either projections or columns respectively.

kimjoaoun:
This Pull Request, if approved, closes #187.

- [x] from_csv argument names
- [x] from_ipc argument names
- [x] tests

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -22,7 +22,7 @@ defmodule Explorer.Backend.DataFrame do
               header? :: boolean(),
               encoding :: String.t(),
               max_rows :: integer() | Inf,
-              with_columns :: list(String.t()) | list(Atom.t()) | list(integer()) | nil,
+              columns :: list(String.t()) | list(Atom.t()) | list(integer()) | nil,
               infer_schema_length :: integer() | nil,
               parse_dates :: boolean()
             ) :: result(df)
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -22,7 +22,7 @@ defmodule Explorer.Backend.DataFrame do
               header? :: boolean(),
               encoding :: String.t(),
               max_rows :: integer() | Inf,
-              with_columns :: list(String.t()) | nil,
+              with_columns :: list(String.t()) | list(Atom.t()) | list(integer()) | nil,
               infer_schema_length :: integer() | nil,
               parse_dates :: boolean()
             ) :: result(df)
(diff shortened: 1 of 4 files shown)


lib/explorer/data_frame.ex
```diff
@@ -233,24 +233,21 @@ defmodule Explorer.DataFrame do
...
+        with_columns: nil
       )
 
     backend = backend_from_options!(opts)
 
     backend.from_ipc(
       filename,
-      opts[:columns],
-      opts[:projection]
+      opts[:with_columns]
```
josevalim, line 250: Let’s rename the option to `columns` too!

kimjoaoun: Ok!

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -70,6 +72,17 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
+  def column_list_check(list) do
+    cond do
+      is_nil(list) -> {nil, nil}
+      Enum.all?(list, &is_atom/1) -> {Enum.map(list, &Atom.to_string/1), nil}
+      Enum.all?(list, &is_binary/1) -> {list, nil}
+      Enum.all?(list, &is_integer/1) -> {nil, list}
+      ## TODO: better error message
+      true -> raise "Invalid column list"
+    end
+  end
```
josevalim, line 84:
```suggestion
defp column_list_check(list) do
    cond do
      is_nil(list) ->
        {nil, nil}

      Enum.all?(list, &is_atom/1) ->
        {Enum.map(list, &Atom.to_string/1), nil}
      Enum.all?(list, &is_binary/1) ->
        {list, nil}

      Enum.all?(list, &is_integer/1) ->
        {nil, list}

      true ->
        raise ArgumentError,
              "expected :columns to be a list of only integers, only atoms, or only binaries, " <>
                "got: #{inspect(list)}"
    end
  end
```

lib/explorer/backend/data_frame.ex
```diff
@@ -22,7 +22,7 @@ defmodule Explorer.Backend.DataFrame do
               header? :: boolean(),
               encoding :: String.t(),
               max_rows :: integer() | Inf,
-              with_columns :: list(String.t()) | nil,
+              with_columns :: list(String.t()) | list(Atom.t()) | list(integer()) | nil,
```
josevalim, line 25:
```suggestion
              columns :: list(String.t()) | list(Atom.t()) | list(integer()) | nil,
```

PR #208, opened 2022-05-04
Optimize backend lookup

josevalim:
It is a hack but one that is safe for us to perform.

This avoid the process of splitting the module name into parts and joining them together.


cigrainger (approved):
TIL

PR #210, opened 2022-05-04
Add support for negative indexes when selecting columns

philss:
Also add support for "negative ranges".

This also removes some special conditions for `drop_nil` and `rename_with` functions.

Closes https://github.com/elixir-nx/explorer/issues/204


philss:
cc/ @josevalim

PR #211, opened 2022-05-04
Add DataFrame.new/2

### LINKED ISSUE #175: Option to manually supply types to `from_rows`
I'm currently using explorer to convert a database table to parquet files, but sometimes batches only hold null values for a column, so type inference fails: `Could not infer row types, because of the null values`

### LINKED ISSUE #191: Add DataFrame.new that uses Table.Reader
/cc @jonatanklosko

jonatanklosko:
Closes #191, fixes #175.

This implementation essentially does `Table.to_columns |> DataFrame.from_columns`, so in case of row data the conversion happens in Elixir.

## Benchmark

Note that this benchmark uses `table` patched with https://github.com/dashbitco/table/pull/6.

```
Operating System: Linux
CPU Information: Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
Number of Available Cores: 4
Available memory: 30.79 GB
Elixir 1.13.0
Erlang 24.1.2

Name                                 ips        average  deviation         median         99th %
columns |> DF.new                   8.06      124.11 ms     ±2.52%      123.79 ms      134.60 ms
columns |> DF.from_columns          7.99      125.18 ms     ±5.42%      123.37 ms      151.22 ms
rows |> DF.new                      2.60      384.58 ms    ±15.70%      357.50 ms      495.09 ms
rows |> DF.from_rows                2.43      412.20 ms     ±2.90%      408.58 ms      441.89 ms

Comparison: 
columns |> DF.new                   8.06
columns |> DF.from_columns          7.99 - 1.01x slower +1.07 ms
rows |> DF.new                      2.60 - 3.10x slower +260.48 ms
rows |> DF.from_rows                2.43 - 3.32x slower +288.10 ms
```

<details>
<summary>Code</summary>

```elixir
defmodule Bench do
  def run do
    dataset =
      Explorer.Datasets.fossil_fuels()
      |> List.duplicate(100)
      |> Explorer.DataFrame.concat_rows()

    rows = Explorer.DataFrame.to_rows(dataset)
    columns = Explorer.DataFrame.to_columns(dataset)

    Benchee.run(%{
      "rows |> DF.from_rows" => fn ->
        Explorer.DataFrame.from_rows(rows)
      end,
      "columns |> DF.from_columns" => fn ->
        Explorer.DataFrame.from_columns(columns)
      end,
      "rows |> DF.new" => fn ->
        Explorer.DataFrame.new(rows)
      end,
      "columns |> DF.new" => fn ->
        Explorer.DataFrame.new(columns)
      end
    })
  end
end

Bench.run()
```

</details>

## Notes

From the benchmark it seems there is no performance benefit of using `DataFrame::from_rows` in Rust, as compared to conversion to columnary data on the Elixir side.

More importantly, the current version of `DataFrame.from_rows` doesn't support data with all-nil column, neither data with mixed integers and floats (while `from_columns` does normalize them to floats). Consequently, handling those cases would be an additional overhead to the current implementation.

## Discussion

With `DataFrame.new` we no longer need `DataFrame.from_rows` and `DataFrame.from_columns`, but there are two differences:

1. `DataFrame.from_rows` supports maps with missing keys and defaults to `nil`s, but `Table` raises on such. I think this restriction is fine.
2. `DataFrame.from_columns` supports a map with series, like `%{floats: Series.from_list([1.0, 2.0]), ints: Series.from_list([1, nil])}`. `Table` doesn't recognize this as tabular, because `Series` doesn't implement the `Enumerable` protocol. One option would be to have a specialized `DataFrame.from_series` function for this case.

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -48,9 +48,8 @@ defmodule Explorer.Backend.DataFrame do
 
   # Conversion
 
-  @callback new(Table.Reader.t()) :: df
-  @callback from_columns(map() | Keyword.t()) :: df
-  @callback from_rows(list(map() | Keyword.t())) :: df
+  @callback from_tabular(Table.Reader.t()) :: df
+  @callback from_series(map() | Keyword.t()) :: df
   @callback to_columns(df, convert_series? :: boolean(), atom_keys? :: boolean()) :: map()
   @callback to_rows(df, atom_keys? :: boolean()) :: [map()]
   @callback dump_csv(df, header? :: boolean(), delimiter :: String.t()) :: String.t()
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -48,6 +48,7 @@ defmodule Explorer.Backend.DataFrame do
 
   # Conversion
 
+  @callback new(Table.Reader.t()) :: df
   @callback from_columns(map() | Keyword.t()) :: df
   @callback from_rows(list(map() | Keyword.t())) :: df
   @callback to_columns(df, convert_series? :: boolean(), atom_keys? :: boolean()) :: map()
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -123,7 +123,7 @@ defmodule Explorer.DataFrame do
   defp to_column_name(column) when is_atom(column), do: Atom.to_string(column)
 
   # Normalize pairs of `{column, value}` where value can be anything.
-  # The `column` is only validated if it's an integer. We check that the index is present. 
+  # The `column` is only validated if it's an integer. We check that the index is present.
   defp to_column_pairs(df, pairs), do: to_column_pairs(df, pairs, & &1)
 
   # The function allows to change the `value` for each pair.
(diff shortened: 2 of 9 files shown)


josevalim:
@jonatanklosko should we special case and allow `%{field: Series...}` in `new` or would you rather not mix those?

jonatanklosko:
For the backend I would definitely keep them separate, so each backend doesn't need to do the recognising on its own.

For the top-level API, I don't have a strong opinion. `Table.Reader.t()` is already a pretty relaxed type, so mixing it with series is a bit weird. I can see someone throwing anything at `DataFrame.new` and expecting it to work, but this itself doesn't feel like a strong argument to me.

josevalim:
Honestly, it feels like `new` should accept the result of both `to_rows` and `to_columns`. However, I now find it a bit misleading if `to_columns` does not return something considered columns by `Table`. Perhaps we should have both `to_columns` and `to_series`? Today there is a `convert_series` option in `to_columns` but a separate function can be clearer. So overall:

1. Remove from_columns and from_row
2. Add `to_series`
3. Add new and it supports tables and the `to_series` result

What do you think @cigrainger and @kimjoaoun?

jonatanklosko:
Is there a good reason for keeping `to_columns` assuming we have `to_series`?

josevalim:
to_columns would be lists as values. But maybe we should have to_list and to_series only, especially now that from_columns and from_rows is gone?

jonatanklosko:
Given that we have `Series.to_list` and even `Series.to_enum`, just `to_series` sounds reasonable to me!

josevalim (approved):
I am happy with this as is. We can tackle the to_columns/to_series in another PR. :)

cigrainger:
I agree with this plan:

> 1. Remove from_columns and from_row
> 2. Add to_series and remove the convert_series option from to_columns
> 3. Add new and it supports tables and the to_series result

jonatanklosko:
1 and 3 is done, let's do 2 in a separate PR, once #212 is merged too :)

cigrainger:
And it makes sense to me to tackle to_columns/to_series in another PR.

Also glad to see the back of `from_rows` on the Rust side, as there was a fair amount of work necessary for that one. It makes sense there's not much benefit as Arrow is columnar. I am curious about the size of the table in the benchmarks though and whether larger ones do see a benefit from the Rust implementation.

I think we may be able to remove `ExAnyValue` after this but I'll check and if so I'll follow this PR with one to remove it.

jonatanklosko:
@cigrainger the bench code is collapsed in the PR description under the results. The data was 100x fossil fuels, which is 109400 rows, 10 columns.

josevalim:
I have added to_series as part of my to_columns PR, so it is all done!

Issue #206, opened 2022-05-04
Remove with_ prefix in options and consistenly use `columns` instead of `cols`


josevalim:
/cc @philss

philss:
Both PRs are merged. I'm closing this. cc/ @josevalim

PR #213, opened 2022-05-04
Remove "with" prefix from options of some functions

philss:
The "with" prefix is a thing from Polars. We don't need to follow the
same style.

This is related to #206

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -422,6 +422,12 @@ defmodule Explorer.Series do
 
   Can sample with or without replacement.
 
+  ## Options
+
+    * `replacement` - If set to `true`, each sample will be independent and therefore values may repeat.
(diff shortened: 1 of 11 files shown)


josevalim (approved):

lib/explorer/series.ex
```diff
@@ -422,6 +422,12 @@ defmodule Explorer.Series do
 
   Can sample with or without replacement.
 
+  ## Options
+
+    * `replacement` - If set to `true`, each sample will be independent and therefore values may repeat.
+      Required to be `true` for `n` greater then the number of rows in the dataframe or `frac` > 1.0. (default: `false`)
```
philss, line 428:
```suggestion
      Required to be `true` for `n` greater then the number of rows in the series or `frac` > 1.0. (default: `false`)
```

PR #214, opened 2022-05-04
Run formatter before running tests on CI

philss:
This ensure that we fail fast if any formatting issue is found.

Reference: https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,20 @@ env:
   MIX_ENV: test
 
 jobs:
+  format:
+    runs-on: ubuntu-latest
+    name: mix format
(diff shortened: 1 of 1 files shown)


josevalim:
I prefer to not fail fast in case there are also test errors?

philss:
@josevalim this way they are not going to start running. My idea was to have a faster feedback that the code is not well formatted before running the tests. WDYT?

josevalim:
Right. I prefer to run as much as possible so we give all relevant information :)

philss:
@josevalim OK, so I'm closing this :)

PR #218, opened 2022-05-05
Normalize dtype handling and CSV options

josevalim:
* Remove `names` option in `from_csv` to remove ambiguity
    (users can call rename afterwards)

  * Normalizea and validate `dtypes` in `from_csv`

  * Remove `Inf` which is not a common Elixir value

  * Support returning a list from `Series.transform` to
    unify `dtypes` handling

### DIFF
--- a/notebooks/exploring_explorer.livemd
+++ b/notebooks/exploring_explorer.livemd
@@ -30,7 +30,7 @@ For CSV, your 'usual suspects' of options are available:
 
 * `delimiter` - A single character used to separate fields within a record. (default: `","`)
 * `dtypes` - A keyword list of `[column_name: dtype]`. If `nil`, dtypes are imputed from the first 1000 rows. (default: `nil`)
-* `header?` - Does the file have a header of column names as the first row or not? (default: `true`)
+* `header` - Does the file have a header of column names as the first row or not? (default: `true`)
 * `max_rows` - Maximum number of lines to read. (default: `Inf`)
 * `names` - A list of column names. Must match the width of the dataframe. (default: nil)
 * `null_character` - The string that should be interpreted as a nil value. (default: `"NA"`)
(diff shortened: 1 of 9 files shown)


notebooks/exploring_explorer.livemd
```diff
@@ -24,19 +24,17 @@ Mix.install([
 
 ## Reading and writing data
 
-Data can be read from delimited files (like CSV), Parquet, and the Arrow IPC (feather) format. You can also load in data from a map or keyword list of columns with `Explorer.DataFrame.new/1`.
+Data can be read from delimited files (like CSV), NDJSon, Parquet, and the Arrow IPC (feather) format. You can also load in data from a map or keyword list of columns with `Explorer.DataFrame.new/1`.
```
cigrainger, line 27:
```suggestion
Data can be read from delimited files (like CSV), NDJSON, Parquet, and the Arrow IPC (feather) format. You can also load in data from a map or keyword list of columns with `Explorer.DataFrame.new/1`.
```

notebooks/exploring_explorer.livemd
```diff
@@ -24,19 +24,17 @@ Mix.install([
...
 
 For CSV, your 'usual suspects' of options are available:
 
 * `delimiter` - A single character used to separate fields within a record. (default: `","`)
-* `dtypes` - A keyword list of `[column_name: dtype]`. If `nil`, dtypes are imputed from the first 1000 rows. (default: `nil`)
-* `header?` - Does the file have a header of column names as the first row or not? (default: `true`)
-* `max_rows` - Maximum number of lines to read. (default: `Inf`)
-* `names` - A list of column names. Must match the width of the dataframe. (default: nil)
```
cigrainger, line 35: This one is meant for renaming when you read it in or if it doesn't have a header. You could just pipe to `rename` I suppose. It's commonly supported in other languages but I don't think I've just about ever used it.

cigrainger (commented):
This is muuuuch tidier.

notebooks/exploring_explorer.livemd
josevalim: Yeah. The issue is that it introduces ambiguity for the other option (the names before or after). We would also have to normalize options, etc. I think going with renaming is the cleaner way to do it. :) If necessary, we can add a mention to the docs about rename!

josevalim:

PR #221, opened 2022-05-06
Do not add special meaning to empty lists

philss:
This changes the behaviour of functions that are treating empty lists as
exceptions. Now we return the original data frame untouched instead of
raising error.

Closes https://github.com/elixir-nx/explorer/issues/205

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1820,25 +1820,20 @@ defmodule Explorer.DataFrame do
     end
 
     id_columns =
-      if is_function(opts[:id_columns]) do
-        opts[:id_columns]
-      else
-        to_existing_columns(df, opts[:id_columns])
-      end
-
-    id_columns =
-      case id_columns do
-        [] ->
-          # TODO: could we check for empty lists after filtering?
-          raise ArgumentError,
-                "id_columns must select at least one existing column, but #{inspect(opts[:id_columns])} selects none"
-
-        [_ | _] = names ->
-          Enum.filter(names, &(&1 not in [names_from, values_from]))
-
+      case opts[:id_columns] do
         fun when is_function(fun) ->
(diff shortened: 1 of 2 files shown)


test/explorer/data_frame_test.exs
```diff
@@ -600,6 +596,55 @@ defmodule Explorer.DataFrameTest do
     assert DF.to_columns(df4) == %{"a" => [1], "b" => [1]}
   end
 
+  test "rename_with/2", %{df: df} do
```
josevalim, line 599: Can you use `describe` and break this into multiple tests? Like:

```
describe "rename_with/2" do
  test "with list" do

  test "with range" do

```

test/explorer/data_frame_test.exs
```diff
@@ -600,6 +596,55 @@ defmodule Explorer.DataFrameTest do
...
+
+    assert df_names -- df4_names == ["total"]
+    assert df4_names -- df_names == ["TOTAL"]
+
+    df5 = DF.rename_with(df, &String.upcase/1, &String.starts_with?(&1, "non-existent"))
+
+    assert df5 == df
+  end
+
+  test "pivot_wider/4" do
```
josevalim, line 629: If you can break this down too!

lib/explorer/data_frame.ex
```diff
@@ -1813,12 +1820,20 @@ defmodule Explorer.DataFrame do
...
+        opts[:id_columns]
+      else
+        to_existing_columns(df, opts[:id_columns])
+      end
+
+    id_columns =
+      case id_columns do
         [] ->
-          Enum.filter(names, &(&1 not in [names_from, values_from]))
+          # TODO: could we check for empty lists after filtering?
```
josevalim, line 1832: Yeah, I would check after the filter.

josevalim (approved):

cigrainger (approved):
Fantastic! ❤️

PR #222, opened 2022-05-07
Make inspect more modular

cigrainger:
This is in preparation for #202. I've added a callback `data_for_inspect/2`, which will maintain control over default inspect while permitting a bit more flexibility for the backend. I haven't done the same for Series yet because I don't yet foresee a similar issue.

I'm holding off on changing all the doctests until we're happy with a solution on this.

I've changed the inspect protocol for dataframes a bit to bring it close to a `tibble` in R and a bit closer to inspect for Series. 

```
#Explorer.DataFrame<
  Polars[178 x 14]
  class integer [1, 1, 1, 1, 1, ...]
  alcohol float [14.23, 13.2, 13.16, 14.37, 13.24, ...]
  malic_acid float [1.71, 1.78, 2.36, 1.95, 2.59, ...]
  ash float [2.43, 2.14, 2.67, 2.5, 2.87, ...]
  alcalinity_of_ash float [15.6, 11.2, 18.6, 16.8, 21.0, ...]
  magnesium integer [127, 100, 101, 113, 118, ...]
  total_phenols float [2.8, 2.65, 2.8, 3.85, 2.8, ...]
  flavanoids float [3.06, 2.76, 3.24, 3.49, 2.69, ...]
  nonflavanoid_phenols float [0.28, 0.26, 0.3, 0.24, 0.39, ...]
  proanthocyanins float [2.29, 1.28, 2.81, 2.18, 1.82, ...]
  color_intensity float [5.64, 4.38, 5.68, 7.8, 4.32, ...]
  hue float [1.04, 1.05, 1.03, 0.86, 1.04, ...]
  od280_over_od315_of_diluted_wines float [3.92, 3.4, 3.17, 3.45, 2.93, ...]
  proline integer [1065, 1050, 1185, 1480, 735, ...]
>
```

And with groups:
```
#Explorer.DataFrame<
  Polars[178 x 14]
  Groups: ["class"]
  class integer [1, 1, 1, 1, 1, ...]
  alcohol float [14.23, 13.2, 13.16, 14.37, 13.24, ...]
  malic_acid float [1.71, 1.78, 2.36, 1.95, 2.59, ...]
  ash float [2.43, 2.14, 2.67, 2.5, 2.87, ...]
  alcalinity_of_ash float [15.6, 11.2, 18.6, 16.8, 21.0, ...]
  magnesium integer [127, 100, 101, 113, 118, ...]
  total_phenols float [2.8, 2.65, 2.8, 3.85, 2.8, ...]
  flavanoids float [3.06, 2.76, 3.24, 3.49, 2.69, ...]
  nonflavanoid_phenols float [0.28, 0.26, 0.3, 0.24, 0.39, ...]
  proanthocyanins float [2.29, 1.28, 2.81, 2.18, 1.82, ...]
  color_intensity float [5.64, 4.38, 5.68, 7.8, 4.32, ...]
  hue float [1.04, 1.05, 1.03, 0.86, 1.04, ...]
  od280_over_od315_of_diluted_wines float [3.92, 3.4, 3.17, 3.45, 2.93, ...]
  proline integer [1065, 1050, 1185, 1480, 735, ...]
>
```

I envisage a polars lazy backend would look like:

```
#Explorer.DataFrame<
  PolarsLazy[?? x 14]
  class integer [1, 1, 1, 1, 1, ...]
  alcohol float [14.23, 13.2, 13.16, 14.37, 13.24, ...]
  malic_acid float [1.71, 1.78, 2.36, 1.95, 2.59, ...]
...
```
  
## Alternative option

It could also be:

```
#Explorer.DataFrame<
  polars_lazy[?? x 14]
```

and 

```
#Explorer.DataFrame<
  polars[178 x 14]
```

I've also dropped the `import` for the `Shared` functions in favour of `alias`. I've also moved `apply_impl` there because it was replicated in `Series` and `DataFrame`.

## Comparison to R

For comparison, here's a 'lazy' dataframe (sqlite backend) from `dbplyr`:
```
#> # Source:   table<mtcars> [?? x 11]
#> # Database: sqlite 3.34.1 [:memory:]
#>      mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#>  1  21       6  160    110  3.9   2.62  16.5     0     1     4     4
#>  2  21       6  160    110  3.9   2.88  17.0     0     1     4     4
#>  3  22.8     4  108     93  3.85  2.32  18.6     1     1     4     1
#>  4  21.4     6  258    110  3.08  3.22  19.4     1     0     3     1
#>  5  18.7     8  360    175  3.15  3.44  17.0     0     0     3     2
#>  6  18.1     6  225    105  2.76  3.46  20.2     1     0     3     1
#>  7  14.3     8  360    245  3.21  3.57  15.8     0     0     3     4
#>  8  24.4     4  147.    62  3.69  3.19  20       1     0     4     2
#>  9  22.8     4  141.    95  3.92  3.15  22.9     1     0     4     2
#> 10  19.2     6  168.   123  3.92  3.44  18.3     1     0     4     4
#> # … with more rows
```

And a grouped eager tibble:
```
#> # A tibble: 32 × 11
#> # Groups:   cyl [3]
#>      mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#>  1  21       6  160    110  3.9   2.62  16.5     0     1     4     4
#>  2  21       6  160    110  3.9   2.88  17.0     0     1     4     4
#>  3  22.8     4  108     93  3.85  2.32  18.6     1     1     4     1
#>  4  21.4     6  258    110  3.08  3.22  19.4     1     0     3     1
#>  5  18.7     8  360    175  3.15  3.44  17.0     0     0     3     2
#>  6  18.1     6  225    105  2.76  3.46  20.2     1     0     3     1
#>  7  14.3     8  360    245  3.21  3.57  15.8     0     0     3     4
#>  8  24.4     4  147.    62  3.69  3.19  20       1     0     4     2
#>  9  22.8     4  141.    95  3.92  3.15  22.9     1     0     4     2
#> 10  19.2     6  168.   123  3.92  3.44  18.3     1     0     4     4
```


josevalim:
The new inspect output looks good to me. However, I also think we need this for series. For example, for lazy dataframes, series will be a lazy column that accumulates operations (akin to a query expression), and I think we need custom printing there.

Now to review the code :)

lib/explorer/backend/data_frame.ex
```diff
@@ -59,6 +59,11 @@ defmodule Explorer.Backend.DataFrame do
   @callback shape(df) :: {integer(), integer()}
   @callback n_rows(df) :: integer()
   @callback n_columns(df) :: integer()
+  @callback data_for_inspect(df, opts :: Inspect.Opts.t()) ::
+              {rows :: nil | non_neg_integer(),
+               columns ::
+                 non_neg_integer(), backend :: atom() | String.t(),
+               series :: [{name :: String.t(), dtype :: atom() | String.t(), values :: list()}]}
```
josevalim, line 66: I think this contract is a bit too constrained. For example, in ExplorerSQL, we may want to show the SQL query when inspecting. Perhaps we could return this:

    {backend :: string,
     rows :: nil | non_neg_integer,
     cols :: nil | non_neg_integer,
     prelude :: Inspect.Algebra.t,
     series :: ...}

Some notes:

  * `prelude` is an algebra document that is placed right below `Backend[rows x cols]`. Most backends can default to `empty()`.

  * I have said that `cols` can be nil based on a conversation with @philss. For example, should `ExplorerSQL` do a query to get the table fields+types as soon as you say the table? Or should we say we simply don't know the table information until when the query is made? We don't have to support this but we might as well make it consistent with rows?

  * I think you can also force more strict types here. Such as requiring the `backend` to be a string and the dtype to be an atom (which is its normal representation)

josevalim (commented):
+1 for the general direction, I have added some comments to tighten and expand the contract a tiny bit more. :)

cigrainger:
> However, I also think we need this for series. For example, for lazy dataframes, series will be a lazy column that accumulates operations (akin to a query expression), and I think we need custom printing there.

This is what I've been getting at before: I don't think this is possible with Polars nor does it make sense. If I use `DataFrame.pull/2`, even on a lazy dataframe, I expect the _realised_ Series. This is the behaviour in dplyr.

lib/explorer/backend/data_frame.ex
cigrainger: I think with SQL it will be quite unwieldy to display the SQL query when inspecting. Why not just allow the user to call for it with a function? But that's predicated on: if it's SQL I'd want to fetch the query with a limit (`opts.limit + 1`), like Polars lazy. This is what `dbplyr` does.

josevalim:
> I think we'd be fighting an uphill battle to try to implement a 'lazy' Series for basically zero benefit.

However, we need an API for lazy collecting filter expressions for example:

    DF.filter(lazy_df, fn df -> Series.less(df["foo"], 10) end)

And even though we may not return it from pull, we want it to be lazy and someone may still inspect it?

>  If I use DataFrame.pull/2, even on a lazy dataframe, I expect the realised Series. This is the behaviour in dplyr.

My concern with this is that someone can write this code:

    lazy_df =
      to_lazy(df)
      |> select(...)
      |> arrange(...)

    {lazy_df["foo"], lazy_df["bar"]}

And now we will execute the query twice. dplyr/polars may be able to deal with this because, as mutable languages, they can mutate something so the query is computed just once. Unfortunately there is no clean abstraction for achieving the same in Elixir. So I am afraid you can't pull anything in Elixir before you collect (and once you collect it effectively becomes another backend).

lib/explorer/backend/data_frame.ex
cigrainger: In `dbplyr` they have `show_query`:

```
> show_query(mtcars2)
<SQL>
SELECT *
FROM `mtcars`
```

josevalim: Yeah, I think we will not end up showing the underlying SQL, but I thought having the provision for such a mechanism is good in case we need it in the future? Otherwise we will have to enlarge/deprecate the behaviour function if/when we need it.

cigrainger:
> And even though we may not return it from pull, we want it to be lazy and someone may still inspect it?

You can't inspect a constituent series from a dataframe without pulling it. The index calls `pull`. In R, indexed access returns `NULL` for remote tables. It's the same in Polars, by the way -- lazy does not extend to series and there's not a mutation under the hood. You have to collect to get the values.

cigrainger:
> we need an API for lazy collecting filter expressions

I agree. This is why I've been so excited about `across` and the possibility of using a macro in this context. If we want to use the `Series` API in all contexts, it needs to be working with expressions in some and actual Series in others.

josevalim:
Right. My thought process is that `across` would compile to functions in the Series API. So `across(column < 10)` would become `fn df -> Series.less(df["column"], 10) end`. The main benefit is that we can reduce the API surface considerably. Introducing a QueryExpr or similar would just end-up becoming even more functions for users to learn. The lazy series in across does not need to be same returned by `lazy_df["column"]` outside of a `across` though.

Still it may make sense to detour a bit of dplyr because I am not convinced "In R, indexed access returns NULL for remote tables." is a good idea. Although it seems you are not convinced returning a lazy series is a good idea as well. :D Let's sync because @philss and I also have some questions about preloading table information.

cigrainger:
Okay... trying to switch this to an inspect callback with a default implementation. Finding it a bit fiddly and I'm wondering if the `__using__` macro should be a default implementation of the `inspect/2` callback or a private function that takes `n_rows`, `n_columns`, `backend`, and `series` and constructs the algebra? The former is... impossible (?) to do unless we want to show the fully qualified backend name (`Explorer.PolarsBackend.DataFrame` or similar) which doesn't allow us to indicate that it's lazy without redefining the _entire_ thing. The latter feels less orthodox but more flexible.

cigrainger:
Just realised I can do the former and use the macro opts. Updates incoming.

lib/explorer/data_frame.ex
```diff
@@ -2321,6 +2322,20 @@ defmodule Explorer.DataFrame do
...
+  defimpl Inspect do
+    import Inspect.Algebra
+
+    alias Explorer.DataFrame
+    alias Explorer.Series
+
+    @default_limit 5
+
+    def inspect(df, opts) do
+      opts = Map.put(opts, :limit, @default_limit)
```
josevalim, line 2335: Perhaps move this inside the inspect implementation too?

josevalim:
I am not thrilled with the macro approach but:

1. this is definitely an improvement
2. i can't think of anything better for now

So my suggestion is to ship this and I will try to come up with alternatives. :)

PR #225, opened 2022-05-07
Changes to rename and rename_with

josevalim:
* Remove rename/2 with function as that was redundant with rename_with/2

* Make the anonymous function in rename_with/3 the last argument

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1291,28 +1291,16 @@ defmodule Explorer.DataFrame do
         b integer [1, 3, 1]
       >
 
-  Or if you want to use a function:
-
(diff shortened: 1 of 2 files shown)


josevalim:
@cigrainger I am not convinced about the second change. Having the function as last argument is more common in Elixir but I am not a biggest fan of having the optional argument in the middle.

At the same time, it feels saying `rename_with(dataframe, what_to_change, how_to_change)` is nicer than saying how to change and then later what.

I will be glad to go with your decision.

cigrainger:
I agree with your reasoning. I'm not a huge fan of an optional argument in the middle, but I think you're right about the function as last argument and saying `dataframe, what_to_change, how_to_change` makes more sense. Let's go with it as is.

cigrainger (approved):

josevalim:

PR #229, opened 2022-05-17
Add `Series.coalesce`

### LINKED ISSUE #226: Function that finds the first non-missing element in a vector
## Introduction

I would be good having in the module `Explorer.Series` a function similar to dplyr's [coalesce](https://dplyr.tidyverse.org/reference/coalesce.html) or postgreSQL's [coalesce](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL).
In Pandas there's a function called [DataFrame.combine_first](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.combine_first.html) that do something similar, but I'm not sure if it achieves the same results as the ones described above.

What coalesce's do is:
Having a pair of vectors of the same length, merge them, keeping the first value of each side of the pair that is not a `:nil`

## Examples

```elixir
iex> s1 = Explorer.Series.from_list([1, 2, 3, nil])
iex> s2 = Explorer.Series.from_list[nil, nil, nil, 4])

iex> Explorer.Series.coalesce(s1, s2)
#Explorer.Series<
  integer[3]
  [1, 2, 3, 4]
>
```

cigrainger:
Closes #226. Sorry @kimjoaoun I couldn't resist :).

NB: `clippy` is failing on the latest nightly rust because it adds a new check by default. I've made the necessary changes in #202 and will rebase once that's merged.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -647,6 +647,32 @@ defmodule Explorer.Series do
   defp concat_reducer(%Series{dtype: dtype1}, %Series{dtype: dtype2}),
     do: raise(ArgumentError, "dtypes must match, found #{dtype1} and #{dtype2}")
 
+  @doc """
+  Finds the first non-missing element at each position.
+
+  ## Examples
+
+      iex> s1 = Explorer.Series.from_list([1, 2, nil, nil])
+      iex> s2 = Explorer.Series.from_list([1, 2, nil, 4])
+      iex> s3 = Explorer.Series.from_list([nil, nil, 3, 4])
+      iex> Explorer.Series.coalesce([s1, s2, s3])
+      #Explorer.Series<
+        integer[4]
+        [1, 2, 3, 4]
+      >
+  """
+  @spec coalesce([Series.t()]) :: Series.t()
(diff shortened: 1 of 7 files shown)


kimjoaoun (approved):
LGTM!

lib/explorer/series.ex
```diff
@@ -645,6 +647,32 @@ defmodule Explorer.Series do
...
+  def coalesce([%Series{} = h | t] = _series),
+    do: Enum.reduce(t, h, &Shared.apply_impl(&2, :coalesce, [&1]))
+
+  @doc """
+  Finds the first non-missing element at each position.
+
+  `coalesce(s1, s2)` is equivalent to `coalesce([s1, s2])`.
+  """
+  @spec coalesce(s1 :: Series.t(), s2 :: Series.t()) :: Series.t()
+  def coalesce(s1, s2), do: concat([s1, s2])
```
josevalim, line 674: Should we support `coalesce(series, value)` as a quick way to replace nil by a given value or is there already such an API?

cigrainger: We have `Series.fill_missing/2` for that. :)

PR #232, opened 2022-05-17
Compose dataframe inspecting

josevalim:
The idea is to keep the "shell" of the inspect representation
constant but delegate the remaining to a shared implementation.

df_as_str was removed because we are missing the polars flag
to make it work, so I thought we could skip it too.

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -110,71 +110,55 @@ defmodule Explorer.Backend.DataFrame do
   @callback ungroup(df, columns :: [column_name]) :: df
   @callback summarise(df, aggregations :: map()) :: df
 
-  defmacro __using__(opts) do
-    quote location: :keep, bind_quoted: [opts: opts] do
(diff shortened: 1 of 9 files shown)


josevalim:
If everyone is happy, I can try something similar for Series.

josevalim:
Any idea what I did wrong with the Rust code? :D

cigrainger:
The Rust code is fine -- same problem I mentioned in #229. :)

cigrainger (approved):
Oh this is great. I'll put my `shape` implementation that returns `{nil, n_columns}` back. :D

philss (approved):
Cool!

josevalim:
Implementation for series is in. It also expects a backend but we don't use it, I will open up a separate issue for that (it is a large change which I can't work on because of my arm injury).

cigrainger:
A rebase will get checks to pass after I merge #234. Then I'll fix #202 to use this and drop `lazy/0` and we can go on from there :).

josevalim:
@cigrainger please go ahead and merge #202 and I will rebase this after (going to bed now) And pls don’t drop lazy/0, just don’t expose it :) I will work on exposing it tomorrow.

PR #235, opened 2022-05-23
Fix minor typespecs issues

philss:
Also, this makes reuse of some type definitions.


cigrainger (approved):
❤️ Wonderful. Thanks @philss!

PR #236, opened 2022-05-24
Remove duplications from `Explorer.Series`

philss:
This makes some functions lean and add helpers to check for types.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -38,6 +38,10 @@ defmodule Explorer.Series do
 
   @compile {:no_warn_undefined, Nx}
 
+  defguardp numeric_dtype?(dtype) when dtype in [:float, :integer]
+  defguardp numeric_or_bool_dtype?(dtype) when dtype in [:float, :integer, :boolean]
+  defguardp numeric_or_date_dtype?(dtype) when dtype in [:float, :integer, :date, :datetime]
+
   @impl true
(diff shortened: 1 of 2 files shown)


cigrainger (approved):

cigrainger:
I like it!

philss:
@cigrainger I was thinking after open the PR: maybe I should use multiple function clauses instead of `case` (with the `{left, right}` tuple) . Do you prefer that?

cigrainger:
Oh -- yes. I think multiple function clauses is better. I often find myself doing exactly that with case or cond statements on tuples.

philss:
@cigrainger updated. I think it looks better now :smiley:

philss:
I'm merging this. If you think something could be improved, let me know so I can fix :)

cigrainger:
Looks good to me!

lib/explorer/series.ex
```diff
@@ -38,6 +38,10 @@ defmodule Explorer.Series do
 
   @compile {:no_warn_undefined, Nx}
 
+  defguardp numeric_dtype?(dtype) when dtype in [:float, :integer]
+  defguardp numeric_or_bool_dtype?(dtype) when dtype in [:float, :integer, :boolean]
+  defguardp numeric_or_date_dtype?(dtype) when dtype in [:float, :integer, :date, :datetime]
```
josevalim, line 43: The convention for guards is to use the `is_` prefix with no question mark. :)

PR #237, opened 2022-05-30
Split `apply_native` in `apply_{series,dataframe}`

philss:
The idea is to make those functions more explicit.

### DIFF
--- a/lib/explorer/polars_backend/shared.ex
+++ b/lib/explorer/polars_backend/shared.ex
@@ -9,33 +9,36 @@ defmodule Explorer.PolarsBackend.Shared do
   alias Explorer.PolarsBackend.Series, as: PolarsSeries
   alias Explorer.Series, as: Series
 
-  def apply_native(df_or_s, fun, args \\ [])
+  defguardp is_polars_df(df) when is_struct(df, PolarsDataFrame) or is_struct(df, PolarsLazyFrame)
+  defguardp is_polars_series(series) when is_struct(series, PolarsSeries)
 
-  def apply_native(%Series{} = series, fun, args) do
+  def apply_series(series, fun, args \\ [])
+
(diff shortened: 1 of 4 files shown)


philss:
cc @josevalim

lib/explorer/polars_backend/shared.ex
```diff
@@ -9,33 +9,36 @@ defmodule Explorer.PolarsBackend.Shared do
   alias Explorer.PolarsBackend.Series, as: PolarsSeries
   alias Explorer.Series, as: Series
 
-  def apply_native(df_or_s, fun, args \\ [])
+  defguardp is_polars_df(df) when is_struct(df, PolarsDataFrame) or is_struct(df, PolarsLazyFrame)
+  defguardp is_polars_series(series) when is_struct(series, PolarsSeries)
```
josevalim, line 13: Nitpick: the pattern matching is much more efficient for checking structs... so I would remove the guards. So even though there is duplication, I think the duplication may be worth it.

josevalim: If you want to encapsulate it, then maybe `@polars_df [PolarsDataFrame, PolarsLazyFrame]` is better. :)

josevalim (approved):
Just one tiny comment and ship it!

lib/explorer/polars_backend/shared.ex
philss: Good to know!

Using `%module{} when module in @polars_df` is equivalent to having a function clause for each, right?
https://github.com/elixir-nx/explorer/pull/237/commits/a09ee9d43b93318d17cc8464684c394e9e53450d

josevalim: In terms of performance it would be pretty close, yeah!

PR #238, opened 2022-05-30
Keep track of names & dtypes and introduce a "virtual" out DF

philss:
The idea of this change is to simplify the implementation of data frame backends
by applying the changes to a "virtual DF" and letting the backend decide how to apply those changes.

In this PR we have the following callbacks updated:

- [x] `select`
- [x] `mutate`
- [x] `rename`
- [x] `pivot_longer`
- [x] `group_by`
- [x] `ungroup`
- [x] `summarise`
- [x] `join` - a smaller change in the `on` specs
- [x] `join`
- [x] `distinct`

~~The PR is in WIP so we can evaluate if the changes makes sense.~~

~~PS: some of the changes here are duplicated from PR https://github.com/elixir-nx/explorer/pull/237, so after that merge I will rebase.~~

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1744,6 +1744,9 @@ defmodule Explorer.DataFrame do
 
   def pivot_longer(df, columns, opts) do
     opts = Keyword.validate!(opts, value_columns: [], names_to: "variable", values_to: "value")
+    names_to = to_column_name(opts[:names_to])
+    values_to = to_column_name(opts[:values_to])
+
     existing_columns = to_existing_columns(df, columns)
 
     names = names(df)
@@ -1772,25 +1775,31 @@ defmodule Explorer.DataFrame do
 
     value_columns = to_existing_columns(df, value_columns)
 
-    dtypes
-    |> Map.take(value_columns)
-    |> Map.values()
-    |> Enum.uniq()
-    |> length()
-    |> case do
-      1 ->
-        :ok
+    values_dtype =
+      dtypes
+      |> Map.take(value_columns)
+      |> Map.values()
+      |> Enum.uniq()
+      |> case do
+        [dtype] ->
+          dtype
 
-      _ ->
-        raise ArgumentError,
-              "value columns may only include one dtype but found multiple dtypes"
-    end
+        [_ | _] ->
+          raise ArgumentError,
+                "value columns may only include one dtype but found multiple dtypes"
+      end
+
+    new_dtypes =
+      dtypes
+      |> Map.take(columns)
+      |> Map.put(names_to, :string)
+      |> Map.put(values_to, values_dtype)
+
+    out_df = %{df | names: columns ++ [names_to, values_to], dtypes: new_dtypes}
 
     Shared.apply_impl(df, :pivot_longer, [
-      columns,
-      value_columns,
-      opts[:names_to],
-      opts[:values_to]
+      out_df,
+      value_columns
     ])
   end
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1333,7 +1333,9 @@ defmodule Explorer.DataFrame do
     new_names = to_column_names(names)
     check_new_names_length!(df, new_names)
 
-    Shared.apply_impl(df, :rename, [new_names])
+    out_df = %{df | names: new_names}
+
+    Shared.apply_impl(df, :rename, [out_df])
   end
 
   def rename(df, names) when is_column_pairs(names) do
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1833,10 +1833,7 @@ defmodule Explorer.DataFrame do
 
     out_df = %{df | names: columns ++ [names_to, values_to], dtypes: new_dtypes}
 
-    Shared.apply_impl(df, :pivot_longer, [
-      out_df,
-      value_columns
-    ])
+    Shared.apply_impl(df, :pivot_longer, [out_df, value_columns])
   end
 
   @doc """
@@ -2404,25 +2401,23 @@ defmodule Explorer.DataFrame do
     groups = for group <- df.groups, do: {group, df.dtypes[group]}
 
     agg_pairs =
-      Enum.flat_map(column_pairs, fn {column_name, aggregations} ->
-        for agg <- aggregations do
-          name = "#{column_name}_#{agg}"
+      for {column_name, aggregations} <- column_pairs, agg <- aggregations do
+        name = "#{column_name}_#{agg}"
 
-          dtype =
-            case agg do
-              :median ->
-                :float
+        dtype =
+          case agg do
+            :median ->
+              :float
 
-              :mean ->
-                :float
+            :mean ->
+              :float
 
-              _other ->
-                df.dtypes[column_name]
-            end
+            _other ->
+              df.dtypes[column_name]
+          end
 
-          {name, dtype}
-        end
-      end)
+        {name, dtype}
+      end
 
     groups ++ agg_pairs
   end
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2362,7 +2362,39 @@ defmodule Explorer.DataFrame do
         end
       end)
 
-    Shared.apply_impl(df, :summarise, [Map.new(column_pairs)])
+    new_dtypes = names_with_dtypes_for_summarise(df, column_pairs)
+    new_names = for {name, _} <- new_dtypes, do: name
+
+    df_out = %{df | names: new_names, dtypes: Map.new(new_dtypes), groups: []}
+
+    Shared.apply_impl(df, :summarise, [df_out, Map.new(column_pairs)])
+  end
+
+  defp names_with_dtypes_for_summarise(df, column_pairs) do
+    groups = for group <- df.groups, do: {group, df.dtypes[group]}
+
+    agg_pairs =
+      Enum.flat_map(column_pairs, fn {column_name, aggregations} ->
+        for agg <- aggregations do
+          name = "#{column_name}_#{agg}"
+
+          dtype =
+            case agg do
+              :median ->
+                :float
+
+              :mean ->
+                :float
+
+              _other ->
+                df.dtypes[column_name]
+            end
+
+          {name, dtype}
+        end
+      end)
+
+    groups ++ agg_pairs
   end
 
   @doc """
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -101,8 +101,7 @@ defmodule Explorer.DataFrame do
   @valid_dtypes Explorer.Shared.dtypes()
 
   @type data :: Explorer.Backend.DataFrame.t()
-  # TODO: enforce names and dtypes
-  @enforce_keys [:data, :groups]
+  @enforce_keys [:data, :groups, :names, :dtypes]
   defstruct [:data, :groups, :names, :dtypes]
 
   @type column_name :: atom() | String.t()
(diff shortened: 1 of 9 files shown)


josevalim:
:+1: for this change. mmmmmm what you say @cigrainger?

cigrainger:
Yep. Loving it. I think this is a really good approach.

philss:
I'm struggling a little bit with the `pivot_wider` operation. It is technically possible to create a new virtual representation, but it would require us to read all values of  the `names_from` column and match the `dtype` with the values from `values_from`. I'm not sure if we want to do that. Also, it's not clear how this would make easier for the backends.

I'm going to skip this function for now, but if you have ideas, please let me know.

PS: I made a small change in the callback signature to reflect the current implementation and the dplyr API.

philss:
@cigrainger hey Chris, would you mind to help me answering some questions about "grouping"?

- for what I could understand, we keep the group names in the DF's struct, but we don't apply then when doing a `DF.group_by/2`. We use the groups list to perform operations lime `summarise`. But we really don't have the "grouped data" before performing that operation, right?
- If so, why do we need to perform the grouping, taking the indexes for each group, applying in the intermediate DFs and doing the ungrouping for operations like `mutate`? Wouldn't be easier to just ignore the grouping and mutate the "original" DF?
- My main question: if grouping is virtual (until we summarise), should we even consider the list of groups for operations like `mutate` or `arrange`? Could we try to "mutate" the original DF keeping the groups intact?

I added  a test to `mutate/2`: https://github.com/elixir-nx/explorer/pull/238/commits/363059337134038a809105f02a1e542109ba0927
The output DF is something like if it was not using groups.

I'm sorry if this is a stupid questionnaire :grimacing: 
I'm really confused with the usage of groups.

josevalim:
So maybe let’s leave the group operations out of this particular refactor and perhaps work on their implementation next?

josevalim:
Actually, I think my suggestion is wrong. We can change those functions in this refactor, since the compatibility with mutate and friends is really about additional functionality. :)

however, I do have some questions about how dplyr manages the group/ungrouped columns. For example, what happens in your code after the first mutate? The other ungrouped columns now no longer make sense to be displayed, because they haven’t been “summarized”, or do they?

also what happens if I try to mutate a grouped dataset but the mutation is not a summarization operation such as cumsum? Say we mutate it so the new column is a negation of current values, does dplyr raise because operation then is not a grouped one? Or does it add a new regular “ungrouped” column?

I guess the question is: because group_by is “kind of a no-op”, as it just annotates the groups and still shows the full data, when does the group become concrete and effectively change the dataframe? Or does it always keep grouped and ungrouped stuff together?

philss:
@cigrainger awesome! Thank you for this great explanation! :purple_heart: 
I'm going to add more tests for those operations against grouped DFs.

@josevalim yeah, I think we can go with the refactor. I think this won't impact :)

> when does the group become concrete and effectively change the dataframe?

For what I could understand, this happens when each operation is applied. For example, the `mutate` does the grouping before adding the new column (or change an existing one). Is that correct, Chris?

cigrainger:
@josevalim `mutate` and verbs other than `summarise` don't aggregate, so you keep all columns. You never need to change the inspect representation. The reason I tracked the groups in the struct is so we can pattern match and _each verb_ does a split-apply-combine. Split on the groups, then apply the transform to each resultant datafframe. Once the combine step is done (with a row stack op), then you just have a regular dataframe again (keeping track of the groups in the struct). There are some things to think about re: implicit vs explicit ungrouping. But the groups are never really 'actualised' in any meaningful way outside of the operation.

> For example, what happens in your code after the first mutate? The other ungrouped columns now no longer make sense to be displayed, because they haven’t been “summarized”, or do they?

So the key here is that `mutate` (and `filter` and others) differs from `summarise` in that you don't need to drop other columns because you're not reducing length. `summarise` does drop everything that isn't a group or summarised result. And see the `.groups` argument [here](https://dplyr.tidyverse.org/reference/summarise.html) for more info on ungrouping after a `summarise` operation.

josevalim:
I got it. So have the group information but that doesn't change the dataframe in any way. For example, the mutate operations above will add the same value to each element in the group.

A couple additional questions:

1. Do you have something like "pandas" get_group to get a representation of a particular group?

2. In the example above that you gave, what happens when you want to get the resulting the dataframe with the groups and the associated `n_prev_participated`? Do you just call summarize with the first/last/mean of `n_prev_participated` (given all values in the same group are the same, all of those operations will return the same result)?

cigrainger:
> Do you have something like "pandas" get_group to get a representation of a particular group?

Nah I've never seen the need when you can just use `filter`.

> In the example above that you gave, what happens when you want to get the resulting the dataframe with the groups and the associated n_prev_participated? Do you just call summarize with the first/last/mean of n_prev_participated (given all values in the same group are the same, all of those operations will return the same result)?

`n_prev_participated` varies per row per group, so the values aren't all the same. The cumsum operation happens over the column within the group.

josevalim:
Oh, you are right. It is a cumsum. All clear now, thank you!

cigrainger:
@philss @josevalim I forgot to mention something quite important, which is that I missed this (or didn't grok it) when I was first setting up grouped operations, but this is the direction I'm planning to go: https://docs.rs/polars/0.21.1/polars/frame/groupby/struct.GroupBy.html#method.apply. That is, my intent is to groupby then apply a closure of the intended operation.

lib/explorer/data_frame.ex
```diff
@@ -1080,8 +1096,51 @@ defmodule Explorer.DataFrame do
...
+
+  defp df_for_mutations(df, mutations) do
+    names = df.names
+    mut_names = mutations |> Map.keys() |> MapSet.new()
+
+    new_names =
+      names
+      |> MapSet.new()
+      |> MapSet.union(mut_names)
+      |> MapSet.to_list()
```
josevalim, line 1113: Using MapSet and Map is problematic here because they will alphabetically reorder the columns. You can add a test where you have a dataframe with keys `[:e, :c, :a]` and you mutate to add the columns `[d: 1, b: 2]`, you should end up with `[:e, :c, :a, :d, :b]`. Note the `Map.new` call in the clause above is also part of the problem.

But I also think this code can be simpler. I think you can do this:

    mut_names = Enum.map(mutations, &elem(&1, 0))
    new_names = Enum.uniq(df.names ++ mut_names)
    mutations = Map.new(mutations)

Something else that I would add is to handle the anonymous function within this function so Explorer.DataFrame does not have to handle it. So overall something like this:

    # First convert mutations to a list and resolve anonymous functions
    mutations =
      for {key, value} <- mutations do
        {key, if(is_function(value), do: value.(df), else: value)}
      end

    # Keep the names ordered
    mut_names = Enum.map(mutations, &elem(&1, 0))

    # Compute the mutable types as you have today (except for functions handled above)
    mut_types =
      for {key, value} <- mutations, into: %{} do
      end

    new_names = Enum.uniq(df.names ++ mut_names)
    new_dtypes = Map.merge(df.dtypes, mut_types)

    out_df = %{df | names: new_names, dtypes: new_dtypes}

    # Send the mutations with functions already resolved
    # and simplify the backend code
    Shared.apply_impl(df, :mutate, [out_df, mutations])

lib/explorer/data_frame.ex
```diff
@@ -1755,25 +1823,31 @@ defmodule Explorer.DataFrame do
...
-    end
+    new_dtypes =
+      dtypes
+      |> Map.take(columns)
+      |> Map.put(names_to, :string)
+      |> Map.put(values_to, values_dtype)
+
+    out_df = %{df | names: columns ++ [names_to, values_to], dtypes: new_dtypes}
 
     Shared.apply_impl(df, :pivot_longer, [
```
josevalim, line 1848: Nitpick: you can likely collapse this into a single line now.

lib/explorer/data_frame.ex
```diff
@@ -2261,7 +2404,39 @@ defmodule Explorer.DataFrame do
...
+
+    Shared.apply_impl(df, :summarise, [df_out, Map.new(column_pairs)])
+  end
+
+  defp names_with_dtypes_for_summarise(df, column_pairs) do
+    groups = for group <- df.groups, do: {group, df.dtypes[group]}
+
+    agg_pairs =
+      Enum.flat_map(column_pairs, fn {column_name, aggregations} ->
+        for agg <- aggregations do
```
josevalim, line 2420: A comprehension would be great here:

```elixir
      for {column_name, aggregations} <- column_pairs,
          agg <- aggregations do
```

josevalim (commented):
This is awesome @philss! I have dropped some minor comments.

lib/explorer/data_frame.ex
philss: Maybe this is a good discussion for https://github.com/elixir-nx/explorer/issues/245

philss: I mean: we don't need to decide it here :grimacing:

lib/explorer/data_frame.ex
```diff
@@ -2151,20 +2247,30 @@ defmodule Explorer.DataFrame do
...
   def group_by(df, groups) when is_list(groups) do
-    names = names(df)
-    Enum.each(groups, fn name -> maybe_raise_column_not_found(names, name) end)
+    groups = to_existing_columns(df, groups)
+
+    all_groups =
+      df.groups
+      |> MapSet.new()
+      |> MapSet.union(MapSet.new(groups))
+      |> MapSet.to_list()
```
josevalim, line 2259: I am almost sure that `Enum.uniq(df.groups ++ groups)` will be more efficient. :)

lib/explorer/data_frame.ex
```diff
@@ -998,6 +1012,8 @@ defmodule Explorer.DataFrame do
   Columns are added with keyword list or maps. New variables overwrite existing variables of the
   same name. Column names are coerced from atoms to strings.
 
+  Be aware that mutating grouped dataframes will cause them to lost ordering.
```
josevalim, line 1015: Small suggestion:

```suggestion
  Be aware that mutating grouped dataframes will cause them to lose ordering.
```

Btw, do you know why this happens? Does it happen with dplyr? In other words, is it a bug or something that will indeed always happen?

lib/explorer/data_frame.ex
josevalim: I agree it should be solved with #245. :)

josevalim:
I pushed just two final comments and we should be good!

lib/explorer/data_frame.ex
philss: I'm not sure if happens with `dplyr`, but I will check.

This happens because we have to group the rows, perform the mutation on each group, and then stack the groups together - we lost the order in this step. I don't know if there is a way to keep the original ordering. Probably we could take the indexes of each row, add to a virtual column, sort it and then remove the column :thinking:

josevalim: I see. I think this can actually happen in several cases, not only mutate, so I would rather document that certain operations in groups will change the current ordering.

josevalim (approved):
I added one last suggestion to move the docs around and that's it. Ship it!!! This is great!

lib/explorer/data_frame.ex
cigrainger: Definitely #245. And solving at the DataFrame level would preclude more efficient backend implementations such as groupby + apply in Polars :).

cigrainger (commented):
Love it! Nothing from me. Ship it!

cigrainger (approved):

lib/explorer/data_frame.ex
philss: @josevalim I could fix the ordering here: https://github.com/elixir-nx/explorer/pull/238/commits/5ba2dd11feb1eb5af9599c5cbd4b7df6725a18ed

I don't know how we are going to do for other backends, but at least the default one is fixed :)

PR #244, opened 2022-06-16
DataFrame.table accepts options with limit param instead of single integer

dkuku:
Similar behaviour as IO.inspect:
Also allows to pass additional options to the table in the future

     Explorer.DataFrame.table(df)
     Explorer.DataFrame.table(df, limit: 1)
     Explorer.DataFrame.table(df, limit: :infinity)


lib/explorer/data_frame.ex
```diff
@@ -2273,13 +2274,21 @@ defmodule Explorer.DataFrame do
 
      df = Explorer.Datasets.iris()
      Explorer.DataFrame.table(df)
+     Explorer.DataFrame.table(df, limit: 1)
+     Explorer.DataFrame.table(df, limit: :infinity)
   """
   @doc type: :single
-  def table(df, nrow \\ 5) when nrow >= 0 do
+  @spec table(df :: DataFrame.t(), opts :: Keyword.t() | map()) :: :ok
```
josevalim, line 2281:
```suggestion
  @spec table(df :: DataFrame.t(), opts :: Keyword.t()) :: :ok
```

josevalim (approved):

philss (approved):

cigrainger (approved):
I like it!

PR #248, opened 2022-06-21
Move some parts of the polars DF to native code

philss:
The idea is to simplify some parts, so we don't need to call some functions from the upper level.

### DIFF
--- a/native/explorer/src/dataframe.rs
+++ b/native/explorer/src/dataframe.rs
@@ -400,10 +400,18 @@ pub fn df_to_dummies(
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn df_with_column(data: ExDataFrame, column: ExSeries) -> Result<ExDataFrame, ExplorerError> {
+pub fn df_with_columns(
+    data: ExDataFrame,
+    columns: Vec<ExSeries>,
+) -> Result<ExDataFrame, ExplorerError> {
     let df = &data.resource.0;
     let mut new_df = df.clone();
-    new_df.with_column(column.resource.0.clone())?;
+
+    for column in columns {
+        let series = column.resource.0.clone();
+        new_df = new_df.with_column(series)?.clone();
+    }
+
     Ok(ExDataFrame::new(new_df))
 }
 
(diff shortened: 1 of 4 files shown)


josevalim (approved):
Related to #251. Elixir part looks great for me although @cigrainger should probably review the Rust code. :)

native/explorer/src/dataframe.rs
```diff
@@ -378,17 +390,28 @@ pub fn df_drop_duplicates(
...
+    let new_df = df.select(&selection).and_then(|df| df.to_dummies())?;
     Ok(ExDataFrame::new(new_df))
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn df_with_column(data: ExDataFrame, column: ExSeries) -> Result<ExDataFrame, ExplorerError> {
+pub fn df_with_columns(
+    data: ExDataFrame,
+    columns: Vec<ExSeries>,
+) -> Result<ExDataFrame, ExplorerError> {
```
cigrainger, line 406: I think this one might be more efficient with an hstack.

native/explorer/src/dataframe.rs
```diff
@@ -244,10 +243,23 @@ pub fn df_width(data: ExDataFrame) -> Result<usize, ExplorerError> {
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn df_vstack(data: ExDataFrame, other: ExDataFrame) -> Result<ExDataFrame, ExplorerError> {
+pub fn df_vstack_many(
```
cigrainger, line 246: On this one, I think it might be more something like:

```rust
pub fn df_vstack(
    data: ExDataFrame,
    others: Vec<ExDataFrame>,
    let df = &mut data.resource.0.clone();
    let new_df = others
        .iter()
        .fold(df, |acc, df2| acc.vstack_mut(&df2.resource.0)?);
    Ok(ExDataFrame::new(new_df.clone()))
}
```

(this is untested)

cigrainger (commented):
Looking good! Just a couple suggestions.

native/explorer/src/dataframe.rs
cigrainger: Just realised that won't work with a mutate on the same name as an old column, so this implementation is all good. :)

josevalim (approved):
Ship it!

native/explorer/src/dataframe.rs
cigrainger: Fantastic!

PR #250, opened 2022-06-21
Align `DF.pivot_longer/3` API with dplyr

### LINKED ISSUE #241: Align pivot_longer with dplyr
In dplyr, the `cols` argument to `pivot_longer` is actually the columns we would like to pivot. I find that API a bit more intuitive. Should we change our implementation to align with theirs? :)

philss:
Reference:
https://tidyr.tidyverse.org/reference/pivot_longer.html

Closes #241

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1882,13 +1882,8 @@ defmodule Explorer.DataFrame do
 
     out_df = %{df | names: columns_to_keep ++ [names_to, values_to], dtypes: new_dtypes}
 
-    Shared.apply_impl(df, :pivot_longer, [
-      out_df,
-      columns_to_pivot,
-      columns_to_keep,
-      names_to,
-      values_to
-    ])
+    args = [out_df, columns_to_pivot, columns_to_keep, names_to, values_to]
+    Shared.apply_impl(df, :pivot_longer, args)
   end
 
   @doc """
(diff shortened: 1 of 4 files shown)


lib/explorer/data_frame.ex
```diff
@@ -1756,85 +1761,93 @@ defmodule Explorer.DataFrame do
...
-      |> Map.take(columns)
+      |> Map.take(keep_columns)
       |> Map.put(names_to, :string)
       |> Map.put(values_to, values_dtype)
 
-    out_df = %{df | names: columns ++ [names_to, values_to], dtypes: new_dtypes}
+    out_df = %{df | names: keep_columns ++ [names_to, values_to], dtypes: new_dtypes}
 
-    Shared.apply_impl(df, :pivot_longer, [out_df, value_columns])
+    Shared.apply_impl(df, :pivot_longer, [out_df, columns_to_pivot])
```
josevalim, line 1850: Let's pass `keep`, `names_to`, and `values_to` to the backend, so it doesn't have to compute it. WDYT? Or at least `names_to` and `values_to`.

cigrainger:
@jsonbecker Can I please get your input on the API and naming here? R has been through so many iterations of this I'm not even sure what the pros and cons are anymore.

jsonbecker:
@cigrainger I think having `names_to` and `values_to` are solid as naming conventions. I also like `columns_to_pivot`.

`keep` is interesting. I think of `keep` as defining ids or keys, conceptually, but I think `keep` would be well understood.

lib/explorer/data_frame.ex
philss: yeah, makes sense!

philss:
@jsonbecker cool, I liked the suggestions! Thanks!

cigrainger (approved):
Looks good to me! <3 Ship it!

PR #254, opened 2022-06-21
Simplify `DF.distinct/2` with groups

philss:
This makes the `distinct/2` function consider the columns from groups as
columns to distinct.

This also makes the DF keep the columns of groups, displaying the groups
in the final dataframe.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1245,6 +1245,18 @@ defmodule Explorer.DataFrame do
         x1 integer [1, 3]
         x2 string ["a", "c"]
       >
+
+  If the dataframe has groups, then the columns of each group will be added to the distinct columns.
+
+      iex> df = Explorer.Datasets.fossil_fuels() |> Explorer.DataFrame.group_by("country")
+      iex> Explorer.DataFrame.distinct(df, columns: ["year"])
+      #Explorer.DataFrame<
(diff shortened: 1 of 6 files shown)


philss:
@josevalim @cigrainger do you think this behaviour is correct?
I couldn't make the previous version work with groups, so I don't know if this should work like this.

cigrainger:
Yes the behaviour is correct. I don't think we should tie the top level API to this particular implementation though. I think this one should be more like #253 -- and we should let the different backends handle the implementation.

philss:
> I think this one should be more like https://github.com/elixir-nx/explorer/pull/253 -- and we should let the different backends handle the implementation.

I changed to make this a responsibility of the native code. The only difference is that the columns were in different order, but the result was the same. This looks fine. WDYT?

cigrainger (approved):
Love it! This should end up being way faster too. I think we should start migrating the grouped tests to the grouped test file I created (@josevalim's suggestion) but that doesn't have to happen in this PR.

:shipit:

philss:
> This should end up being way faster too.

I'm not sure if it is faster, but is certainly more clear on both sides now :D 

>  I think we should start migrating the grouped tests to the grouped test file I created (@josevalim's suggestion) but that doesn't have to happen in this PR.

I moved them to this new file. Thanks for the advice!

PR #255, opened 2022-06-22
Do not sort on distinct

josevalim:
The previous implementation were sorting the grouped
datasets on distinct. I also changed the doctest to
use a dataframe with distinct groups.


cigrainger (approved):
🙏

PR #258, opened 2022-06-22
Remove unused introspection callbacks

### LINKED ISSUE #231: n_columns and shape return different values for grouped dataframes

josevalim:
We may need to add some of them back,
but we can wait until there is a use case.

Closes #231.


josevalim:
Actions do not want to run but it only failed the `mix format` test before, which i fixed. :)

josevalim:

PR #268, opened 2022-06-24
Add note about EXPLORER_BUILD environment variable to documentation

joshuataylor:
I'm not 100% happy with the wording.

Without this you receive an error, and you should be able to use it in the config but it's overridden by this variable, should this be changed to be inline with [Rustler precompiled](https://hexdocs.pm/rustler_precompiled/RustlerPrecompiled.html#module-options) `:force_build` setting?


cigrainger:
I don't think this should be under the contributing section -- if you're working with head, it has a `-dev` suffix on the mix.exs version, which will force the build. If you're working with a release, why might you want to compile? What error are you getting?

joshuataylor:
Ah, it was from https://github.com/elixir-nx/explorer/commit/baaea78ec95f7a7a92e59aba21427357437f5759

PR #269, opened 2022-06-24
Upgrade to Polars 0.22.8

joshuataylor:
This allows us the next step of upgrading to support Arrow IPC Streams.

Let me know what the thoughts around window_size_duration is, I think that's right?

### DIFF
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -308,8 +308,8 @@ pub fn s_len(data: ExSeries) -> Result<usize, ExplorerError> {
 #[rustler::nif(schedule = "DirtyCpu")]
 pub fn s_fill_none(data: ExSeries, strategy: &str) -> Result<ExSeries, ExplorerError> {
     let strat = match strategy {
-        "backward" => FillNullStrategy::Backward,
-        "forward" => FillNullStrategy::Forward,
+        "backward" => FillNullStrategy::Backward(None),
+        "forward" => FillNullStrategy::Forward(None),
         "min" => FillNullStrategy::Min,
         "max" => FillNullStrategy::Max,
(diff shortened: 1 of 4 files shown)


native/explorer/src/series.rs
```diff
@@ -308,8 +308,8 @@ pub fn s_len(data: ExSeries) -> Result<usize, ExplorerError> {
 #[rustler::nif(schedule = "DirtyCpu")]
 pub fn s_fill_none(data: ExSeries, strategy: &str) -> Result<ExSeries, ExplorerError> {
     let strat = match strategy {
-        "backward" => FillNullStrategy::Backward,
-        "forward" => FillNullStrategy::Forward,
+        "backward" => FillNullStrategy::Backward(None),
+        "forward" => FillNullStrategy::Forward(None),
```
cigrainger, line 312: Do you know what the enum value is for here? I couldn't find it in the docs.

cigrainger (approved):
I think the window size duration looks right to me. And the tests pass :).

philss (approved):
LGTM! 

For what I could understand, the change related to the `window_size` is correct. :+1:

native/explorer/src/series.rs
joshuataylor: So, had a dig as I was also curious!

Boils down to this issue: https://github.com/pola-rs/polars/issues/3545
Which was implemented here: https://github.com/pola-rs/polars/pull/3559

It's defined in

 https://docs.rs/polars/latest/polars/prelude/enum.FillNullStrategy.html

cigrainger: Ohhhh interesting! Okay! Thanks for finding that :).

PR #272, opened 2022-06-24
Add Cross.toml in order to pass the NIF version to cross

philss:
This is important because `cross` don't have access by default to the
env vars in the system.

This is related to https://github.com/philss/rustler_precompiled/issues/23


cigrainger (approved):
👍

philss:
Would be cool to launch a patch version for this. No rush: it's only affecting a small portion of users.

PR #274, opened 2022-06-25
rename_with instead of rename

ingvardsen:
The pivot example cell with rename fails. To pass the rename function down, it should be `rename_with` I guess.

My humble support for this completely awesome library, and the livebook example just makes it crystal clear.


josevalim:

PR #275, opened 2022-06-25
Use BufWriter/BufReader for reading/writing files

joshuataylor:
It's being used [here](https://github.com/elixir-nx/explorer/blob/main/native/explorer/src/dataframe.rs#L164), so it should be used elsewhere. This should help with reading/writing CSVs/JSONs/etc that can be slower to serialise/deserialise.


philss (approved):

cigrainger (approved):

cigrainger:
LGTM! Thanks @joshuataylor!

PR #282, opened 2022-07-03
Build Dates and DateTimes as maps instead of using NifStruct when encoding as a list

joshuataylor:
This is the first part of the PR, which is just adding dates. Before I go and add Datetimes (the biggest win for this is Datetimes!) I would like to ensure that everyone is happy with this approach.

The change I would probably make is moving/cleaning up the macro?

This benchmark is against ~137k Arrow IPC Streaming dates, with a mix of nulls. It also includes reading the file (the majority of the time for a 28mb file), so having many columns makes the biggest difference.

In the Snowflake arrow project, I see drops the biggest drops around datetimes. Switching to this method is ~4x faster, making parsing go from ~220ms down to 60ms!

```
Operating System: macOS
CPU Information: Apple M1
Number of Available Cores: 8
Available memory: 16 GB
Elixir 1.13.4
Erlang 25.0.2

warmup: 2 s
time: 30 s
memory time: 1 s
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 33 s

Benchmarking with date change ...

Name                       ips        average  deviation         median         99th %
with date change         48.46       20.64 ms    ±24.47%       20.06 ms       33.71 ms
with date struct         27.55       36.30 ms     ±4.15%       36.25 ms       41.39 ms
```

### DIFF
--- a/native/explorer/src/datatypes.rs
+++ b/native/explorer/src/datatypes.rs
@@ -4,9 +4,12 @@ use rustler::resource::ResourceArc;
 use rustler::{Atom, Encoder, Env, NifStruct, Term};
 use std::convert::TryInto;
 
-use std::result::Result;
-
 use crate::atoms;
+use crate::atoms::{calendar, calendar_atom, day, month, year};
+use rustler::types::atom::__struct__;
+use rustler::wrapper::list::make_list;
+use rustler::wrapper::{map, NIF_TERM};
+use std::result::Result;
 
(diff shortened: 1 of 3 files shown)


josevalim (approved):

philss (approved):
Looks great!

cigrainger (approved):
I love it!

PR #288, opened 2022-07-06
Add `filter_with/2` that works with lazy series

philss:
This is an attempt to implement the basic functionality for the `filter_with/2` function
discussed in https://github.com/elixir-nx/explorer/issues/223

This change adds two new opaque backends - one for data frame and one for series.
They work in conjunction to create lazy series and accumulate operations on them before sending to the backend.

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -19,7 +19,7 @@ defmodule Explorer.Backend.DataFrame do
           | [basic_types()]
           | (df() -> series() | basic_types() | [basic_types()])
 
-  @type opaque_ldf :: Explorer.Backend.LazyFrame.t()
+  @type lazy_frame :: Explorer.Backend.LazyFrame.t()
   @type lazy_series :: Explorer.Backend.LazySeries.t()
 
   # IO
@@ -78,7 +78,7 @@ defmodule Explorer.Backend.DataFrame do
   @callback tail(df, rows :: integer()) :: df
   @callback select(df, out_df :: df()) :: df
   @callback filter(df, mask :: series) :: df
-  @callback filter_with(df, (opaque_ldf() -> lazy_series())) :: df
+  @callback filter_with(df, (lazy_frame() -> lazy_series())) :: df
   @callback mutate(df, out_df :: df(), mutations :: [{column_name(), mutate_value()}]) :: df
   @callback arrange(df, columns :: [column_name() | {:asc | :desc, column_name()}]) :: df
   @callback distinct(df, out_df :: df(), columns :: [column_name()], keep_all? :: boolean()) :: df
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -128,9 +128,14 @@ defmodule Explorer.Backend.DataFrame do
   @doc """
   Creates a new DataFrame for a given backend.
   """
-  def new(data, names, dtypes) do
-    dtypes_pairs = Enum.zip(names, dtypes)
-    %DataFrame{data: data, names: names, dtypes: Map.new(dtypes_pairs), groups: []}
+  def new(data, names, dtypes) when is_list(dtypes) do
+    dtypes = Map.new(Enum.zip(names, dtypes))
+
+    new(data, names, dtypes)
+  end
+
+  def new(data, names, dtypes) when is_list(names) and is_map(dtypes) do
+    %DataFrame{data: data, names: names, dtypes: dtypes, groups: []}
   end
 
   @default_limit 5
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -10,6 +10,7 @@ defmodule Explorer.Backend.DataFrame do
   @type series :: Explorer.Series.t()
   @type column_name :: String.t()
   @type dtype :: Explorer.Series.dtype()
+  @type dtypes :: %{column_name() => dtype()}
 
   @typep basic_types :: float() | integer() | String.t() | Date.t() | DateTime.t()
   @type mutate_value ::
@@ -17,6 +18,9 @@ defmodule Explorer.Backend.DataFrame do
           | basic_types()
           | [basic_types()]
           | (df() -> series() | basic_types() | [basic_types()])
+
+  @type opaque_ldf :: Explorer.Backend.LazyFrame.t()
+  @type lazy_series :: Explorer.Backend.LazySeries.t()
 
   # IO
 
@@ -74,6 +78,7 @@ defmodule Explorer.Backend.DataFrame do
   @callback tail(df, rows :: integer()) :: df
   @callback select(df, out_df :: df()) :: df
   @callback filter(df, mask :: series) :: df
+  @callback filter_with(df, (opaque_ldf() -> lazy_series())) :: df
   @callback mutate(df, out_df :: df(), mutations :: [{column_name(), mutate_value()}]) :: df
   @callback arrange(df, columns :: [column_name() | {:asc | :desc, column_name()}]) :: df
   @callback distinct(df, out_df :: df(), columns :: [column_name()], keep_all? :: boolean()) :: df
--- a/lib/explorer/backend/lazy_frame.ex
+++ b/lib/explorer/backend/lazy_frame.ex
@@ -1,0 +1,49 @@
+defmodule Explorer.Backend.LazyFrame do
+  @moduledoc """
+  Represents a dataframe in a lazy format.
+  """
+
+  defstruct dtypes: %{}, names: []
+  alias Explorer.Backend.LazySeries
+
+  @behaviour Explorer.Backend.DataFrame
+
+  @doc false
+  def new(df) do
+    %__MODULE__{names: df.names, dtypes: df.dtypes}
+  end
+
+  @impl true
+  def lazy, do: __MODULE__
+
+  @impl true
+  def to_lazy(ldf), do: ldf
+
+  @impl true
+  def inspect(_ldf, opts) do
+    df = Explorer.DataFrame.new(foo: [1, 2, 3], bar: ["a", "b", "c"])
+    Explorer.Backend.DataFrame.inspect(df, "LazyFrame", nil, opts)
+  end
+
+  @impl true
+  def pull(df, column) do
+    dtype_for_column = df.dtypes[column]
+    LazySeries.new(dtype_for_column, :column, [column])
+  end
+
+  # TODO: Make the functions of non-implemented functions
+  # explicit once the lazy interface is ready.
+  funs =
+    Explorer.Backend.DataFrame.behaviour_info(:callbacks) --
+      (Explorer.Backend.DataFrame.behaviour_info(:optional_callbacks) ++
+         Module.definitions_in(__MODULE__, :def) ++ [{:inspect, 2}])
+
+  for {fun, arity} <- funs do
+    args = Macro.generate_arguments(arity, __MODULE__)
+
+    @impl true
+    def unquote(fun)(unquote_splicing(args)) do
+      raise "cannot perform operation on an Explorer.Backend.LazyFrame"
+    end
+  end
+end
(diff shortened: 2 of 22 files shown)


philss:
I think the `inspect/2` functions could be improved, but I'm not sure how we should inspect those structs. Do you have something in mind?

lib/explorer/backend/data_frame.ex
```diff
@@ -18,6 +19,9 @@ defmodule Explorer.Backend.DataFrame do
           | [basic_types()]
           | (df() -> series() | basic_types() | [basic_types()])
 
+  @type opaque_ldf :: Explorer.Backend.LazyFrame.t()
+  @type lazy_series :: Explorer.Backend.LazySeries.t()
```
josevalim, line 23:
```suggestion
  @type lazy_frame :: Explorer.Backend.LazyFrame.t()
  @type lazy_series :: Explorer.Backend.LazySeries.t()
```

for consistency?

lib/explorer/backend/lazy_frame.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  @impl true
+  def lazy, do: __MODULE__
+
+  @impl true
+  def to_lazy(ldf), do: ldf
+
+  @impl true
+  def inspect(ldf, opts) do
+    Backend.DataFrame.inspect(ldf.data.original, "OpaqueLazyFrame", nil, opts)
```
josevalim, line 30: TODO: We likely want to inspect this as a query expression, instead of the original data frame? I am unsure... I would personally not keep the original dataframe at all.

josevalim, line 30: Do we want to inspect the original dataset? Or just say it is a lazy frame?

I am unsure... Perhaps we should not keep the original dataframe at all.

In any case:

```suggestion
    Backend.DataFrame.inspect(ldf.data.original, "LazyFrame", nil, opts)
```

lib/explorer/backend/lazy_frame.ex
```diff
@@ -0,0 +1,55 @@
+defmodule Explorer.Backend.LazyFrame do
+  @moduledoc """
+  Represents a dataframe in an opaque lazy format.
```
josevalim, line 3: You can add more information that this is automatically created by functions like `filter_with`, `mutate_with`, etc and it is used to build dataframe expressions.

lib/explorer/backend/lazy_series.ex
```diff
@@ -0,0 +1,77 @@
...
+      open,
+      "???",
+      close,
+      line(),
+      color("Operation: ", :atom, opts),
+      color("#{series.data.op}", :atom, opts),
+      line(),
+      color("Args: ", :atom, opts),
+      data
+    ])
```
josevalim, line 59: My suggestion here is to convert the frames to Elixir AST, something like this:

```elixir
def to_elixir_ast(%{op: op, args: args}) do
  {op, meta, Enum.map(args, &to_elixir_ast/2)}
end

def to_elixir_ast(other) do
  other
end
```

And then you call `Macro.to_string` on it to get a visual representation!

The only question is if we are going to translate things like `equal` to `eq`, `add` to `+`, and so on (I think we should). If so, we do this:


```elixir
@to_elixir_op %{
  add: :+,
  subtract: :-,
  ...,
  equal: :==,
  ...
}

def to_elixir_ast(%{op: op, args: args}) do
  {Map.get(@to_elixir_op, op, op), meta, Enum.map(args, &to_elixir_ast/2)}
end

def to_elixir_ast(other) do
  other
end
```

lib/explorer/data_frame.ex
```diff
@@ -1010,6 +1010,22 @@ defmodule Explorer.DataFrame do
...
       )
 
+  @doc false
+  def filter_with(df, fun) when is_function(fun) do
+    ldf =
+      df
+      |> Explorer.Backend.LazyFrame.new()
+      |> Explorer.Backend.DataFrame.new(
+        df.names,
+        Enum.map(df.names, fn name -> df.dtypes[name] end)
```
josevalim, line 1020: Hrm... why not pass `df.dtypes` directly?

lib/explorer/data_frame.ex
```diff
@@ -1010,6 +1010,22 @@ defmodule Explorer.DataFrame do
...
+      )
+
+    case fun.(ldf) do
+      %Series{data: %Explorer.Backend.LazySeries{} = data} ->
+        Shared.apply_impl(df, :filter_with, [data])
```
josevalim, line 1025: Raise a good error message if it returns anything else. :)

native/explorer/src/dataframe.rs
```diff
@@ -366,6 +366,16 @@ pub fn df_filter(data: ExDataFrame, mask: ExSeries) -> Result<ExDataFrame, Explo
     }
 }
 
+#[rustler::nif(schedule = "DirtyCpu")]
+pub fn df_filter_with(data: ExDataFrame, ex_expr: ExExpr) -> Result<ExDataFrame, ExplorerError> {
+    let df: DataFrame = data.resource.0.clone();
+    let exp: Expr = ex_expr.resource.0.clone();
```
josevalim, line 372: Do we need to clone? Is it mutable otherwise?

lib/explorer/data_frame.ex
```diff
@@ -1010,6 +1010,22 @@ defmodule Explorer.DataFrame do
...
+    ldf =
+      df
+      |> Explorer.Backend.LazyFrame.new()
+      |> Explorer.Backend.DataFrame.new(
+        df.names,
+        Enum.map(df.names, fn name -> df.dtypes[name] end)
+      )
+
+    case fun.(ldf) do
+      %Series{data: %Explorer.Backend.LazySeries{} = data} ->
```
josevalim, line 1024: Please raise a good error message if it returns anything else. :)

lib/explorer/backend/lazy_frame.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  @impl true
+  def pull(df, column) do
+    dtype_for_column = df.dtypes[column]
+    data = LazySeries.new(:column, [column])
+    Backend.Series.new(data, dtype_for_column)
+  end
+
+  # TODO: Make the functions of non-implemented functions
+  # explicit once the lazy interface is ready.
```
josevalim, line 41: I don't think this TODO applies. We literally just want to raise for everything that is not a pull:

```suggestion
```

lib/explorer/backend/lazy_frame.ex
```diff
@@ -0,0 +1,55 @@
...
+  funs =
+    Backend.DataFrame.behaviour_info(:callbacks) --
+      (Backend.DataFrame.behaviour_info(:optional_callbacks) ++
+         Module.definitions_in(__MODULE__, :def) ++ [{:inspect, 2}])
```
josevalim, line 45:
```suggestion
         Module.definitions_in(__MODULE__, :def)
```

lib/explorer/backend/lazy_frame.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  for {fun, arity} <- funs do
+    args = Macro.generate_arguments(arity, __MODULE__)
+
+    @impl true
+    def unquote(fun)(unquote_splicing(args)) do
+      raise "cannot perform operation on an Explorer.Backend.LazyFrame"
```
josevalim, line 52:
```suggestion
      raise """
      cannot perform operation on an Explorer.Backend.LazyFrame.
      
      The LazyFrame is available inside filter_with, mutate_with, and \
      similar to build query expressions and you cannot perform any \
      operation on them except accessing its series
      """
```

lib/explorer/backend/lazy_frame.ex
josevalim, line 3:
```suggestion
  Represents a lazy dataframe for building query expressions.
  
  The LazyFrame is available inside `filter_with`, `mutate_with`, and
  similar. You cannot perform any operation on them except accessing
  its underlying series.
```

lib/explorer/polars_backend/expression.ex
```diff
@@ -0,0 +1,37 @@
...
+  def to_expr(%LazySeries{op: :column, args: [name]}) do
+    Native.expr_column(name)
+  end
+
+  def to_expr(%LazySeries{op: :equal, args: [left, right]}) do
+    left = to_expr(left)
+    right = to_expr(right)
+
+    Native.expr_equal(left, right)
+  end
```
josevalim, line 23: We can automate the generation of these functions, so ping me later! :D

lib/explorer/data_frame.ex
philss: Yeah, I think we can change that :+1:

lib/explorer/backend/lazy_series.ex
philss: Cool! I will do that. Thanks!

native/explorer/src/dataframe.rs
philss: It's because otherwise we would need to borrow, which is more complicated to handle.

The clone here is not really cloning: it's incrementing a [ResourceArc](https://docs.rs/rustler/latest/rustler/resource/struct.ResourceArc.html) for the references to this value. So this is cheap :)

lib/explorer/backend/lazy_frame.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  @type t :: %__MODULE__{
+          dtypes: Backend.DataFrame.dtypes(),
+          names: Backend.DataFrame.column_name()
+        }
+  @behaviour Backend.DataFrame
+
+  @doc false
+  def new(df) do
+    %__MODULE__{names: df.names, dtypes: df.dtypes, original: df}
```
josevalim, line 23: I don't think we should keep the original.

josevalim (approved):

lib/explorer/backend/lazy_series.ex
philss: I did with `Code.quoted_to_algebra/1` in order to preserve formatting. Is this OK?

philss:
@josevalim I changed some points we discussed and added more comparison operations.
I'm going to merge this. If you think there is something we can change, I can fix in another PR :)

lib/explorer/backend/lazy_series.ex
```diff
@@ -0,0 +1,113 @@
...
+  # accepts Series or number on the right-hand side.
+  for op <- @comparison_ops_with_numbers do
+    @impl true
+    def unquote(op)(%Series{} = left, %Series{} = right), do: unquote(op)(left, right.data)
+
+    def unquote(op)(%Series{} = left, value) do
+      data = new(unquote(op), [left.data, value])
+
+      Backend.Series.new(data, :boolean)
+    end
```
josevalim, line 45: I wonder if we should make this explicit?

```elixir
    def unquote(op)(%Series{data: left}, %Series{data: right}) do
      data = new(unquote(op), [left, right])
      Backend.Series.new(data, :boolean)
    end

    def unquote(op)(%Series{data: data}, value) when is_number(value) do
      data = new(unquote(op), [data, value])
      Backend.Series.new(data, :boolean)
    end
```

Is there anything we accept besides numbers?

lib/explorer/polars_backend/expression.ex
```diff
@@ -0,0 +1,40 @@
...
+    expr_op = :"expr_#{op}"
+
+    def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
+      Native.unquote(expr_op)(unquote_splicing(updates))
+    end
+  end
+
+  def to_expr(number) when is_integer(number), do: Native.expr_integer(number)
+  def to_expr(number) when is_float(number), do: Native.expr_float(number)
+  def to_expr(binary) when is_binary(binary), do: binary
```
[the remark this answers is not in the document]
philss: We should, but `column` only accepts a raw string, so I kept this for simplicity. I'm going to change in the next PR :)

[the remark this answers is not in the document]
philss: Cool! I couldn't use `definitions_in/2` because it uses the name of the function (to_expr). But I added it manually https://github.com/elixir-nx/explorer/pull/290/files#diff-d28abb6325b1243b94eb349b2f2b8c22b9f775ce5cf5710f4589d4ebf4f2f415R20

PR #292, opened 2022-07-08
Support more dtypes for series in native code

philss:
This is because some operations like aggregations can return dtypes we
were not expecting, specially for integer columns.

Closes https://github.com/elixir-nx/explorer/issues/287

### DIFF
--- a/native/explorer/src/dataframe.rs
+++ b/native/explorer/src/dataframe.rs
@@ -321,7 +321,7 @@ pub fn df_drop_nulls(
     data: ExDataFrame,
     subset: Option<Vec<String>>,
 ) -> Result<ExDataFrame, ExplorerError> {
-    let df = &data.resource.0;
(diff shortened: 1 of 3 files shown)


philss:
I decided to cast all integer series to i64 for simplicity. I think this works for all, except if we have a u64 series.

native/explorer/src/dataframe.rs
```diff
@@ -456,7 +456,7 @@ pub fn df_drop_duplicates(
     maintain_order: bool,
     subset: Vec<String>,
 ) -> Result<ExDataFrame, ExplorerError> {
-    let df = &data.resource.0;
+    let df: &DataFrame = &data.resource.0;
```
philss, line 459: I left some of these hints because my language server was having a bad time without it. Not sure why this is not working for all functions :/

cigrainger (approved):
🎉 Nice one!

PR #293, opened 2022-07-08
add compression for writing parquet

### LINKED ISSUE #291: [Question] Compression settings support for write_parquet
Hi!
Thanks for a fantastic library.
I'm wondering if the parquet write method supports compression settings?
If the underlying polars library support it that is.

It seems the python wrapper has an api for it?

DataFrame.write_parquet(file: str | Path | BytesIO, *, compression: None | Literal['uncompressed', 'snappy', 'gzip', 'lzo', 'brotli', 'lz4', 'zstd'] | str = 'lz4', compression_level: int | None = None, statistics: bool = False, row_group_size: int | None = None, use_pyarrow: bool = False, **kwargs: Any) → None¶

cigrainger:
Adds compression as an option for the parquet writer. Closes #291. For reasons I can't figure out, lz4 and lzo are not supported. @joshuataylor can you see anything I can't re: those two?

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -41,7 +41,7 @@ defmodule Explorer.Backend.DataFrame do
               result(String.t())
 
   @callback from_parquet(filename :: String.t()) :: result(df)
-  @callback to_parquet(df, filename :: String.t()) :: result(String.t())
+  @callback to_parquet(df, filename :: String.t(), compression :: atom()) :: result(String.t())
 
   @callback from_ipc(
               filename :: String.t(),
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -41,7 +41,12 @@ defmodule Explorer.Backend.DataFrame do
               result(String.t())
 
   @callback from_parquet(filename :: String.t()) :: result(df)
-  @callback to_parquet(df, filename :: String.t(), compression :: atom()) :: result(String.t())
+  @callback to_parquet(
+              df,
+              filename :: String.t(),
+              compression :: {nil | atom(), nil | integer()}
+            ) ::
+              result(String.t())
 
   @callback from_ipc(
               filename :: String.t(),
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -52,7 +52,7 @@ defmodule Explorer.Backend.DataFrame do
               filename :: String.t(),
               columns :: list(String.t()) | list(atom()) | list(integer()) | nil
             ) :: result(df)
-  @callback to_ipc(df, filename :: String.t(), compression :: String.t()) ::
+  @callback to_ipc(df, filename :: String.t(), compression :: nil | atom()) ::
               result(String.t())
 
   @callback from_ndjson(
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -349,16 +349,47 @@ defmodule Explorer.DataFrame do
 
   ## Options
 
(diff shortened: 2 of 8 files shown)


lib/explorer/backend/data_frame.ex
```diff
@@ -41,7 +41,7 @@ defmodule Explorer.Backend.DataFrame do
               result(String.t())
 
   @callback from_parquet(filename :: String.t()) :: result(df)
-  @callback to_parquet(df, filename :: String.t()) :: result(String.t())
+  @callback to_parquet(df, filename :: String.t(), compression :: atom()) :: result(String.t())
```
philss, line 44: For what I could understand, this is nullable, right?
```suggestion
  @callback to_parquet(df, filename :: String.t(), compression :: atom() | nil) :: result(String.t())
```

I know `nil` is an atom, but it's good to make it explicit. WDYT?

josevalim:
Maybe those are optional packages we need to add?

josevalim (approved):

joshuataylor:
Can we also pass in the compression level? For example with zstd, you can pass in 1-22, most of the time I use 3 as that's good enough for me.

What isn't working for lz4/lzo, would this work?
```
        "lz4" => ParquetCompression::Lz4,
        "lzo" => ParquetCompression::Lzo,
```

https://docs.rs/polars/0.22.8/polars/prelude/enum.ParquetCompression.html

philss (approved):
Just a minor detail, LGTM! :shipit:

lib/explorer/backend/data_frame.ex
cigrainger: Good point :)

joshuataylor:
Using `parquet` in polars should opt you in for `arrow/io_parquet_compression` which is needed for compression.

cigrainger:
> Maybe those are optional packages we need to add?

I thought that but couldn't find a related flag. It's very possible it's just undocumented.

joshuataylor:
Yeah, default to nil for compression level. I'll see if I can replicate the lz4 issue, should work OOTB.

josevalim:
> Good idea. Do you have a suggestion for the API? Maybe just compression_level as an opt, defaulting to nil?

Erlang allows the field to be an integer. But I would go with a tuple: {:foo, 3}. If the user gives an atom, you normalize it to tuple before calling the backend :)

joshuataylor:
Might also be related to this:

https://github.com/apache/parquet-format/blob/master/Compression.md#lz4

Conversation also here about malformed. https://github.com/pola-rs/polars/issues/3929

Might be better to not just support lz4?

Lzo should be supported, not sure why it isn't working, as even with default-features it doesn't work. I'll poke around with polars and see why, but IMHO neither lzo or lz4 should be a blocker for this.

ritchie46:
> => ParquetCompression::Lz4,

You should choose Lz4Raw over Lz4. That is recommended by the format itself and I recall having had problems with Lz4 myself.

joshuataylor (approved):

cigrainger:
How's that now? Not exactly DRY but does the job.

lib/explorer/data_frame.ex
```diff
@@ -398,7 +424,7 @@ defmodule Explorer.DataFrame do
   ## Options
 
     * `compression` - Sets the algorithm used to compress the IPC file.
-      It accepts `"ZSTD"` or `"LZ4"` compression. (default: `nil`)
+      It accepts `:zstd` or `:lz4` compression. (default: `nil`)
```
josevalim, line 427: We should probably normalize `ipc` to accept the same options (and perhaps also lz4raw instead of lz4)?

cigrainger: I'd love to do that but polars doesn't support it: https://docs.rs/polars/latest/polars/prelude/enum.IpcCompression.html

josevalim: I see, we can probably normalize the backend API though, I will try a PR!

Issue #289, opened 2022-07-06
Introduce `_with` APIs


josevalim:
The goal is to introduce `filter_with`, `summarize_with`, `mutate_with`, and `arrange_with`.

## Attack plan

  * [x] Support `filter_with` with row-based series operations
  * [x] Support `summarize_with` with aggregation-based series operations
  * [x] Support `mutate_with` with row, group, and aggregation-based series operations
  * [x] Support `arrange_with`
  * [x] Decide on #224

This will unblock us to fully tackle #223, #227, and #245.

## Complications

`arrange`/`distinct` introduce one particular issue. We have added the `_with` prefix to disambiguate the macro-api from the non-macro API. This was easy because the non-macro API for `mutate`/`summarize`/`filter` are function based. However, `arrange`/`distinct` already have a non-macro API that is not function based, for example:

    arrange(df, desc: "my_field")

But we also want to support this:

    arrange(df, desc: my_field)

We have three choices:

  * Keep `arrange(df, desc: "my_field")` and `arrange(df, desc: my_field)`, under the same function/arity. This may be doable but it may also raise ambiguities. For example, should we allow `arrange(df, desc: my_field, asc: "another-field")`?

  * Move the non-macro API to `arrange_with`, which will support keywords or functions, such as `arrange_with(df, desc: "my_field")`

  * Remove the `arrange(df, desc: "my_field")` version. People can either use `arrange(df, desc: my_field)` or `arrange_with(df, fn df -> [desc: df["my_field"]] end)`

EDIT: `distinct` has further complications, because the columns are passed as options and we will have to revisit that.

josevalim:
I think the arithmetic row-based operations are pending: add, subtract, pow (rename to power for Nx compatibility?), and so on.

I also propose to rename `nil?` and `not_nil?` to `is_nil` and `is_not_nil` because they don't return booleans themselves, but rather change the column. `is_nil` makes them look like guards but we thought it is less confusing in Nx and went this route.

I think we can also support `coalesce/2` as row-based.

PR #295, opened 2022-07-20
Add missing lazy operations for `DF.filter_with/2`

philss:
These operations are useful to filter for null values.


philss:
I believe there is no other [callback](https://github.com/elixir-nx/explorer/blob/4f118e366cfa46d993a07c856e67ed7a37c4ef3f/lib/explorer/backend/series.ex#L73) missing for `filter_with/2` to work properly. WDYT? 

Related to #289

cigrainger (approved):
I think that's right! This looks great. :)

PR #298, opened 2022-07-24
Add arithmetic operations to LazySeries

philss:
This also changes `DF.filter_with/2` to require a boolean operation at
the "root level".

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -24,9 +24,16 @@ defmodule Explorer.Backend.LazySeries do
     is_nil: 1,
     is_not_nil: 1,
     binary_and: 2,
-    binary_or: 2
+    binary_or: 2,
+    add: 2,
+    subtract: 2,
+    multiply: 2,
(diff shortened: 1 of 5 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -59,6 +66,38 @@ defmodule Explorer.Backend.LazySeries do
     end
   end
 
+  for op <- @arithmetic_operations do
+    @impl true
+    def unquote(op)(%Series{} = left, value_or_series) do
```
josevalim, line 71: Should we be enforcing somehow that those are series with the LazySeries as backend? Maybe have a helper called `lazy_series!(series)`, that receives a series, checks the argument is a series with lazy backend, and return the data?

josevalim (approved):
Looks great, just one minor comments that can be tackled in a next PR!

lib/explorer/backend/lazy_series.ex
philss: Good point! I'm going to change that in a following PR.

PR #299, opened 2022-07-25
Use nightly version with a specific date

philss:
This is needed because the build of arrow2 is broken after this version.
Using nightly with a date is more predictable too.


philss:
This should fix the build for #298

cigrainger (approved):
Good call 👍

PR #301, opened 2022-07-26
Add `DF.summarise_with/2` that accepts lazy series

philss:
This is a version of `summarise` that works with lazy series in a
callback function. The main difference is that more complex expressions
can be made, and each aggregation must be named.

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -83,7 +83,7 @@ defmodule Explorer.Backend.DataFrame do
   @callback tail(df, rows :: integer()) :: df
   @callback select(df, out_df :: df()) :: df
   @callback filter(df, mask :: series) :: df
-  @callback filter_with(df, (lazy_frame() -> lazy_series())) :: df
+  @callback filter_with(df, out_df :: df(), lazy_series()) :: df
   @callback mutate(df, out_df :: df(), mutations :: [{column_name(), mutate_value()}]) :: df
   @callback arrange(df, columns :: [column_name() | {:asc | :desc, column_name()}]) :: df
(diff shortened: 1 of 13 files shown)


philss:
PS: this still WIP. I'm going to finish tomorrow.
But would be nice to have your opinion on how the API should look like. Is this what you expected? 
I mean, like in the test: https://github.com/elixir-nx/explorer/pull/301/files#diff-a8dbae367a98eddc4049331a0ddc27690a3c9441ce21c663024f7d917cce9df5R229-R235

josevalim:
Beautiful!

philss:
@josevalim I think this is it. But we have some operations that don't work only for aggregations, like `:first, :last and :count`, that I skipped. We also have the `:n_unique` aggregation that does not have a correspondent function at `Series`, but have `n_distinct` that seems to do the same - we may need to rename one of them. I'm planning to tackle these missing gaps in another PR. WDYT?

josevalim:
@philss what do you mean about first/last/count not working only for aggregations? where else are they allowed?

josevalim (approved):

josevalim:
Btw, let's go with `n_distinct` so it mirrors `DataFrame.distinct` (confirmed with @cigrainger). There are also other renames in https://github.com/elixir-nx/explorer/issues/289#issuecomment-1191134730, so we can likely tackle it all at once.

philss:
> what do you mean about first/last/count not working only for aggregations? where else are they allowed?

@josevalim they are allowed in the common series. What I wanted to say is that they are not exclusive for aggregations. So using, for example, `Series.count(ldf["column"])` can work for both grouped and ungrouped DFs.

PR #302, opened 2022-07-27
Add lazy series operations that maybe work for aggregations

philss:
This makes support of operations like: first, last and count work for
both `filter_with/2` and `summarise_with/2`


philss:
After we talked (me and José), we decided to go in another direction, so I'm closing this one.

PR #303, opened 2022-07-29
Add missing lazy series operations for aggregations

philss:
This change adds support for operations like: `first`, `last` and `count`
to `filter_with/2` and `summarise_with/2`

Also this:
- removes the restriction for aggregated operations on `filter_with/2`
- renames `n_unique` to `n_distinct`
- renames `nil?/not_nil?` to `is_nil/is_not_nil` (Series)
- fix a small bug in the `Series.coalesce/2` function 
- adds `coalesce` operation to lazy series

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -40,14 +40,18 @@ defmodule Explorer.Backend.LazySeries do
     median: 1,
     var: 1,
     std: 1,
-    quantile: 2
+    quantile: 2,
+    # Maybe aggregations
+    first: 1,
+    last: 1,
+    count: 1
   ]
 
   @comparison_operations [:eq, :neq, :gt, :gt_eq, :lt, :lt_eq]
 
   @arithmetic_operations [:add, :subtract, :multiply, :divide, :pow]
 
-  @aggregation_operations [:sum, :min, :max, :mean, :median, :var, :std]
+  @aggregation_operations [:sum, :min, :max, :mean, :median, :var, :std, :count]
 
   @doc false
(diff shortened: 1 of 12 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -40,14 +40,20 @@ defmodule Explorer.Backend.LazySeries do
     median: 1,
     var: 1,
     std: 1,
-    quantile: 2
+    quantile: 2,
+    # Maybe aggregations
```
josevalim, line 44: Nitpick: I think they are always aggregations. Maybe the name aggregation is confusing here but the idea is that they only make sense when considering multiple rows, not a single one.

```suggestion
```

lib/explorer/backend/lazy_series.ex
```diff
@@ -40,14 +40,20 @@ defmodule Explorer.Backend.LazySeries do
...
+    first: 1,
+    last: 1,
+    count: 1,
+    # Slice and dice
+    coalesce: 2
```
josevalim, line 49:
```suggestion
    # Slice and dice
    coalesce: 2
```

I would move this to the before "Aggregations" and we will likely have a bunch more operations like this in the future. We will have starts_with, contains, ends_with and so on.

native/explorer/src/lib.rs
```diff
@@ -133,6 +133,12 @@ rustler::init!(
         expr_var,
         expr_quantile,
         expr_alias,
+        // maybe agg expressions
```
josevalim, line 136: Same comments here.

josevalim (approved):
Beautiful, I have dropped only some very minor comments.

PR #304, opened 2022-08-05
Upgrade to polars 0.23

joshuataylor:
This means we can merge https://github.com/elixir-nx/explorer/pull/270 !

Lots of great changes from arrow2: https://github.com/jorgecarleitao/arrow2/releases/tag/v0.13.0

And from polars: https://github.com/pola-rs/polars/releases/tag/rust-polars-v0.23.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-07-22
+          toolchain: nightly-2022-07-24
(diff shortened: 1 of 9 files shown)


josevalim:
Awesome!

do we also need to bump the SIMD library or something?

philss:
> do we also need to bump the SIMD library or something?

Maybe we need to bump the nightly version we are using. Arrow2 is using a version that is 2 days newer then ours: https://github.com/jorgecarleitao/arrow2/blob/3b9d86bfb65bc786bf9315a41737a77895716443/.github/workflows/test.yml#L186

cigrainger:
Yep that sounds right to me @philss.

joshuataylor:
Cool, yeah bumped to that version and seems to be good now.

PR #305, opened 2022-08-05
Normalize how column | columns are handled

josevalim:
Make sure that we support column | columns on as many
places as possible. In addition, make sure that all of
them can be one of:

  * a list of columns indexes or names as atoms and strings

  * a range

  * a one-arity function that receives column names and returns
    true for column names to keep

  * a two-arity function that receives column names and types and
    returns true for column names to keep

The fourth is a new addition in this PR that can be useful
in functions like dummy.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -27,10 +27,8 @@ defmodule Explorer.Series do
 
   @valid_dtypes Explorer.Shared.dtypes()
 
-  @type data :: Explorer.Backend.Series.t()
   @type dtype :: Explorer.Backend.Series.dtype()
-  @type t :: %Series{data: data, dtype: dtype()}
-  @type lazy :: %Series{data: Explorer.Backend.LazySeries.t(), dtype: dtype()}
+  @type t :: %Series{data: Explorer.Backend.Series.t(), dtype: dtype()}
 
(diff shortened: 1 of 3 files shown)


lib/explorer/series.ex
```diff
@@ -27,10 +27,8 @@ defmodule Explorer.Series do
 
   @valid_dtypes Explorer.Shared.dtypes()
 
-  @type data :: Explorer.Backend.Series.t()
   @type dtype :: Explorer.Backend.Series.dtype()
-  @type t :: %Series{data: data, dtype: dtype()}
-  @type lazy :: %Series{data: Explorer.Backend.LazySeries.t(), dtype: dtype()}
```
philss, line 33: I think this type is being used by [`Explorer.Backend.Series`](https://github.com/elixir-nx/explorer/blob/0b446a7cb833aebe36514704340ba0ca295c4d7b/lib/explorer/backend/series.ex#L11)

philss (approved):
Awesome! Just one little comment. :shipit:

lib/explorer/series.ex
josevalim: Tks, fixed!

cigrainger (approved):
<3 🚢

josevalim:

PR #309, opened 2022-08-11
Ensure we can compare boolean series

cigrainger:
Came across this one in my own work. I was unable to compare (`equal/2` and `not_equal/2`) boolean series. I don't think we need any guard on `valid_for_bool_mask_operation?` if the dtypes are the same, no?


josevalim (approved):
LGTM! Although I would probably add an actual test instead of relying on doctests. :)

philss (approved):

PR #311, opened 2022-08-12
Fix rename of columns after a summarise

philss:
We can´t be sure the order of columns after a summarise is performed by
polars. So using a proper `rename` makes sure that we are rename the
correct - named - column.

Fixes https://github.com/elixir-nx/explorer/issues/310

### DIFF
--- a/test/explorer/data_frame/grouped_test.exs
+++ b/test/explorer/data_frame/grouped_test.exs
@@ -109,7 +109,7 @@ defmodule Explorer.DataFrame.GroupedTest do
              }
     end
 
-    test "pivote_wider and then summarise with a rename" do
+    test "pivot_wider and then summarise with a rename" do
       df =
         DF.new(
           names: ["cou", "adv", "spo", "cou", "adv", "spo"],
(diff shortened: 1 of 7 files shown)


test/explorer/data_frame/grouped_test.exs
```diff
@@ -109,6 +109,26 @@ defmodule Explorer.DataFrame.GroupedTest do
              }
     end
 
+    test "pivote_wider and then summarise with a rename" do
```
josevalim, line 112:
```suggestion
    test "pivot_wider and then summarise with a rename" do
```

josevalim:
Maybe as a fix it would be better if we guaranteed that summarize has the given order? Or is that not possible? We may have a similar issue with mutate too, as it has a similar API.

josevalim:
Apparently polars does not guarantee an order, so I think this is the correct approach. I will just push a tiny improvement.

josevalim (approved):
The previous code was optimized to rename all columns. Now that we always pass `[{old, new}]`, I optimized it to this new version. :)

philss:
@josevalim awesome! I can't approve my own PR, but I assume I can merge now, right? :smile:

PR #313, opened 2022-08-17
Add window function operations to LazySeries

philss:
This is related to https://github.com/elixir-nx/explorer/issues/289

All the `window_` and `cumulative_` function were added. We also have a restriction for `summarise`, to only work with window function if they are not at the root of the operation tree.

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -9,6 +9,7 @@ defmodule Explorer.Backend.LazySeries do
 
   @behaviour Explorer.Backend.Series
 
+  # TODO: Validate if the window field is really required once we have distinct_with/arrange_with
   defstruct op: nil, args: [], aggregation: false, window: false
 
   @type t :: %__MODULE__{op: atom(), args: list(), aggregation: boolean(), window: boolean()}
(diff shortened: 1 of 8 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -106,26 +120,62 @@ defmodule Explorer.Backend.LazySeries do
...
+    @impl true
+    def unquote(op)(%Series{} = series, window_size, opts) do
+      weights = Keyword.fetch!(opts, :weights)
+      min_periods = Keyword.fetch!(opts, :min_periods)
+      center = Keyword.fetch!(opts, :center)
+
+      args = [lazy_series!(series), window_size, weights, min_periods, center]
+
+      if aggregations?(args) do
+        raise "it's not possible to have an aggregation operation inside a window function"
```
josevalim, line 141:
```suggestion
        raise "it's not possible to have an aggregation operation inside #{op}, which is a window function"
```

lib/explorer/backend/lazy_series.ex
```diff
@@ -9,9 +9,9 @@ defmodule Explorer.Backend.LazySeries do
 
   @behaviour Explorer.Backend.Series
 
-  defstruct op: nil, args: [], aggregation: false
+  defstruct op: nil, args: [], aggregation: false, window: false
```
josevalim, line 12:
```suggestion
  # TODO: Validate if the window field is really required once we have distinct_with/arrange_with
  defstruct op: nil, args: [], aggregation: false, window: false
```

josevalim (approved):

PR #315, opened 2022-08-18
Improve datasets reads by caching them

philss:
This change makes the read of datasets (from the Explorer.Datasets
module) way faster, because it avoids touching the filesystem too often.

### Before
before-persistent-term-explorer

### After

after-persistent-term-explorer-2

### DIFF
--- a/lib/explorer/datasets.ex
+++ b/lib/explorer/datasets.ex
@@ -56,8 +56,7 @@ defmodule Explorer.Datasets do
     # Persistent term is used as a cache, in order to avoid
     # several calls to the filesystem. This is mostly useful
     # to speed up reads in tests.
-    :persistent_term.get(key, nil)
-    |> then(fn
+    case :persistent_term.get(key, nil) do
       nil ->
         @datasets_dir
         |> Path.join("#{name}.csv")
@@ -70,6 +69,6 @@ defmodule Explorer.Datasets do
       {names, df_as_map} ->
         data = for name <- names, do: {name, Map.fetch!(df_as_map, name)}
         DataFrame.new(data)
-    end)
+    end
   end
 end
--- a/lib/explorer/datasets.ex
+++ b/lib/explorer/datasets.ex
@@ -51,8 +51,25 @@ defmodule Explorer.Datasets do
   def iris, do: read_dataset!("iris")
 
   defp read_dataset!(name) do
-    @datasets_dir
-    |> Path.join("#{name}.csv")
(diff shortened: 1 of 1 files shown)


lib/explorer/datasets.ex
```diff
@@ -51,8 +51,25 @@ defmodule Explorer.Datasets do
...
+    # to speed up reads in tests.
+    :persistent_term.get(key, nil)
+    |> then(fn
+      nil ->
+        @datasets_dir
+        |> Path.join("#{name}.csv")
+        |> DataFrame.from_csv!()
+        |> tap(fn df ->
+          df_as_map = DataFrame.to_columns(df)
+          :persistent_term.put(key, {df.names, df_as_map})
```
josevalim, line 67: Why do you store names and columnes? Wouldn't it be better to store the thing as is?

lib/explorer/datasets.ex
```diff
@@ -51,8 +51,25 @@ defmodule Explorer.Datasets do
...
+        end)
+
+      {names, df_as_map} ->
+        data = for name <- names, do: {name, Map.fetch!(df_as_map, name)}
+        DataFrame.new(data)
+    end)
```
josevalim, line 73:
```suggestion
    case :persistent_term.get(key, nil) do
      nil ->
        @datasets_dir
        |> Path.join("#{name}.csv")
        |> DataFrame.from_csv!()
        |> tap(fn df ->
          df_as_map = DataFrame.to_columns(df)
          :persistent_term.put(key, {df.names, df_as_map})
        end)

      {names, df_as_map} ->
        data = for name <- names, do: {name, Map.fetch!(df_as_map, name)}
        DataFrame.new(data)
    end
```

lib/explorer/datasets.ex
philss: The problem is that we don't have the order of the columns, so I'm storing that too.
If I read the map directly, then we may have some tests failing because they rely on ordering.

cigrainger (approved):
Oh cool!!

lib/explorer/datasets.ex
josevalim: I mean, can’t you store the whole dataframe instead? :)

philss: @josevalim it can be! I changed in a810ffe
My previous intuition was that it could be bad for changes in the `DF` structure in the future. But I guess we don't need to worry about that.

josevalim (approved):

PR #316, opened 2022-08-19
Rename `DF.filter/2` to `DF.mask/2`

### LINKED ISSUE #224: Split filter/2 in two functions
@cigrainger I was wondering if we should break `filter/2` in two functions: `mask(dataframe, series)` and `filter(dataframe, callback)`.

The rationale for doing so is that we don't want to encourage `filter(dataframe, series)`, as that won't be efficient as a lazy query. So I thought we should make `filter/2` to be exclusively function based (and `across` based in the future based on #223) and provide a specific `mask/2` function for masking.

Thoughts?

philss:
This is to make clear that we are using boolean masks for filtering in
`DF.mask/2`.

We kept `DF.filter/2` accepting a callback, but that version is
deprecated.

PS: I'm planning to document the new APIs in another PR.

Closes #224


notebooks/exploring_explorer.livemd
```diff
@@ -358,23 +358,32 @@ DataFrame.select(df, &String.ends_with?(&1, "fuel"), :drop)
 
 ### Filter
 
-The next verb we'll look at is `filter`. You can pass in a boolean mask series of the same size as the dataframe:
+The next verb we'll look at is `filter`. Filter can be done by using masks, which
+are boolean lists or series, or functions that allow more complex operations.
+
+You can pass in a boolean mask series of the same size as the dataframe:
```
josevalim, line 364: I think we should not talk about masks. Masks are mostly an anti-pattern. I would remove this section. :)

josevalim (approved):
I dropped comments on the docs and pushed a commit. Code is good!

PR #317, opened 2022-08-19
Rename `take/2` to `slice/2`

philss:
We also added support for ranges in `DF.slice/2` and `Series.slice/2`.

Closes https://github.com/elixir-nx/explorer/issues/307

~~PS: we also have `Series.take/2` which I didn't touch. LMK if you think we should rename that as well.~~ => Renamed also for Series.


lib/explorer/data_frame.ex
```diff
@@ -1932,7 +1940,11 @@ defmodule Explorer.DataFrame do
           )
     end)
 
-    Shared.apply_impl(df, :take, [row_indices])
+    Shared.apply_impl(df, :slice, [row_indices])
+  end
+
+  def slice(df, %Range{} = range) do
+    slice(df, Enum.slice(0..(n_rows(df) - 1), range))
```
josevalim, line 1947: I don't know if we can have an empty dataframe but just in case:

```suggestion
    slice(df, Enum.slice(0..(n_rows(df) - 1)//1, range))
```

josevalim:
@philss let's please do the same for series (i.e. add slice/2 and slice/3). :)

josevalim (approved):

PR #318, opened 2022-08-19
Bump versions of "cache" action and rust at CI files

philss:
This makes the nightly version used in our CI equal to the used in
polars.


josevalim (approved):

philss:
I'm doing a little experiment trying to improve cache usage. But it shouldn't change much.

PR #319, opened 2022-08-19
Raise if `DF.pivot_wider/4` has floats as ID columns

philss:
This change makes more explicit that we are not expecting ID columns as
floats, because floats are harder to compare (`NaN` and `Infinity` things).

This should fix https://github.com/elixir-nx/explorer/issues/312

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2185,9 +2185,11 @@ defmodule Explorer.DataFrame do
     with existing variables. May accept a filter callback, a list or a range of column names.
     Default value is `0..-1`. If an empty list is passed, or a range that results in a empty list of
     column names, it raises an error.
-    ID columns cannot be of the float type. So if there is any column name whose type is float, it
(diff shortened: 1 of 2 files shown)


lib/explorer/data_frame.ex
```diff
@@ -2185,6 +2185,9 @@ defmodule Explorer.DataFrame do
     with existing variables. May accept a filter callback, a list or a range of column names.
     Default value is `0..-1`. If an empty list is passed, or a range that results in a empty list of
     column names, it raises an error.
+    ID columns cannot be of the float type. So if there is any column name whose type is float, it
+    raises an error. In case you need to consider a float column, please try to cast it to an integer
+    or string column before.
```
josevalim, line 2190:
```suggestion

    ID columns cannot be of the float type and attempting so will raise an error.
    If you need to use float columns as IDs, you must carefully consider rounding
    or truncating the column and converting it to integer, as long as doing so
    preserves the properties of the column.
```

PR #323, opened 2022-08-23
Add docs to the `_with` dataframe functions

philss:
This change adds docs to the newer dataframe `_with` APIs (see https://github.com/elixir-nx/explorer/issues/289)

This PR also includes some fixes and include two lazy series operations: `cast` and `n_distinct`.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -74,7 +74,7 @@ defmodule Explorer.DataFrame do
   - `summarise_with/2`
   - `mutate_with/2`
   - `arrange_with/2`
-  
+
   Those functions work by having a "lazy" representation of the dataframe and series,
   which adds the possibility to perform complex operations that are optimized by the backend.
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1146,18 +1146,7 @@ defmodule Explorer.DataFrame do
 
   This function is efficient because it uses a representation of the
   series without pulling them. The only restriction is that
-  you need to use one of the following functions to perform a comparison:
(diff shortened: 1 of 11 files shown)


lib/explorer/data_frame.ex
```diff
@@ -1122,8 +1141,64 @@ defmodule Explorer.DataFrame do
...
   end
 
-  @doc false
-  def filter_with(df, fun) when is_function(fun) do
+  @doc """
+  Picks rows based on a callback function. 
+
+  This function is efficient because it uses a representation of the
+  series without pulling them. The only restriction is that
+  you need to use one of the following functions to perform a comparison:
```
josevalim, line 1149: I don't think it is a matter of the function but rather a matter of returning booleans?

lib/explorer/data_frame.ex
```diff
@@ -1122,8 +1141,64 @@ defmodule Explorer.DataFrame do
...
+        Polars[1 x 2]
+        col1 string ["b"]
+        col2 integer [2]
+      >
+
+      iex> df = Explorer.DataFrame.new(col1: ["a", "b", "c"], col2: [1, 2, 3])
+      iex> Explorer.DataFrame.filter_with(df, fn df -> Explorer.Series.cumulative_max(df["col2"]) end)
+      ** (ArgumentError) expecting the function to return a boolean LazySeries, but instead it returned a LazySeries of type :integer
+     
+      But it's possible to use a boolean operation based on another function:
```
josevalim, line 1186:
```suggestion
  But it's possible to use a boolean operation based on another function:
```

josevalim (approved):

PR #324, opened 2022-08-24
Add more lazy series operations and improve Series docs

philss:
This change adds some missing operations, fix some typespecs and organize
the docs of the `Series` module.

Please let me know if the categorization of functions - at Series - is good for you.

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -20,7 +20,10 @@ defmodule Explorer.Backend.LazySeries do
     reverse: 1,
     argsort: 2,
     sort: 2,
+    distinct: 1,
+    unordered_distinct: 1,
     # Comparisons
+    is_all_equal: 2,
     eq: 2,
     neq: 2,
     gt: 2,
@@ -122,6 +125,22 @@ defmodule Explorer.Backend.LazySeries do
   def sort(%Series{} = s, reverse?) do
     args = [lazy_series!(s), reverse?]
     data = new(:sort, args, aggregations?(args), window_functions?(args))
+
+    Backend.Series.new(data, s.dtype)
+  end
+
+  @impl true
+  def distinct(%Series{} = s) do
+    args = [lazy_series!(s)]
+    data = new(:distinct, args, aggregations?(args), window_functions?(args))
+
+    Backend.Series.new(data, s.dtype)
+  end
+
+  @impl true
+  def unordered_distinct(%Series{} = s) do
+    args = [lazy_series!(s)]
+    data = new(:unordered_distinct, args, aggregations?(args), window_functions?(args))
 
     Backend.Series.new(data, s.dtype)
   end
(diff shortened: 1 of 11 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -47,6 +40,21 @@ defmodule Explorer.Backend.LazySeries do
...
+    sort: 2,
+    distinct: 1,
+    unordered_distinct: 1,
+    slice: 3,
+    head: 2,
+    tail: 2,
+    peaks: 2,
+    fill_missing: 2,
+    fill_missing_with_value: 2,
+    coalesce: 2,
```
josevalim, line 57: I think coalesce, cast, fill_missing and fill_missing_eith_value are all element wise because they consider each element in isolation, right?

lib/explorer/backend/lazy_series.ex
```diff
@@ -299,6 +375,14 @@ defmodule Explorer.Backend.LazySeries do
     Backend.Series.new(data, :boolean)
   end
 
+  @impl true
+  def all_equal?(%Series{} = left, %Series{} = right) do
```
josevalim, line 379: Shouldn’t this be `is_all_equal`?

lib/explorer/series.ex
```diff
@@ -2008,6 +2095,7 @@ defmodule Explorer.Series do
         [4, 2, 5]
       >
   """
+  @doc type: :element_wise
```
josevalim, line 2098: This should be transformation?

lib/explorer/backend/lazy_series.ex
josevalim: Actually, fill missing is window. With value is element wise.

mix.exs
```diff
@@ -65,7 +65,12 @@ defmodule Explorer.MixProject do
         "Functions: Multi-table": &(&1[:type] == :multi),
         "Functions: Row-based": &(&1[:type] == :rows),
         "Functions: Introspection": &(&1[:type] == :introspection),
-        "Functions: IO": &(&1[:type] == :io)
+        "Functions: IO": &(&1[:type] == :io),
+        # For series
+        "Functions: Aggregation": &(&1[:type] == :aggregation),
+        "Functions: Element-wise": &(&1[:type] == :element_wise),
+        "Functions: Transformation": &(&1[:type] == :transformation),
+        "Functions: Window": &(&1[:type] == :window)
```
josevalim, line 73: We are missing introspection! Generate the docs and make sure that “Functions” subcategory does not appear in the Series page.

philss: Introspection was already in use by the `DataFrame` module, so I didn't add it.

lib/explorer/backend/lazy_series.ex
philss: This makes me wonder if we could have this functionality splitted. Today is only one function that accepts both. I was thinking maybe we should have `Series.fill_missing/2` with the strategy, and `Series.fill_missing_with_value/2` for the variation with value. WDYT?

lib/explorer/backend/lazy_series.ex
philss: I renamed to `Series.all_equal/2`, without the `?`.

josevalim (approved):

PR #325, opened 2022-08-25
Fix inspect of Series containing NaN

philss:
Closes https://github.com/elixir-nx/explorer/issues/322

PS: this was inspired in the encoding of dates and datetime. @joshuataylor, if you could take a look would be great! Thanks!

### DIFF
--- a/native/explorer/src/datatypes.rs
+++ b/native/explorer/src/datatypes.rs
@@ -4,8 +4,10 @@ use rustler::resource::ResourceArc;
 use rustler::{Atom, Encoder, Env, NifStruct, Term};
 use std::convert::TryInto;
 
-use crate::atoms::{self, infinity, nan, neg_infinity};
-use crate::atoms::{calendar, calendar_atom, day, hour, microsecond, minute, month, second, year};
+use crate::atoms::{
+    self, calendar, calendar_atom, day, hour, infinity, microsecond, minute, month, nan,
+    neg_infinity, second, year,
+};
 
 use rustler::types::atom::__struct__;
 use rustler::wrapper::list::make_list;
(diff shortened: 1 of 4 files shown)


josevalim (approved):
LGTM but @cigrainger should review the Rust code :)

native/explorer/src/datatypes.rs
```diff
@@ -380,7 +380,37 @@ fn encode_datetime_series<'b>(s: &Series, time_unit: TimeUnit, env: Env<'b>) ->
...
+        .downcast_iter()
+        .flat_map(move |iter| {
+            iter.into_iter().map(move |opt_v| {
+                opt_v
+                    .copied()
+                    .map(|x| {
+                        if x.is_finite() {
+                            x.encode(env)
+                        } else {
+                            match (x.is_nan(), x.is_sign_negative()) {
```
joshuataylor, line 402: If you put the `nan`, `neg_infinity`, `infinity` encode part at the top of the function so it doesn't have to encode each time, as these don't change.

joshuataylor (commented):
Looks great! One note about encoding.

native/explorer/src/datatypes.rs
philss: Oh, nice! I changed the code to fix that. Thank you!

native/explorer/src/datatypes.rs
```diff
@@ -4,7 +4,7 @@ use rustler::resource::ResourceArc;
 use rustler::{Atom, Encoder, Env, NifStruct, Term};
 use std::convert::TryInto;
 
-use crate::atoms;
+use crate::atoms::{self, infinity, nan, neg_infinity};
 use crate::atoms::{calendar, calendar_atom, day, hour, microsecond, minute, month, second, year};
```
cigrainger, line 8: Do we want to combine these two use declarations?

cigrainger (approved):
Looks great!

native/explorer/src/datatypes.rs
philss: I changed that to one `use` declaration. Thanks!

PR #335, opened 2022-09-07
Make dtype for `count` and `n_distinct` aggregations explicit in `summarise/2`

andreyuhai:
The columns we use for `count` or `n_distinct` aggregations are not always integer columns. Hence when we perform those aggregations on columns other than of integers we get the `dtype` the same as the column's `dtype`. For example:

```elixir
iex(31)> df = DF.new(%{id: [1, 2, 3, 1, 1], name: ["foo", "bar", "foo", "bar", "bar"], price: [1.1, 1.2, 1.3, 1.5, 1.2]})
#Explorer.DataFrame<
  Polars[5 x 3]
  id integer [1, 2, 3, 1, 1]
  name string ["foo", "bar", "foo", "bar", "bar"]
  price float [1.1, 1.2, 1.3, 1.5, 1.2]
>

iex(32)> summed = df |> DF.group_by("id") |> DF.summarise(name: [:n_distinct], price: [:count])
#Explorer.DataFrame<
  Polars[3 x 3]
  id integer [1, 2, 3]
  name_n_distinct integer [2, 1, 1]
  price_count integer [3, 1, 1]
>

iex(33)> summed.dtypes
%{"id" => :integer, "name_n_distinct" => :string, "price_count" => :float}
```

This PR just explicitly mentions the `dtype` for `n_distinct` and `count` aggregations since they'll always be the same.

Please let me know what you think and also whether I should write a test for this under `summarise/2` in `grouped_tests`. I was just not sure whether there should be a test case for this one or not as I did not see for the others but I am happy to write.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2935,6 +2935,9 @@ defmodule Explorer.DataFrame do
             :mean ->
               :float
 
+            agg when agg in [:count, :n_distinct] ->
+              :integer
+
             _other ->
               df.dtypes[column_name]
           end


josevalim:
Can you reproduce the problem with `summarize_with`? The current API for summarize is being replaced by `summarize_with`.

andreyuhai:
> Can you reproduce the problem with `summarize_with`? The current API for summarize is being replaced by `summarize_with`.

Only with the `n_distinct` as far as I checked which gives you the `dtype` as `float` but that's not the exact same thing as I described above so I am not sure whether that's expected.👇 

```elixir
iex(47)> df = DF.new(%{id: [1, 2, 3, 1, 1], name: ["foo", "bar", "foo", "bar", "bar"], price: [1.1, 1.2, 1.3, 1.5, 1.2]})
#Explorer.DataFrame<
  Polars[5 x 3]
  id integer [1, 2, 3, 1, 1]
  name string ["foo", "bar", "foo", "bar", "bar"]
  price float [1.1, 1.2, 1.3, 1.5, 1.2]
>
```

#### `Series.n_distinct/1`

```elixir
iex(48)> summed = df |> DF.group_by("id") |> DF.summarise_with(&[name_distinct: S.n_distinct(&1["name"])])
#Explorer.DataFrame<
  Polars[3 x 2]
  id integer [1, 2, 3]
  name_distinct integer [2, 1, 1]
>

iex(49)> summed.dtypes
%{"id" => :integer, "name_distinct" => :float}
```

#### `Series.count/1`

```elixir
iex(52)> summed = df |> DF.group_by("id") |> DF.summarise_with(&[name_count: S.count(&1["name"])])
#Explorer.DataFrame<
  Polars[3 x 2]
  id integer [1, 2, 3]
  name_count integer [3, 1, 1]
>

iex(53)> summed.dtypes
%{"id" => :integer, "name_count" => :integer}
```

josevalim:
@philss we have a bug where the types in the dataframe do not match. I have pushed a commit to main that adds some verification that the polars frame match our actual frame and it does catch this bug. You can set the flag to true in `config/config.exs`. :)

philss:
I'm merging this as it is. I'm going to add the tests in conjunction with what I'm working on (in another PR).
Thanks for the details, @andreyuhai! :purple_heart:

PR #338, opened 2022-09-08
Optimize and simplify encoding

### LINKED ISSUE #279: Encoding is inefficient for series
In the macro [encode](https://github.com/elixir-nx/explorer/blob/main/native/explorer/src/datatypes.rs#L217) and encode_list, what happens is this:

```
macro_rules! encode {
    ($s:ident, $env:ident, $convert_function:ident, $out_type:ty) => {
        $s.$convert_function()
            .unwrap()
            .into_iter()
            .collect::<Vec<Option<$out_type>>>()
            .encode($env)
    };
}
```
It's called using `encode!(s, env, bool)`

So what happens here is:
1. Use .bool() (a Polars helper function) to get a `BooleanChunked`
2. Unwrap the result, then convert to iter
3. Collect into a vector of booleans, which can be None (for nil)
4. Rustler then iterates over the vec, encoding each value inside:
4a.
```rust
        let term_array: Vec<NIF_TERM> = self.iter().map(|x| x.encode(env).as_c_arg()).collect();
        unsafe { Term::new(env, list::make_list(env.as_c_arg(), &term_array)) }
```
6. It's then returned as a Term.

Instead, we could not collect in step 3, and encode the terms ourselves as a list (quick code below).
```rust
        let list = $s.$convert_function()
            .unwrap()
            .into_iter()
            .map(|x| x.map(|d| d as $ty).encode(env).as_c_arg())
            .collect::<Vec<NIF_TERM>>();
        
        unsafe { Term::new(env, list::make_list(env.as_c_arg(), &list)) }
```

This saves a collection, heap allocation, etc. I've found not having Rustler reiterate or rebuild over a collected list is much faster as it doesn't need to be rebuilt. See also [iterators](https://nnethercote.github.io/perf-book/iterators.html) here for a great explanation why.

When optimising the Snowflake Arrow code, this made a big difference for large collections.

I'd be happy to do a PR for this with some benchmarks if people are happy with this.

### LINKED ISSUE #281: Encoding strings using enif_make_sub_binary
When returning a large amount of strings, this can be very slow. See the discussion here https://github.com/rusterlium/rustler/discussions/463.

There was a great idea there to use the offsets and values to create subbinaries. Doing this has dramatically decreased the time it takes to return binaries back to Elixir for Utf8.

Here is what I do:
```rust
                    // This is goofy and stupid, but drastically speeds up converting binaries.
                    // What we do here is take all the values/offsets, create a binary, then subslice into it.
                    // @todo fix this later to be less terrible.
                    let utf8 = series.utf8().unwrap();
                    let len = utf8.len();
                    let mut values = Vec::with_capacity(len);
                    let mut offsets: Vec<usize> = Vec::with_capacity(len);
                    let mut last_offset: usize = 0;

                    for array in utf8.downcast_iter() {
                        let mut offset_loop: usize = 0;
                        for offset in array.offsets().iter().skip(1) {
                            offset_loop = *offset as usize;
                            offsets.push(last_offset + offset_loop);
                        }

                        last_offset += offset_loop;
                        values.extend(array.values().as_slice());
                    }

                    // Can we make a binary straight from the slice?
                    // This **might** be faster, but I think that allocating it this way is okay
                    // as we know the length already?
                    let mut values_binary = NewBinary::new(env, values.len());
                    values_binary.copy_from_slice(values.as_slice());

                    let binary: Binary = values_binary.into();

                    // Now we slice into the binary.
                    let mut last_offset: usize = 0;
                    let mut binaries: Vec<NIF_TERM> = Vec::with_capacity(offsets.len());
                    for offset in offsets {
                        if last_offset == offset {
                            last_offset = offset;
                            binaries.push(nil);
                        } else {
                            let slice =
                                make_subbinary(env, &binary, last_offset, offset - last_offset)
                                    .unwrap()
                                    .as_c_arg();
                            binaries.push(slice);
                            last_offset = offset;
                        }
                    }

                    binaries
```

```rust
pub fn make_subbinary<'a>(
    env: Env<'a>,
    binary: &Binary,
    offset: usize,
    length: usize,
) -> NifResult<Term<'a>> {
    let min_len = length.checked_add(offset);
    if min_len.ok_or(Error::BadArg)? > binary.len() {
        return Err(Error::BadArg);
    }
    let term = binary.to_term(env);

    let raw_term = unsafe {
        rustler_sys::enif_make_sub_binary(
            term.get_env().as_c_arg(),
            term.as_c_arg(),
            offset,
            length,
        )
    };

    Ok(unsafe { Term::new(env, raw_term) })
}
```

I'd be happy to create a PR for this.

josevalim:
Closes #279.
Closes #281.


philss (approved):
Looks great! :sparkles: 

But I would like to hear what @cigrainger and @joshuataylor think :)

cigrainger (approved):
I can't fault this! Keen to hear from @joshuataylor but this looks like an improvement to me.

native/explorer/src/datatypes.rs
```diff
@@ -350,92 +351,122 @@ fn encode_datetime_series<'b>(s: &Series, time_unit: TimeUnit, env: Env<'b>) ->
...
+                .as_c_arg()
         })
         .collect();
 
     unsafe { Term::new(env, make_list(env.as_c_arg(), &items)) }
 }
 
+fn encode_utf8_series<'b>(s: &Series, env: Env<'b>) -> Term<'b> {
+    let utf8 = s.utf8().unwrap();
+    let mut items: Vec<usize> = Vec::with_capacity(utf8.len());
```
joshuataylor, line 390: Looks great, I tried to make this work without pushing to an array, instead returning the map value but couldn't figure it out from my quick look.

We should also add the nil atom here so we don't have to re-encode it later.

```rust
let nil = atom::nil().to_term(env).as_c_arg()
...
                None => items.push(nil),
```

josevalim:

PR #339, opened 2022-09-08
Fix dtype mismatch between Explorer and Polars

philss:
This is a continuation of https://github.com/elixir-nx/explorer/pull/335

There are still two series operations with mismatched types:
- `quantile` returns `float` when it should return the series dtype (e.g. `:integer` in case of a series of ints).
- `pow` has the same problem.

**EDIT**: these two operations are incorrect in Polars. I should open an issue there.


josevalim (approved):

philss:
I think I was not clear in the PR description. The two operations, `pow` and `quantile` are wrong in the Polars side, when using Expressions. Since this is correct in our side, I'm merging and I will open an issue upstream.

Issue #342, opened 2022-09-09
Improvements to encoding


josevalim:
I have just learned that Polars uses [its own reference counting system](https://pola-rs.github.io/polars/polars/prelude/struct.Arc.html). This means that we could directly point to the underlying Polars/Arrow binaries by using make_resource_binary, then bumping the ref count on ARC, and decreasing the ARC refcount when the binary is GCed. Doing so means we don't need to double the allocation for string series.

Something else to explore to reduce allocation is to allow Iterators to be converted to lists. I have implemented this with a macro in #344 but finding a mechanism to encapsulate it would be grand. Perhaps Encoder could be implemented for `DoubleEndedIterator<Item: Encoder>`?

Unfortunately both of these are above my paygrade. :)

_Originally posted by @josevalim in https://github.com/elixir-nx/explorer/issues/281#issuecomment-1241938332_

josevalim:
Another idea to explore is to preload NifStructs keys. Currently we are doing this to speed up encoding, specially in a loop:

https://github.com/elixir-nx/explorer/blob/9166d739e43eee7457e6e93ac7b918cf54f8bee4/native/explorer/src/encoding.rs#L65

However, I think we can do this automatically. There are a couple options:

1. When encoding a Vec<NifStruct>, we automatically encode the keys before hand, as we did manually above. We should also apply this to iterators.

2. When the nif is loaded, [we can use the C load callback to preload already encoded atoms and keys into enif_priv_data](https://www.erlang.org/doc/man/erl_nif.html#initialization). This may require declaring structs as part of `rustler::init`

josevalim:
This is in!

PR #344, opened 2022-09-10
Encode series as lists in reverse

josevalim:
This makes encoding 20% faster and reduces memory usage.

We should find ways to encapsulate this as per #342.


josevalim:
Merging cause I have more changes. :) If there is any feedback, drop it and I will catch it later. :)

josevalim:

PR #346, opened 2022-09-11
Use resource binaries on Series.to_list for strings

josevalim:
Still WIP, depends on https://github.com/rusterlium/rustler/pull/485.

### DIFF
--- a/native/explorer/Cargo.lock
+++ b/native/explorer/Cargo.lock
@@ -1216,6 +1216,7 @@ checksum = "a3f87b73ce11b1619a3c6332f45341e0047173771e8b8b73f87bfeefb7b56244"
 [[package]]
 name = "rustler"
 version = "0.26.0"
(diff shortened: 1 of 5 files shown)


josevalim:
If you folks are OK with cargo temporarily having a Git dependency, this is good to go.

Compared to the latest release, main encodes a utf-8 series to list 10x faster and uses twice less memory.

josevalim:
Note: this PR also has the foundation so we get zero-copy interop with Nx.

philss (approved):
> If you folks are OK with cargo temporarily having a Git dependency, this is good to go.

> Compared to the latest release, main encodes a utf-8 series to list 10x faster and uses twice less memory.

Awesome! :D 
I'm OK with using the GitHub version :)

cigrainger (approved):
Incredible. I'm definitely okay with the git dependency!

josevalim:

PR #349, opened 2022-09-12
Adapt and document operations using groups - part I

philss:
This is following the list of https://github.com/elixir-nx/explorer/issues/245
It covers:

- [x] arrange
- [x] collect
- [x] head
- [x] mutate
- [x] select
- [x] tail

Some questions so far:
- should `select` with `drop` raise when trying to remove a group? In this PR it is no-op.


lib/explorer/data_frame.ex
```diff
@@ -648,6 +648,7 @@ defmodule Explorer.DataFrame do
   This collects the lazy data frame into an eager one, computing the query.
 
   If already eager, this is a noop.
+  Collecting a grouped dataframe should return a grouped dataframe.
```
josevalim, line 651: Just FYI, this will show up as a single paragraph in docs. If you want a separate paragraph, you need two newlines:

```suggestion

  Collecting a grouped dataframe should return a grouped dataframe.
```

The same applies to examples below.

lib/explorer/data_frame.ex
```diff
@@ -1486,6 +1570,28 @@ defmodule Explorer.DataFrame do
         b integer [3, 1, 2]
       >
 
+  When used in a grouped dataframe, arrange is going to sort each group individually and
+  then return the entire dataframe with the existing groups. If one of the arrange columns
+  is also a group, the sorting for that column is not going to work. It is necessary to
+  summarise and arrange to get any effect for that columns.
```
josevalim, line 1576:
```suggestion
  is also a group, the sorting for that column is not going to work. It is necessary to
  first summarise the desired column and then arrange it.
```

lib/explorer/data_frame.ex
```diff
@@ -1219,6 +1297,9 @@ defmodule Explorer.DataFrame do
   Columns are added with keyword list or maps. New variables overwrite existing variables of the
   same name. Column names are coerced from atoms to strings.
 
+  Mutate can work with groups, but it must have the same length, or be a scalar value.
+  If a complex operation with groups is needed, consider using `mutate_with/2`.
+
```
josevalim, line 1302: This function will be removed soon, so don't bother adding this. :D

josevalim (approved):

PR #352, opened 2022-09-13
Change to_* functions to return only :ok

### LINKED ISSUE #347: `to_*` functions should just return `:ok`
The filename is given as input, no need to return them as output.

Ian-GL:
Changing the `to_*` functions to write to files on the Dataframe module to just return `:ok` on success (as the filename is an already an input of the function).

Closes #347


josevalim:

PR #360, opened 2022-09-23
Update polars to v0.24.2

philss:
This is necessary to fix https://github.com/elixir-nx/explorer/issues/358

Closes https://github.com/elixir-nx/explorer/issues/359

PS: there is a small regression for the `from_ndjson/2` function that I'm checking - it is reading some floats wrongly.
PS2: I removed the `summarise` callback because I realize it can be implemented with `summarise_with/2` :D

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -128,8 +128,6 @@ defmodule Explorer.Backend.DataFrame do
 
   # Groups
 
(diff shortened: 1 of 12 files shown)


josevalim (approved):

philss:
@josevalim I'm going to merge the way it is, and prepare a report of the small bug I found in the (ND)JSON parser tomorrow.

Issue #284, opened 2022-07-04
Proposal: Integration testing for CSV, IPC, NDJSON, Parquet


joshuataylor:
I think it would be worthwhile adding tests for CSV, IPC, NDJSON and Parquet, with all having the same data.

On the other hand, [Apache supplies Arrow testing files](https://github.com/apache/arrow-testing) and [Parquet files](https://github.com/apache/parquet-testing) which Arrow2 tests against (Arrow2 is used by polars). We could add these as submodules.

[We could also just borrow the tests from Arrow2 for the various formats](https://github.com/jorgecarleitao/arrow2/tree/main/tests/it/io) :o)

The use case for this is that we'll find any weird issues encoding into/out of Elixir, for example with https://github.com/elixir-nx/explorer/issues/283 . It'll ensure the datatypes work as expected, and if/when there are additional backends the new backend can ensure compatibility if all the tests for various formats pass. 

I would be happy to write these initial tests.

josevalim:
Definitely! :+1:

cigrainger:
For sure! Good idea.

thbar:
I am starting to implement tests for the CSV part, based on the arrow2 test suite, which turns out to be a good inspiration.

josevalim:
We have started on this for CSV and Parquet. More will come as we add/remove features.

PR #362, opened 2022-10-02
Start adding CSV tests

thbar:
This is just a draft to open discussion to implement CSV tests (#284). As I worked on it I had a few a-ha moments (partly due to my discovering of the `from_csv` API).

I used https://github.com/jorgecarleitao/arrow2/tree/main/src/io/csv/read as an initial basis.

I used a data-based approach to iterate and generate tests using `quote`/`unquote`, something I found useful to discover the API, but maybe we'll want something more explicit & more readable (e.g. one describe block per "dtype", manually written).

I have not attempted to cover in full what can be done, partly due to personal bandwidth, and also to get feedback.

Things I have noticed/questions I have that could be of interest to continue this:
- the `quote/unquote` implementation is not very readable, but makes it easy to add an extra sample
- My understanding is that Polars is the default backend, and `from_csv` goes via the backend, so presumably we'll want something that can be made backend-agnostic but still run on all the backends? (I haven't looked much into the backends code at this point)
- There would be a lot of interest to add tests around many more things: a misparsed datetime will result in nil (I believe), is the library failing fast on errors such as a random string in the middle of an int column with an explicitly passed dtype, how does float truncation occurs (I believe extra digits in the CSV column will be just lost at this point), etc etc
- `frame["city"] == frame[0]` returns false (a different reference) and I presume this could be linked to the Rust interop, but haven't looked into it yet

### DIFF
--- a/test/explorer/data_frame/csv_test.exs
+++ b/test/explorer/data_frame/csv_test.exs
@@ -48,39 +48,30 @@ defmodule Explorer.DataFrame.CSVTest do
     assert_in_delta(57.653484, frame["lat"][0], @f64_epsilon)
 
     city = frame["city"]
-    # TODO: write a meaningful test, but references are apparently changing
-    # assert city == frame[0]
 
     assert city[0] == "Elgin, Scotland, the UK"
     assert city[13] == "Aberdeen, Aberdeen City, UK"
   end
 
-  # TODO: expose what is in Explorer.Backend.Series @valid_dtypes attributes, so that
-  # we ensure the tests remain up to date?
-
-  # NOTE: just a preliminary batch of data, we may want to add more than just one value,
-  # with all covered cases.
-  @valid_dtypes [
-    {:integer, "100", 100},
-    {:float, "2.3", 2.3},
-    {:boolean, "true", true},
-    {:string, "some string", "some string"},
-    {:date, "2022-12-01", ~D[2022-12-01]},
-    {:datetime, "2022-10-01T11:34:10.123456", ~N[2022-10-01 11:34:10.123456]}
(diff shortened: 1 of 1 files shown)


josevalim:
> The quote/unquote implementation is not very readable, but makes it easy to add an extra sample

I think you can use functions without loss of behaviour:

    assert_csv :datetime, "\"string\"", "string"

> My understanding is that Polars is the default backend, 

So far we only test the default backend, so I would say this approach is fine!

> frame["city"] == frame[0] returns false

Correct, you want to use Series.equal or similar. Another test you want to consider adding is that we can decode and then encode the CSV and it goes back to the same value. :)

josevalim (approved):

thbar:
Good enough for a first round, thanks @josevalim for the suggestions & review!

I will likely come back with edge-cases (round-trip giving different outcome, what happens with digits/precision, error handling etc).

test/explorer/data_frame/csv_test.exs
```diff
@@ -0,0 +1,107 @@
...
+      assert_csv(:date, "2022-12-01", ~D[2022-12-01])
+      assert_csv(:date, "1960-01-31", ~D[1960-01-31])
+    end
+
+    test "datetime" do
+      assert_csv(:datetime, "2022-10-01T11:34:10.123456", ~N[2022-10-01 11:34:10.123456])
+    end
+
+    @tag :skip
+    test "list"
```
josevalim, line 105:
```suggestion
```

philss (approved):

philss:
@thbar thank you! :purple_heart:

PR #365, opened 2022-10-08
Add `datetime[ns]` dtype

xaviRodri:
Added here a missing dtype needed to import datasets that contain data of this type.

Closes https://github.com/elixir-nx/explorer/issues/364

### DIFF
--- a/native/explorer/src/encoding.rs
+++ b/native/explorer/src/encoding.rs
@@ -145,7 +145,7 @@ fn time_to_microseconds(v: i64, time_unit: TimeUnit) -> i64 {
         TimeUnit::Nanoseconds => (v as f64 * 0.001) as i64,
         _ => unreachable!(),
     }
-  }
+}
 
 #[inline]
 fn encode_datetime(v: i64, time_unit: TimeUnit, env: Env) -> Term {
--- a/native/explorer/src/encoding.rs
+++ b/native/explorer/src/encoding.rs
@@ -138,24 +138,24 @@ fn naive_datetime_struct_keys(env: Env) -> [NIF_TERM; 9] {
 }
 
 #[inline]
-fn time_unit_to_factor(time_unit: TimeUnit) -> f64 {
+fn time_to_microseconds(v: i64, time_unit: TimeUnit) -> i64 {
     match time_unit {
-        TimeUnit::Milliseconds => 1000.0,
-        TimeUnit::Microseconds => 1.0,
-        TimeUnit::Nanoseconds => 0.001,
+        TimeUnit::Milliseconds => v * 1000,
+        TimeUnit::Microseconds => v,
+        TimeUnit::Nanoseconds => (v as f64 * 0.001) as i64,
         _ => unreachable!(),
     }
-}
+  }
 
 #[inline]
 fn encode_datetime(v: i64, time_unit: TimeUnit, env: Env) -> Term {
     let naive_datetime_struct_keys = &naive_datetime_struct_keys(env);
     let calendar_iso_module = atoms::calendar_iso_module().encode(env).as_c_arg();
     let naive_datetime_module = atoms::naive_datetime_module().encode(env).as_c_arg();
-    let factor = time_unit_to_factor(time_unit);
+    let microseconds_time = time_to_microseconds(v, time_unit);
 
     unsafe_encode_datetime!(
-        (v as f64 * factor) as i64,
+        microseconds_time,
         naive_datetime_struct_keys,
         calendar_iso_module,
         naive_datetime_module,
@@ -168,14 +168,15 @@ fn encode_datetime_series<'b>(s: &Series, time_unit: TimeUnit, env: Env<'b>) -> 
     let naive_datetime_struct_keys = &naive_datetime_struct_keys(env);
     let calendar_iso_module = atoms::calendar_iso_module().encode(env).as_c_arg();
     let naive_datetime_module = atoms::naive_datetime_module().encode(env).as_c_arg();
-    let factor = time_unit_to_factor(time_unit);
 
     unsafe_iterator_to_list!(
         env,
         s.datetime().unwrap().into_iter().map(|option| option
             .map(|v| {
+                let microseconds_time = time_to_microseconds(v, time_unit);
+
                 unsafe_encode_datetime!(
-                    (v as f64 * factor) as i64,
+                    microseconds_time,
                     naive_datetime_struct_keys,
                     calendar_iso_module,
                     naive_datetime_module,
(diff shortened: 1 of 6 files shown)


josevalim:
I believe we can merge this but we will probably need other changes on functions like Series.to_list for it to work properly

josevalim:
Maybe it is worth adding a test in the style of #362 for Parquet files?

xaviRodri:
Adding here a comment to track the process..

As you said @josevalim , we had to change more than the update from the first commit.
I finally tested the `from_parquet/2` function with real data and works fine 👍 

I was trying to add some tests for parquet files but I realised that I made some others fail...

Let me put some context:
> When we encode datetimes, we use a multiplication factor (milliseconds -> 1000, microseconds -> 1, and now nanoseconds -> 0.001). In order to add this new one, I had to convert also the incoming microseconds (i64) first into a float (f64) using `v as f64`. Also tried `f64::from(v)`  but "the trait `From<i64>` is not implemented for `f64`".

For some reason I cannot recognise yet, this casting is (in some cases) changing the incoming integer. 
See an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=fn%20main()%20%7B%0A%20%20%20%20let%20int_var%3Ai64%20%3D%2012091941575702789%3B%0A%0A%20%20%20%20let%20converted%3Af64%20%3D%20int_var%20as%20f64%3B%0A%0A%20%20%20%20println!(%22NUM%3A%20%7B%3A.1%7D%22%2C%20converted)%3B%0A%7D) 

```Rust
fn main() {
    let int_var:i64 = 12091941575702789;

    let converted:f64 = int_var as f64;

    println!("NUM: {:.1}", converted);
}

// Prints
NUM: 12091941575702788.0
```

I haven't worked with Rust before, so maybe someone more expert on it could help on this!
Thanks in advance!

philss:
Hi @xaviRodri :wave: 

> For some reason I cannot recognise yet, this casting is (in some cases) changing the incoming integer.

This is due to the way float works. They cannot represent every number, and sometimes an approximation is necessary. You can find the same behavior in Elixir itself:

```elixir
:erlang.float_to_binary(12091941575702789.0, decimals: 2)
#=> "12091941575702788.00"
```

You can play with this tool to see how the float is stored: https://www.h-schmidt.net/FloatConverter/IEEE754.html

native/explorer/src/encoding.rs
```diff
@@ -138,10 +138,11 @@ fn naive_datetime_struct_keys(env: Env) -> [NIF_TERM; 9] {
 }
 
 #[inline]
-fn time_unit_to_factor(time_unit: TimeUnit) -> i64 {
+fn time_unit_to_factor(time_unit: TimeUnit) -> f64 {
```
philss, line 141: @xaviRodri I would suggest use this function directly to convert the value to microseconds, and you only use  floats for the case of nanoseconds. With this we should only have small imprecision with nanoseconds.

something like:

```rust
fn time_to_microseconds(v: i64, time_unit: TimeUnit) -> i64 {
  match time_unit {
      TimeUnit::Milliseconds => v * 1000,
      TimeUnit::Microseconds => v,
      TimeUnit::Nanoseconds => (v as f64 * 0.001) as i64,
      _ => unreachable!(),
  }
}
```

WDYT?

xaviRodri: Hey!!
@philss thanks a lot for the suggestion! 👏 
Makes more sense so we minimise the imprecision as much as possible 👍 
I will try then to implement some `.parquet` tests with this and push again for another review!

test/explorer/data_frame/parquet_test.ex
```diff
@@ -0,0 +1,82 @@
...
+  alias Explorer.Series
+
+  # https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.EPSILON
+  @f64_epsilon 2.2204460492503131e-16
+
+  def tmp_file!(data) do
+    filename = System.tmp_dir!() |> Path.join("data.parquet")
+    :ok = DF.to_parquet(data, filename)
+    filename
+  end
```
philss, line 16: Just another way to write. But the main point is rename `data` by `df`. WDYT?

```suggestion
  def tmp_file!(df) do
    System.tmp_dir!() |> Path.join("data.parquet") |> tap(fn filename ->
      :ok = DF.to_parquet(df, filename)
    end)
  end
```

test/explorer/data_frame/parquet_test.ex
```diff
@@ -0,0 +1,82 @@
...
+
+  test "read" do
+    data = Explorer.Datasets.iris()
+
+    {:ok, frame} = DF.from_parquet(tmp_file!(data))
```
philss, line 21:
```suggestion
    parquet = tmp_file!(Explorer.Datasets.iris())

    {:ok, frame} = DF.from_parquet(parquet)
```

test/explorer/data_frame/parquet_test.ex
```diff
@@ -0,0 +1,82 @@
...
+    species = frame["species"]
+
+    assert species[0] == "Iris-setosa"
+    assert species[149] == "Iris-virginica"
+  end
+
+  def assert_parquet(type, value, parsed_value) do
+    data = [value] |> Series.from_list() |> Series.cast(type) |> then(&DF.new(column: &1))
+    # parsing should work as expected
+    {:ok, frame} = DF.from_parquet(tmp_file!(data))
```
philss, line 45:
```suggestion
    df = [value] |> Series.from_list() |> Series.cast(type) |> then(&DF.new(column: &1))
    # parsing should work as expected
    {:ok, frame} = DF.from_parquet(tmp_file!(df))
```

philss (commented):
Just a small detail about naming, but looks good to me! :+1:

philss:
@xaviRodri running `MIX_ENV=test mix ci` should fix the build. You could also jump into `native/explorer` and run `cargo fmt` :)

test/explorer/data_frame/parquet_test.ex
xaviRodri: 100% agree!
Also, always in favour of saving some variables 👍

test/explorer/data_frame/parquet_test.ex
xaviRodri: more clear again 💯

philss:
@xaviRodri It's because you are covering all values of the [`TimeUnit`](https://docs.rs/polars-core/latest/polars_core/datatypes/enum.TimeUnit.html) enum. So the solution is just to remove that line :)

philss (approved):
:shipit:

josevalim:
awesome job! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #366, opened 2022-10-09
Add missing "concat" operation for lazy series

nallwhy:
This PR adds `concat/2` operation for LazySeries.

### DIFF
--- a/native/explorer/src/expressions.rs
+++ b/native/explorer/src/expressions.rs
@@ -368,6 +368,14 @@ pub fn expr_last(expr: ExExpr) -> ExExpr {
 }
 
 #[rustler::nif]
+pub fn expr_concat(left: ExExpr, right: ExExpr) -> ExExpr {
+    let left_expr: Expr = left.resource.0.clone();
+    let right_expr: Expr = right.resource.0.clone();
+
+    ExExpr::new(left_expr.append(right_expr, false))
+}
+
+#[rustler::nif]
 pub fn expr_coalesce(left: ExExpr, right: ExExpr) -> ExExpr {
     let predicate: Expr = left.resource.0.clone().is_not_null();
     let left_expr: Expr = left.resource.0.clone();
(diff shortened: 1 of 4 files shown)


philss (approved):
Cool! Looks good to me! :ship:

josevalim:
Thank you @nallwhy! If you want to explore a PR where only one of them is a lazy series, that will also be welcome. :)

native/explorer/src/expressions.rs
```diff
@@ -367,6 +367,14 @@ pub fn expr_last(expr: ExExpr) -> ExExpr {
     ExExpr::new(expr.last())
 }
 
+#[rustler::nif]
+pub fn expr_concat(left: ExExpr, right: ExExpr) -> ExExpr {
+    let left_expr: Expr = left.resource.0.clone();
+    let right_expr: Expr = right.resource.0.clone();
+
+    ExExpr::new(left_expr.append(right_expr, false))
```
nallwhy, line 375: @philss Oh, I forgot to mention it.
the 2nd arg of append is [`append_chunks`](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.Series.append.html).
> 
> Prefer extend over append_chunks when you want to do a query after a single append. For instance during online operations where you add n rows and rerun a query.

> Prefer append_chunks over extend when you want to append many times before doing a query. For instance when you read in multiple files and when to store them in a single Series. In the latter case, finish the sequence of append_chunks operations with a rechunk.

I chose `extend` because I think it is a general way.

nallwhy:
@josevalim Is there any reference for that? I have no idea how to implement that.

josevalim:
I thought `add` was such an example, but maybe I am mistaken?

nallwhy:
> I thought `add` was such an example, but maybe I am mistaken?

`add/2` can take Series/LazySeries + number but not Series + LazySeries.
And the broadcasting happens in Elixir not Rust.

philss:
I think the problem is in the `PolarsBackend.Series` only. We should try to work with expressions when using literal values instead of creating new series. I'm going to open a new issue for that.

Issue #367, opened 2022-10-10
Allow operation between Series and LazySeries


nallwhy:
- [x] Series + LazySeries -> LazySeries
- [x] LazySeries + Series -> LazySeries

The code below should work.

```elixir
df = Explorer.DataFrame.new(%{a: [1, 2, 3]})

df
|> Explorer.DataFrame.mutate_with(&%{
  b: Explorer.Series.add(&1[:a], Explorer.Series.from_list([1, 2, 3])),
  c: Explorer.Series.add(Explorer.Series.from_list([1, 2, 3]), &1[:a])
})
```

Reference: https://github.com/pola-rs/polars/pull/4988

nallwhy:
Let me handle this issue

nallwhy:
Hmm.. maybe I misunderstood the reference PR.

For the arithmetic operations, polars do series + expr -> series. right?

I need to check series + expr -> expr, expr + series -> series works first.

philss:
@nallwhy I think you need to transform the series to a literal value expression first. Maybe https://docs.rs/polars/latest/polars/series/struct.Series.html#method.lit does the trick, but I'm not sure.

nallwhy:
@philss Thanks! It works :)

PR #376, opened 2022-10-16
Allow arithmetic ops on (series, lazy series)

nallwhy:
#367

This PR allows arithmetic ops on (series, lazy series).

I made `Explorer.Shared.apply_series_impl/2`.

It determines impl following rules below
- if `series_or_scalars` contain at least one lazy series -> lazy series
- else if `series_or_scalars` contain at least one series -> series
- else if `series_or_scalars` don't contain any series or lazy series -> raise

Remain tasks

- Apply it to comparison ops and slice and dice ops
- Replace `Explorer.Shared.apply_binary_op_impl/3` with it
- Merge it into `Explorer.Shared.apply_impl/2`

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -240,21 +240,10 @@ defmodule Explorer.Backend.LazySeries do
 
   for op <- @arithmetic_operations do
     @impl true
-    def unquote(op)(%Series{} = left, value_or_series) do
-      dtype = resolve_numeric_dtype([left, value_or_series])
-
-      value = with %Series{} <- value_or_series, do: series_or_lazy_series!(value_or_series)
-
-      args = [lazy_series!(left), value]
-      data = new(unquote(op), args, aggregations?(args), window_functions?(args))
-
-      Backend.Series.new(data, dtype)
-    end
-
-    def unquote(op)(left, %Series{} = right) when is_number(left) do
+    def unquote(op)(left, right) do
       dtype = resolve_numeric_dtype([left, right])
 
-      args = [left, lazy_series!(right)]
+      args = has_lazy_series!([left, right])
       data = new(unquote(op), args, aggregations?(args), window_functions?(args))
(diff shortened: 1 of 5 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -376,15 +365,36 @@ defmodule Explorer.Backend.LazySeries do
...
+
+  defp has_lazy_series!(items) do
+    has_lazy_series =
+      items
+      |> Enum.any?(fn
+        %Series{data: %__MODULE__{}} -> true
+        _ -> false
+      end)
+
+    case has_lazy_series do
```
josevalim, line 386: Do we need this check? If we got here, it is because at least one of them was lazy, no?

lib/explorer/shared.ex
```diff
@@ -35,6 +37,24 @@ defmodule Explorer.Shared do
...
+  def series_impl!(series_or_scalars) when is_list(series_or_scalars) do
+    impl =
+      Enum.reduce(series_or_scalars, nil, fn
+        %{data: %struct{}}, nil -> struct
+        %{data: %struct{}}, impl -> pick_series_impl(impl, struct)
+        _scalar, impl -> impl
+      end)
+
+    if impl == nil do
+      raise ArgumentError, "could not find a implementation for: #{inspect(series_or_scalars)}"
```
josevalim, line 52:
```suggestion
      raise ArgumentError,
            "expected at least one series to be given as argument, got: #{inspect(series_or_scalars)}"
```

lib/explorer/backend/lazy_series.ex
nallwhy: Right! It's a defensive code, but we can remove it.

josevalim: Yes, let's remove it then! :)

josevalim (approved):

josevalim:

Issue #379, opened 2022-10-21
Allow `Explorer.Series.pow/2` on (number, lazy series)


nallwhy:
Specs related to `Explorer.Series.pow/2` also should be changed.

Issue #374, opened 2022-10-16
Mismatch dtypes of `Explorer.Series.pow/2` result in data frame and series when mutate with lazy series


nallwhy:
```elixir
df = Explorer.DataFrame.new(%{a: [1, 2, 3]})

df =
  df
  |> Explorer.DataFrame.mutate_with(fn ldf ->
    %{
      b: Explorer.Series.pow(ldf[:a], ldf[:a]),
      c: Explorer.Series.divide(ldf[:a], ldf[:a])
    }
  end)

# dtype of "b" should be float
%{"b" => :integer, "c" => :float} = df |> Explorer.DataFrame.dtypes()

:float = df[:b] |> Explorer.Series.dtype()
:float = df[:c] |> Explorer.Series.dtype()
```

nallwhy:
It will happen after pow/2 between series is allowed.

philss:
@nallwhy I choose to keep it as float because it's common to have `pow/2` operations with integers that the result is a float. Please take a look: https://github.com/pola-rs/polars/issues/4796

I think we should standardize the eager implementation to always return float as well. WDYT?

cc/ @josevalim @cigrainger

josevalim:
Nx allows power of integers and the result is an integer. If a negative number is given as exponent, it raises.

nallwhy:
Is it possible to don't follow the result of polars?

philss:
> Is it possible to don't follow the result of polars?

I think only with casting. We would need to cast the resultant series.

josevalim:
Maybe `pow` in Polars has the same semantics as Python, which is that it returns integer for integers and float for floats. If that's the case, I would rather align Explorer to expect the same result as Polars/Python. Which is also how Nx and Elixir behaves.

josevalim:
I would even go ahead and suggest to rename to `power` to match `Nx`.

billylanchantin:
Yesterday I mentioned that passing an `out_dtype` to the backend may be the way to address this issue. After digging more into it I think I was mistaken.

After @philss's original issue the Polars team has since raised a follow-up issue:

  * https://github.com/pola-rs/polars/issues/9051

It seems they weren't satisfied with the unnecessary cast to float either. In that issue they describe the [plan](https://github.com/pola-rs/polars/issues/9051#issuecomment-1672841511) (paraphrased):

```
pow(uint,  uint)  = uint
pow(int,   uint)  = int
pow(int,   int)   = float
[...]

PR #382, opened 2022-10-22
Allow `Explorer.Series.pow/2` on (number, lazy series)

nallwhy:
This PR solves #379

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1429,8 +1429,8 @@ defmodule Explorer.Series do
       >
   """
   @doc type: :element_wise
-  @spec pow(series :: Series.t(), right :: Series.t() | number()) :: Series.t()
-  def pow(%Series{} = series, right), do: basic_numeric_operation(:pow, series, right)
+  @spec pow(left :: Series.t() | number(), right :: Series.t() | number()) :: Series.t()
+  def pow(left, right), do: basic_numeric_operation(:pow, left, right)
 
   @doc """
   Element-wise integer division.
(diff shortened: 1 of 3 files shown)


test/explorer/data_frame_test.exs
```diff
@@ -560,7 +560,7 @@ defmodule Explorer.DataFrameTest do
                "calc2" => :integer,
                "calc3" => :integer,
                "calc4" => :float,
-               #  "calc5" => :integer,
+               "calc5" => :integer,
```
josevalim, line 563: Is this correct? The results above are float, no?

nallwhy: It's a bug #374.

dtype of series is float, but dtype of dataframe column is integer.

josevalim: So maybe we should leave this one commented and add a test where the first argument is a float?

josevalim (approved):

lib/explorer/series.ex
```diff
@@ -1429,8 +1429,8 @@ defmodule Explorer.Series do
       >
   """
   @doc type: :element_wise
-  @spec pow(series :: Series.t(), right :: Series.t() | number()) :: Series.t()
-  def pow(%Series{} = series, right), do: basic_numeric_operation(:pow, series, right)
+  @spec pow(left :: Series.t() | number(), right :: Series.t() | number()) :: Series.t()
+  def pow(left, right), do: basic_numeric_operation(:pow, left, right)
```
philss, line 1433: I think we can merge this, but for what I understand the eager version of `Series.pow/2` is not going to work with numbers on the left-hand side.

philss (approved):

philss:
@nallwhy thank you! :purple_heart:

PR #384, opened 2022-10-24
Add docs about DF operations when using groups - part 2

philss:
This is related to https://github.com/elixir-nx/explorer/issues/245 and a continuation of https://github.com/elixir-nx/explorer/pull/349

It includes the following:

- `to_{csv, ipc, ndjson, parquet}` (IO operations)
- `arrange_with` (with a fix for stable grouping)
- `to_lazy`
- `to_columns`
- `to_series`
- `shape` / `n_columns` / `n_rows`

I should open more PRs this week related to this.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -439,6 +439,8 @@ defmodule Explorer.DataFrame do
   @doc """
   Writes a dataframe to a parquet file.
 
+  Groups are ignored if the dataframe is using any.
+
   ## Options
 
     * `compression` - The compression algorithm to use when writing files.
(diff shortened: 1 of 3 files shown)


lib/explorer/data_frame.ex
```diff
@@ -1744,6 +1769,30 @@ defmodule Explorer.DataFrame do
         a integer [3, 3, 1]
         b integer [2, 3, 1]
       >
+
+  ## Grouped examples
+
+  When used in a grouped dataframe, `arrange_with/2` is going to sort each group individually
+  and then return the entire dataframe with the existing groups. If one of the arrange columns
```
josevalim, line 1776:
```suggestion
  and then return the entire dataframe with the existing groups. If one of the arranged columns
```

Or "one of the columns to arrange".

lib/explorer/data_frame.ex
```diff
@@ -1744,6 +1769,30 @@ defmodule Explorer.DataFrame do
...
+  is also a group, the sorting for that column is not going to work. It is necessary to
```
josevalim, line 1777: What do you mean by "it is not going to work?". Does it raise? Does it return the wrong value?

josevalim, line 1777:
```suggestion
  and then return the entire dataframe with the existing groups. Therefore, if you attempt
  to arrange a grouped column, it won't have any effect and work as no-op. It is necessary to
```

philss:
@josevalim I'm merging this, ok?

josevalim (approved):

PR #388, opened 2022-10-30
Simplify target section of release Action

tmr08c:
Update the matrix in `release.yml` to have a list of `nifs` and `jobs`. This allows the targets to be re-used across different `nif` versions, avoiding the need to list the same targets for each `nif` version.

Suggesting this in the hopes that it makes maintaining the file a bit easier. 

This was inspired by:

- https://github.com/philss/rustler_precompilation_example/blob/6a2baafc1995448bdd70dcb9c799bf85d2368df3/.github/workflows/release.yml#L25-L35
- https://github.com/kloeckner-i/mail_parser/blob/fe20dc953ea10f86b1ff3ad8532866a733a406ea/.github/workflows/release.yml#L29-L40

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,55 +26,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        nif: ["2.16", "2.15", "2.14"]
         job:
-          # NIF version 2.16
           - {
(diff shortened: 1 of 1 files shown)


philss:
@tmr08c much better now! Thank you! :purple_heart:

PR #394, opened 2022-11-04
Add docs for DF operations when using groups - part III

philss:
This is another PR related to https://github.com/elixir-nx/explorer/issues/245

This covers:

* sample
* rename
* rename_with
* mutate

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1442,6 +1442,11 @@ defmodule Explorer.DataFrame do
   Columns are added with keyword list or maps. New variables overwrite existing variables of the
   same name. Column names are coerced from atoms to strings.
 
+  Notice that working with grouped dataframes is different from `mutate_with/2` because
+  it's not possible to work with aggregations per group in this version. Another drawback is that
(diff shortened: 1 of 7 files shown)


lib/explorer/data_frame.ex
```diff
@@ -2047,6 +2052,8 @@ defmodule Explorer.DataFrame do
   @doc """
   Renames columns.
 
+  Note that renaming a column that is also a group is going to rename
+  the group as well.
```
josevalim, line 2056:
```suggestion
  Renaming a column that is also a group is going to rename the group as well.
```

lib/explorer/data_frame.ex
```diff
@@ -2139,6 +2146,9 @@ defmodule Explorer.DataFrame do
   @doc """
   Renames columns with a function.
 
+  Note that renaming a column that is also a group is going to rename
+  the group as well.
```
josevalim, line 2150:
```suggestion
  Renaming a column that is also a group is going to rename the group as well.
```

josevalim (approved):

PR #395, opened 2022-11-04
Add support to groups for `DataFrame.slice/{2,3}`

philss:
This will enable slicing by each group inside the dataframe.

### DIFF
--- a/lib/explorer/polars_backend/data_frame.ex
+++ b/lib/explorer/polars_backend/data_frame.ex
@@ -437,13 +437,7 @@ defmodule Explorer.PolarsBackend.DataFrame do
     selected_indices =
       df
       |> indices_by_groups()
-      |> Enum.flat_map(fn group ->
-        group
-        |> Stream.with_index()
-        |> Stream.filter(fn {_row_idx, idx} -> idx in row_indices end)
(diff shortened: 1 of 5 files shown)


lib/explorer/polars_backend/data_frame.ex
```diff
@@ -429,10 +429,49 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
+  def slice(%DataFrame{} = df, row_indices) when is_list(row_indices) do
+    selected_indices =
+      df
+      |> indices_by_groups()
+      |> Enum.flat_map(fn group ->
+        group
+        |> Stream.with_index()
+        |> Stream.filter(fn {_row_idx, idx} -> idx in row_indices end)
+        |> Stream.map(fn {row_idx, _idx} -> row_idx end)
+        |> Enum.to_list()
```
josevalim, line 445: Using a custom recursive function will likely be much faster:

```
def filter_indexes([row_idx | indexes], idx, row_indices) do
  if idx in row_indices do
    [row_idx | filter_indexes(indexes, idx + 1, row_indices)]
  else
    filter_indexes(indexes, idx + 1, row_indices)
  end
end
```

josevalim (approved):
One comment for optimizing and ship it!

josevalim (approved):

PR #397, opened 2022-11-08
Fix pivot operations to work with groups

philss:
This also adds documentation for both operations when using groups.

The behaviour is inspired in what is described in the Tidyr article about pivoting: https://tidyr.tidyverse.org/articles/pivot.html 
There is no explicit mention about groups, but the examples shows how they behave.


josevalim (approved):
Beautiful!

PR #401, opened 2022-11-10
WIP: initial work to add support for list dtypes

philss:
This is work-in-progress and adds the bases of "list" dtypes.

It also adds the `to_list/1` lazy operation that takes a series and creates a list series with the elements of that series. This is useful for aggregations, when you want to capture the elements of a given group. Eg.:

```elixir
df = Explorer.DataFrame.new(a: [1, 1, 2, 2], b: [9, 8, 7, 6])

grouped = Explorer.DataFrame.group_by(df, :a)

Explorer.DataFrame.summarise_with(grouped, fn df -> [b_merged: Explorer.Series.to_list(df["b"])] end)
```

The result is going to be something like this:

      #Explorer.DataFrame<
        Polars[2 x 2]
        a integer [1, 2]
        b_merged list(integer) [[ 9, 8 ], [ 7, 6 ]]
      >

Related to:
- https://github.com/elixir-nx/explorer/issues/296
- https://github.com/elixir-nx/explorer/issues/400

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -27,7 +27,7 @@ defmodule Explorer.Series do
 
   @valid_dtypes Explorer.Shared.dtypes()
 
-  @type dtype :: :integer | :float | :boolean | :string | :date | :datetime
+  @type dtype :: :integer | :float | :boolean | :string | :date | :datetime | {:list, :integer}
   @type t :: %Series{data: Explorer.Backend.Series.t(), dtype: dtype()}
   @type lazy_t :: %Series{data: Explorer.Backend.LazySeries.t(), dtype: dtype()}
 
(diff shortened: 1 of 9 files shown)


philss:
I didn't think too much about the inspecting, but I would like suggestions :smiley: 
Also, I should add all the dtypes encodings and normalization soon.

lib/explorer/series.ex
```diff
@@ -27,7 +27,7 @@ defmodule Explorer.Series do
 
   @valid_dtypes Explorer.Shared.dtypes()
 
-  @type dtype :: :integer | :float | :boolean | :string | :date | :datetime
+  @type dtype :: :integer | :float | :boolean | :string | :date | :datetime | {:list, :integer}
```
josevalim, line 30: We probably want to make it recursive further on:

```suggestion
  @type dtype :: :integer | :float | :boolean | :string | :date | :datetime | {:list, dtype}
```

lambdaofgod:
Hey @philss, how much work do you think it would take to make it work in 0.7?
I'm motivated to do this, but I'm an Elixir newbie and I don't want to get anyone hopes up in case I'll get bogged down

philss:
@lambdaofgod I think the representation part of lists is the easiest part, which is what I did in this PR. The real problem is to support operations with lists. We are postponing this since last year because of the unknown complexity of maintaining the lists dtypes. But I think we are near to start looking into this again.

> how much work do you think it would take to make it work in 0.7?

Short answer is: we don't know yet :sweat_smile:

Issue #403, opened 2022-11-10
Rename var, std, and pow


josevalim:
Elixir generally avoids shortcut in names. So we should rename them to `variance`, `standard_deviation`, and `power`, which mirrors Nx.

PR #409, opened 2022-11-12
Rename pow to power

nallwhy:
#403


josevalim:
We decided to change Nx.power to Nx.pow instead to align with Elixir. Thank you!

Issue #411, opened 2022-11-15
Typespec of `Series.max/1` and `Series.min/1` are missing `nil` as a possible result type


linusdm:
As far as I can see the return value of `Series.max/1` and `Series.min/1` can be `nil` if the series passed to it is empty. When matching on the result, dialyzer warns about `nil` not being a possible outcome for these functions.

See here for the typespec: https://github.com/elixir-nx/explorer/blob/main/lib/explorer/series.ex#L850

Is it as easy as this sounds? Just add `nil` as a possible type? Or is there more to it (maybe other affected aggregate functions)? Or am I just using it wrong (I first filter the a `DataFrame`, so the result can be empty, and thus the min/max can be `nil`)?

philss:
Hi @linusdm 
Yeah, you are correct. It is missing `nil` as a possible type. We need to add there and in the callback specs here: https://github.com/elixir-nx/explorer/blob/8fcbb4ce876f02f855d3d8fda471673f37406cc3/lib/explorer/backend/series.ex#L47

I think this may be a problem for all the functions that may return a scalar value (number, string, or datetime). If you want to fix that, PRs are welcome :)

PR #412, opened 2022-11-16
Add  as possible return type of aggregate functions (#411)

linusdm:
Add `nil` as possible return type ofo aggregate functions:
- sum
- min
- max
- mean
- median
- variance
- standard_deviation
- quantile

Each one of these functions can return `nil` when invoked with an empty Series.


philss:
@linusdm thank you! :purple_heart:

PR #413, opened 2022-11-18
Implement `mutate/2` as a macro using `Explorer.Query`

philss:
This is part of https://github.com/elixir-nx/explorer/issues/402

This also refactors the filtering tests to use the macro, and includes more binary operators.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1331,7 +1331,7 @@ defmodule Explorer.DataFrame do
 
   > #### Notice {: .notice}
   >
-  > Given this function is a macro, you must require `Explorer.DataFrame` before using it.
+  > This is macro, therefore you must `require  Explorer.DataFrame` before using it.
 
   Besides element-wise series operations, you can also use window functions
   and aggregations inside comparisons. In such cases, grouped dataframes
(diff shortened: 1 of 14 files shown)


lib/explorer/data_frame.ex
```diff
@@ -1329,8 +1327,11 @@ defmodule Explorer.DataFrame do
   Picks rows based on `Explorer.Query`.
 
   The query is compiled and runs efficiently against the dataframe.
-  The query must return a boolean expression. Given this function is
-  a macro, you must require `Explorer.DataFrame` before using it.
+  The query must return a boolean expression.
+
+  > #### Notice {: .notice}
+  >
+  > Given this function is a macro, you must require `Explorer.DataFrame` before using it.
```
josevalim, line 1334:
```suggestion
  > This is macro, therefore you must `require  Explorer.DataFrame` before using it.
```

lib/explorer/data_frame.ex
```diff
@@ -1478,15 +1479,23 @@ defmodule Explorer.DataFrame do
...
-  Creates and modifies columns.
+  Creates or modifies columns based on `Explorer.Query`.
+
+  The query is compiled and runs efficiently against the dataframe.
+  New variables overwrite existing variables of the same name.
+  Column names are coerced from atoms to strings.
+
+  > #### Notice {: .notice}
+  >
+  > Given this function is a macro, you must require `Explorer.DataFrame` before using it.
```
josevalim, line 1490:
```suggestion
  > This is macro, therefore you must `require  Explorer.DataFrame` before using it.
```

josevalim: Another idea is to find all places where we do  `alias Explorer.DataFrame, as: DF` and replace it by `require Explorer.DataFrame, as: DF`.

josevalim (approved):
Beautiful!

lib/explorer/data_frame.ex
philss: I'm going to try that in the next PR. Thanks! :)

PR #417, opened 2022-11-18
Add `Explorer.DataFrame.put/3`

### LINKED ISSUE #414: Add Explorer.DataFrame.put/3
Just to answer the [question](https://github.com/elixir-nx/explorer/issues/355#issuecomment-1250243355), I think the second option looks better: 

```elixir
Explorer.DataFrame.put(df, :c, Explorer.Series.transform(df[:b], fn p -> p end))
```

With `put` you can add or replace a column.

philss:
Closes #414

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1500,10 +1500,7 @@ defmodule Explorer.DataFrame do
 
   ## Examples
 
-  Mutations are useful to add or modify variables in your dataframe.
-  For example, if we want to add "1" to the value of a column, we can
-  simply do a `add(column, 1)`, or even `column + 1`, thanks to our
-  macro:
+  Mutations are useful to add or modify columns in your dataframe:
 
       iex> df = Explorer.DataFrame.new(a: ["a", "b", "c"], b: [1, 2, 3])
       iex> Explorer.DataFrame.mutate(df, c: b + 1)
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1623,8 +1623,7 @@ defmodule Explorer.DataFrame do
         b float [1.0, 4.0, 9.0]
       >
 
-  Different from the macro `mutate/2`, it's possible to "reuse" a variable
-  for different computations:
+  It's possible to "reuse" a variable for different computations:
 
       iex> df = Explorer.DataFrame.new(a: [4, 5, 6], b: [1, 2, 3])
       iex> Explorer.DataFrame.mutate_with(df, fn ldf ->
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -353,8 +353,9 @@ defmodule Explorer.DataFrame do
     value = pull(df, column)
     {current_value, new_value} = fun.(value)
 
-    new_data = mutate_with(df, fn _ -> %{column => new_value} end)
-    {current_value, new_data}
+    new_df = put(df, column, new_value)
+
+    {current_value, new_df}
   end
 
   # IO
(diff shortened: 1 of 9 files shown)


lib/explorer/data_frame.ex
```diff
@@ -1701,6 +1701,75 @@ defmodule Explorer.DataFrame do
     Shared.apply_impl(df, :mutate_with, [df_out, column_pairs])
   end
 
+  @doc """
+  Creates or modifies a single column.
+
+  This is a simplified way to add or modify one column.
```
josevalim, line 1707: let's add a warning to prefer `mutate` whenever possible?

lib/explorer/data_frame.ex
```diff
@@ -1701,6 +1701,75 @@ defmodule Explorer.DataFrame do
...
+  @spec put(
+          DataFrame.t(),
+          column_name(),
+          Series.t()
+          | list(Explorer.Backend.Series.valid_types())
+          | Explorer.Backend.Series.valid_types()
+        ) ::
+          DataFrame.t()
+  def put(%DataFrame{} = df, column_name, series) do
+    mutate_with(df, fn _ -> [{column_name, series}] end)
```
josevalim, line 1770: Shouldn't mutate_with warn if the series is not a LazySeries? I think we would need a new callback for this.

josevalim: BTW, if we introduce a new operation, I would make it so `put` behaves the same regardless if grouped or not. That's because I think having all groups matching is rather unlikely?

philss: > Shouldn't mutate_with warn if the series is not a LazySeries? I think we would need a new callback for this.

Now we are implicitly converting a Polars Series to a LazySeries, because a lot of our examples were using series. This is the reason we are not warning.

> BTW, if we introduce a new operation, I would make it so put behaves the same regardless if grouped or not. That's because I think having all groups matching is rather unlikely?

Yeah, having all groups matching is unlikely.
I can add another callback and ignore groups for that. WDYT?

lib/explorer/data_frame.ex
```diff
@@ -1630,11 +1626,27 @@ defmodule Explorer.DataFrame do
         b float [1.0, 4.0, 9.0]
       >
 
+  Different from the macro `mutate/2`, it's possible to "reuse" a variable
+  for different computations:
```
josevalim, line 1630: I think you can do this as well in queries, it just a bit weird:

```
mutate(df, (
  c = a + b
  [c: c, d: window_sum(c, 2)]
))
```

lib/explorer/data_frame.ex
```diff
@@ -1499,36 +1500,27 @@ defmodule Explorer.DataFrame do
...
-        a string ["a", "b", "c"]
-        b integer [1, 2, 3]
-        c integer [4, 5, 6]
-      >
-
-  Or you can pass in a series:
+  Mutations are useful to add or modify variables in your dataframe.
+  For example, if we want to add "1" to the value of a column, we can
+  simply do a `add(column, 1)`, or even `column + 1`, thanks to our
+  macro:
```
josevalim, line 1506:
```suggestion
  Mutations are useful to add or modify columns in your dataframe:
```

josevalim (approved):

lib/explorer/data_frame.ex
philss: @josevalim I think this is not possible today. It is giving me this error:

```
== Compilation error in file test/explorer/data_frame_test.exs ==
** (CompileError) (for doctest at) lib/explorer/data_frame.ex:1559: cannot invoke remote function Explorer.DataFrame.pull/2 inside a match
    (elixir 1.14.0) src/elixir_fn.erl:17: anonymous fn/4 in :elixir_fn.expand/4
    (stdlib 4.1) lists.erl:1462: :lists.mapfoldl_1/3
    (elixir 1.14.0) src/elixir_fn.erl:23: :elixir_fn.expand/4
    (explorer 0.4.0-dev) expanding macro: Explorer.Query.query/1
    (for doctest at) lib/explorer/data_frame.ex:1558: Explorer.DataFrameTest."doctest Explorer.DataFrame.mutate/2 (135)"/1
    (explorer 0.4.0-dev) expanding macro: Explorer.DataFrame.mutate/2
    (for doctest at) lib/explorer/data_frame.ex:1558: Explorer.DataFrameTest."doctest Explorer.DataFrame.mutate/2 (135)"/1
```

But if I do with `^c`, it works:

```elixir
mutate(df, (
  ^c = a + b
  [c: ^c, d: window_sum(^c, 2)]
))
```

I think the only strange line is the first one. Assign with `^c =` is kind of weird, but the rest makes sense to me.
WDYT?

philss: Since it's not a problem with `put/3`, I'm merging this. We can tackle this in another PR.

PR #418, opened 2022-11-18
Add `Series.select/3` operation

thmsmlr:
Solves https://github.com/elixir-nx/explorer/issues/356

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -39,6 +39,7 @@ defmodule Explorer.Backend.LazySeries do
     concat: 2,
     coalesce: 2,
     cast: 2,
+    select: 3,
     # Window functions
     cumulative_max: 2,
     cumulative_min: 2,
@@ -345,6 +346,26 @@ defmodule Explorer.Backend.LazySeries do
         resolve_numeric_dtype([left, right])
       else
         left.dtype
+      end
+
+    Backend.Series.new(data, dtype)
+  end
+
+  @impl true
+  def select(%Series{} = pred, %Series{} = on_true, %Series{} = on_false) do
+    args = [
+      series_or_lazy_series!(pred),
+      series_or_lazy_series!(on_true),
+      series_or_lazy_series!(on_false)
+    ]
+
+    data = new(:select, args, aggregations?(args), window_functions?(args))
+
+    dtype =
+      if on_true.dtype != on_false.dtype do
+        raise "on_true dtype #{on_true.dtype} must equal on_false dtype #{on_false.dtype}"
+      else
+        on_true.dtype
       end
 
     Backend.Series.new(data, dtype)
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -352,9 +352,9 @@ defmodule Explorer.Backend.LazySeries do
   end
 
   @impl true
-  def select(%Series{} = pred, %Series{} = on_true, %Series{} = on_false) do
+  def select(%Series{} = predicate, %Series{} = on_true, %Series{} = on_false) do
     args = [
-      series_or_lazy_series!(pred),
+      series_or_lazy_series!(predicate),
       series_or_lazy_series!(on_true),
       series_or_lazy_series!(on_false)
     ]
@@ -362,10 +362,15 @@ defmodule Explorer.Backend.LazySeries do
     data = new(:select, args, aggregations?(args), window_functions?(args))
 
     dtype =
-      if on_true.dtype != on_false.dtype do
-        raise "on_true dtype #{on_true.dtype} must equal on_false dtype #{on_false.dtype}"
-      else
-        on_true.dtype
+      cond do
+        on_true.dtype in [:float, :integer] ->
+          resolve_numeric_dtype([on_true, on_false])
+
+        on_true.dtype != on_false.dtype ->
+          raise "on_true dtype #{on_true.dtype} must equal on_false dtype #{on_false.dtype}"
+
+        true ->
+          on_true.dtype
       end
 
     Backend.Series.new(data, dtype)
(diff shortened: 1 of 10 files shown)


philss:
@thmsmlr thanks for the PR! :purple_heart: 

Please run `mix rust.fmt` (or the cargo command if you are inside the native/explorer path).

lib/explorer/backend/lazy_series.ex
```diff
@@ -350,6 +351,26 @@ defmodule Explorer.Backend.LazySeries do
...
+    args = [
+      series_or_lazy_series!(pred),
+      series_or_lazy_series!(on_true),
+      series_or_lazy_series!(on_false)
+    ]
+
+    data = new(:select, args, aggregations?(args), window_functions?(args))
+
+    dtype =
+      if on_true.dtype != on_false.dtype do
```
philss, line 365: I think they can be different if they are numeric. So if the `on_true` is integer and `on_false` is float, the final dtype must be `float`.

lib/explorer/series.ex
```diff
@@ -476,6 +476,20 @@ defmodule Explorer.Series do
   @spec last(series :: Series.t()) :: any()
   def last(series), do: Shared.apply_impl(series, :last, [])
 
+  @doc """
+  return a series from two series, based on a predicate.
```
philss, line 480:
```suggestion
  Returns a series from two series, based on a predicate.
```

lib/explorer/series.ex
```diff
@@ -476,6 +476,20 @@ defmodule Explorer.Series do
...
+
+  The resulting series is built by evaluating each element of
+  `pred` and returning either the corresponding element from
```
philss, line 483: I think it's better if we use the full name: `predicate`. Same as the arguments for the functions.
It's common in our codebase to not use abbreviations.

test/explorer/data_frame_test.exs
```diff
@@ -685,6 +685,27 @@ defmodule Explorer.DataFrameTest do
...
+          ]
+        end)
+
+      assert DF.to_columns(df1, atom_keys: true) == %{
+               a: [true, false, true],
+               b: [3, 4, 2],
+               c: [6, 2, 1],
+               select1: [3, 2, 2]
+             }
+    end
```
philss, line 707: Cool! Would you mind to add a test for the `Explorer.Series.select/3` function as well? The idea is to also have a test for the Series module.

philss (commented):
Awesome! :D 

I added some minor points :)

lib/explorer/backend/lazy_series.ex
thmsmlr: good point, i'll leverage some of the helper functions to get this right

lib/explorer/series.ex
```diff
@@ -476,6 +476,41 @@ defmodule Explorer.Series do
   @spec last(series :: Series.t()) :: any()
   def last(series), do: Shared.apply_impl(series, :last, [])
 
+  @doc """
+  Returns a series from two series, based on a predicate.
+
+  The resulting series is built by evaluating each element of
+  `pred` and returning either the corresponding element from
```
philss, line 483:
```suggestion
  `predicate` and returning either the corresponding element from
```

lib/explorer/series.ex
```diff
@@ -476,6 +476,41 @@ defmodule Explorer.Series do
...
+  `on_true` or `on_false`.
+
+  `pred` must be a boolean series. `on_true` and `on_false` must be
```
philss, line 486:
```suggestion
  `predicate` must be a boolean series. `on_true` and `on_false` must be
```

lib/explorer/series.ex
```diff
@@ -476,6 +476,41 @@ defmodule Explorer.Series do
...
+  a series of the same length as `pred`.
+  """
+  @doc type: :transformation
+  @spec select(predicate :: Series.t(), on_true :: Series.t(), on_false :: Series.t()) :: any()
```
philss, line 490: I believe we always return a `Series` here.

```suggestion
  @spec select(predicate :: Series.t(), on_true :: Series.t(), on_false :: Series.t()) :: Series.t()
```

philss (approved):
Just a few minor points and :shipit: 

I will let this open for José or Cris to review as well :)

lib/explorer/series.ex
```diff
@@ -476,6 +476,41 @@ defmodule Explorer.Series do
...
+      when K.and(numeric_dtype?(on_true_dtype), numeric_dtype?(on_false_dtype)),
+      do: Shared.apply_impl(predicate, :select, [on_true, on_false])
+
+  def select(
+        predicate,
+        %Series{dtype: on_true_dtype} = on_true,
+        %Series{dtype: on_false_dtype} = on_false
+      )
+      when on_true_dtype == on_false_dtype,
+      do: Shared.apply_impl(predicate, :select, [on_true, on_false])
```
josevalim, line 505: We should check on both cases that `predicate` is a boolean. :)

lib/explorer/backend/lazy_series.ex
```diff
@@ -350,6 +351,31 @@ defmodule Explorer.Backend.LazySeries do
...
+
+    data = new(:select, args, aggregations?(args), window_functions?(args))
+
+    dtype =
+      cond do
+        on_true.dtype in [:float, :integer] ->
+          resolve_numeric_dtype([on_true, on_false])
+
+        on_true.dtype != on_false.dtype ->
+          raise "on_true dtype #{on_true.dtype} must equal on_false dtype #{on_false.dtype}"
```
josevalim, line 370: We don't need to check this here because Series.select already checks they both match or not. So we only need the first and third clause (it can be a regular if/else). :)

josevalim:
Awesome work @thmsmlr! I have dropped only some minor comments related to checking the predicate!

lib/explorer/series.ex
thmsmlr: Great point, will do

lib/explorer/backend/lazy_series.ex
thmsmlr: Ah ya makes sense. tbh i'm new to Elixir and to Rust, haven't fully grokked the codebase. I was mostly just pattern matching my way through this PR. Totally makes sense now that you mention it.

thmsmlr:
Alright I think made all the changes, ready for another round of review :)

josevalim:

PR #420, opened 2022-11-19
Adds `Series.in/2` function

### LINKED ISSUE #273: Support filtering on a list of values
See: https://elixirforum.com/t/explorer-subsetting-rows-by-columns-that-match-one-of-multiple-values/48564?u=cigrainger

This is a very common use case so we should prioritise it.

kimjoaoun:
This PR closes #273.
It creates the function `Explorer.Series.in/2` that uses [Polars' IsIn trait](https://pola-rs.github.io/polars/polars/chunked_array/object/trait.IsIn.html#method.is_in) to check if values from a list are contained in a Series.

I really don't know how to describe the function in the docstring, can someone help me?


native/explorer/src/series.rs
```diff
@@ -749,3 +749,24 @@ pub fn s_coalesce(data: ExSeries, other: ExSeries) -> Result<ExSeries, ExplorerE
...
+        DataType::Int8
+        | DataType::UInt8
+        | DataType::Int16
+        | DataType::UInt16
+        | DataType::Int32
+        | DataType::UInt32
+        | DataType::Int64 => s.i64().unwrap().is_in(rhs)?,
+        DataType::Float32 | DataType::Float64 => s.f64().unwrap().is_in(rhs)?,
+        DataType::Utf8 => s.utf8().unwrap().is_in(rhs)?,
+        dt => panic!("is_in/2 not implemented for {:?}", dt),
```
josevalim, line 768: Do we need to unwrap? Could we do `s.is_in(rhs)` directly?

lib/explorer/series.ex
```diff
@@ -1819,6 +1819,22 @@ defmodule Explorer.Series do
...
+
+      iex> s1 = Explorer.Series.from_list([1, 2, 3])
+      iex> Explorer.Series.is_in(s1, [1, 2])
+      #Explorer.Series<
+        Polars[3]
+        boolean [true, true, false]
+      >
+  """
+  def is_in(left, right) do
+    Shared.apply_series_impl(:is_in, [left, right])
```
josevalim, line 1835: I think you can copy the implementation from `equal`, because it is the same type of comparison?

native/explorer/src/series.rs
josevalim: Maybe `equal` could be used as reference here too.

kimjoaoun: I don'tit is possible to do directly, the compiler type-checker returns an error and suggests using `s.utf8()?.is_in(rhs)?` . That happens because `i64/f64/uft8` returns a Result. 
I can change to the `?` operator, though.

josevalim:
> I really don't know how to describe the function in the docstring, can someone help me?

Yeah, this is a good question. To be honest, I am not the biggest fan of the name. What does dplyr call it? It feels like `within?`, `included?`, etc would be better but I am not sure.

native/explorer/src/series.rs
josevalim: I see. I think we should include date/datetime comparisons then. Or is that not possible?

kimjoaoun: I don't see any reason why it wouldn't be. Let me try.

kimjoaoun:
> > I really don't know how to describe the function in the docstring, can someone help me?
> 
> Yeah, this is a good question. To be honest, I am not the biggest fan of the name. What does dplyr call it? It feels like `within?`, `included?`, etc would be better but I am not sure.

In R we don't have this operation in `dplyr`, because it is natively supported by the `%in%` operator.  But also in _base R_ there's a rarely used function called `within` that does exactly the same.
Between `within?` and `included?`, I think the former is the better.

Note: Pandas has [pd.Series.isin](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.isin.html).

lib/explorer/series.ex
kimjoaoun: > I think you can copy the implementation from `equal`, because it is the same type of comparison, except this one occurs with all elements?

Can I? When I try using the same implementation from  `equal`,  the comparison with lists stop working, see below:

```Elixir
# Keeps working
      iex> s1 = Explorer.Series.from_list([1, 2, 3])
      iex> Explorer.Series.is_in(s1, [1, 2])
      #Explorer.Series<
        Polars[3]
        boolean [true, true, false]
      >

# Stops working
      iex> s1 = Explorer.Series.from_list([1, 2, 3])
      iex> Explorer.Series.is_in(s1, [1, 2])
      #Explorer.Series<
        Polars[3]
        boolean [true, true, false]
      >
```

I guess I would need to make changes on `valid_for_bool_mask_operation?/2`, right?

kimjoaoun: > I think you can copy the implementation from `equal`, because it is the same type of comparison, except this one occurs with all elements?

Can I? When I try using the same implementation from  `equal`,  the comparison with lists stop working, see below:

```Elixir
# Keeps working
      iex> s3 = Explorer.Series.from_list([1, 2, 3])
      iex> s4 = Explorer.Series.from_list([1, 2, 4])
      iex> Explorer.Series.is_in(s3, s4)
      #Explorer.Series<
        Polars[3]
        boolean [true, true, false]
      >

# Stops working
      iex> s1 = Explorer.Series.from_list([1, 2, 3])
      iex> Explorer.Series.is_in(s1, [1, 2])
      #Explorer.Series<
        Polars[3]
        boolean [true, true, false]
      >
```

I guess I would need to make changes on `valid_for_bool_mask_operation?/2`, right?

lib/explorer/backend/series.ex
```diff
@@ -82,6 +82,7 @@ defmodule Explorer.Backend.Series do
   @callback less(s | valid_types(), s | valid_types()) :: s
   @callback less_equal(s | valid_types(), s | valid_types()) :: s
   @callback all_equal(s, s) :: boolean() | lazy_s()
+  @callback is_in(s, list() | s) :: s
```
josevalim, line 85: We convert to series in Explorer.Series, so the callback always receives series.

```suggestion
  @callback is_in(s, s) :: s
```

josevalim:
I see. So I think we should simply call it `in/2`. We can document it as "Check if the elements of the series in the left exists in the series in the right."

josevalim:
Note that, if we call it `in`, then you will have to write it as:

```
def %Series{} = left in %Series{} = right do
```

and

```
def %Series{} = left in right when is_list(right) do
```

The callback will be `@callback (s() in s()) :: s()` and so on.

lib/explorer/series.ex
```diff
@@ -1819,6 +1819,40 @@ defmodule Explorer.Series do
...
+      iex> s4 = Explorer.Series.from_list([1, 2, 4])
+      iex> Explorer.Series.is_in(s3, s4)
+      #Explorer.Series<
+        Polars[3]
+        boolean [true, true, false]
+      >
+  """
+  @doc type: :element_wise
+  def is_in(%Series{} = left, %Series{} = right),
+    do: Shared.apply_series_impl(:is_in, [left, right])
```
josevalim, line 1851: Regardless of the rename, I believe we want this:

```suggestion
  def is_in(%Series{} = left, %Series{} = right) do
    if valid_for_bool_mask_operation?(left, right) do
      Shared.apply_series_impl(:is_in, [left, right])
    else
      dtype_mismatch_error("is_in/2", left, right)
    end
    do: Shared.apply_series_impl(:is_in, [left, right])
```

kimjoaoun:
> I see. So I think we should simply call it `in/2`. We can document it as "Check if the elements of the series in the left exists in the series in the right."

Would not it be too similar to `Kernel.in`?

josevalim:
I think that's the idea! Then it will automatically work in Explorer.Query as: `series in [1, 2, 3]` and `left in right`, similar to R.

lib/explorer/polars_backend/native.ex
```diff
@@ -184,6 +184,7 @@ defmodule Explorer.PolarsBackend.Native do
   def s_window_mean(_s, _window_size, _weight, _ignore_null, _min_periods), do: err()
   def s_window_min(_s, _window_size, _weight, _ignore_null, _min_periods), do: err()
   def s_window_sum(_s, _window_size, _weight, _ignore_null, _min_periods), do: err()
+  def s_is_in(_s, _other), do: err()
```
josevalim, line 187:
```suggestion
  def s_in(_s, _other), do: err()
```

lib/explorer/polars_backend/series.ex
```diff
@@ -286,6 +286,10 @@ defmodule Explorer.PolarsBackend.Series do
   def all_equal(%Series{} = left, %Series{} = right),
     do: Shared.apply_series(left, :s_series_equal, [right.data, true])
 
+  @impl true
+  def is_in(%Series{} = left, right),
```
josevalim, line 290:
```suggestion
  def %Series{} = left in right,
```

native/explorer/src/series.rs
```diff
@@ -760,6 +760,28 @@ pub fn s_coalesce(data: ExSeries, other: ExSeries) -> Result<ExSeries, ExplorerE
...
+        DataType::Int8
+        | DataType::UInt8
+        | DataType::Int16
+        | DataType::UInt16
+        | DataType::Int32
+        | DataType::UInt32
+        | DataType::Int64 => s.i64()?.is_in(rhs)?,
+        DataType::Float32 | DataType::Float64 => s.f64()?.is_in(rhs)?,
+        DataType::Utf8 => s.utf8()?.is_in(rhs)?,
+        DataType::Datetime(_, _) => s.datetime()?.is_in(rhs)?,
```
josevalim, line 778: You are missing DataType::Date.

native/explorer/src/series.rs
```diff
@@ -760,6 +760,28 @@ pub fn s_coalesce(data: ExSeries, other: ExSeries) -> Result<ExSeries, ExplorerE
...
+
+    let s = match s.dtype() {
+        DataType::Int8
+        | DataType::UInt8
+        | DataType::Int16
+        | DataType::UInt16
+        | DataType::Int32
+        | DataType::UInt32
+        | DataType::Int64 => s.i64()?.is_in(rhs)?,
+        DataType::Float32 | DataType::Float64 => s.f64()?.is_in(rhs)?,
```
josevalim, line 776: We only effectively use these types:

```suggestion
        DataType::UInt32 | DataType::Int64 => s.i64()?.is_in(rhs)?,
        DataType::Float64 => s.f64()?.is_in(rhs)?,
```

cigrainger:
Given that this is failing for 1.13... should we go ahead and bump Explorer to minimum 1.14 for the next release? I'm surprised we're not already doing that.

josevalim:
@cigrainger fixing for v1.13 should be easy and mostly a matter of using `K.in(left, right)` instead of `in(left, right)`. I believe the biggest blocker for this PR is if you actually agree with the API. :)

cigrainger:
Hm okay. Jumping into this... it's a vectorised op right? So what we get back is a boolean series where each element is true if in the right and false if in the left? I like `in/2`. I think the confusion is coming from the fact that it's vectorised. But yeah... let's go with `in/2`

In terms of the docstring:

```elixir
@doc """
Checks if each element of the series in the left exists in the series in the right, returning a boolean mask.

  ## Examples

      iex> left = Explorer.Series.from_list([1, 2, 3])
      iex> right = Explorer.Series.from_list([1, 2])
      iex> Series.in(left, right)
      #Explorer.Series<
        Polars[3]
        boolean [true, true, false]
      >
"""
```

josevalim:

PR #424, opened 2022-11-23
Fix behaviour of Series' `sort/2` and `argsort/2`

philss:
This adds an option for selecting if `nils` are going to be in the beginning or the end of a series.

There is a small bug for `argsort` descending, which is not working if we select "nils first".

Closes https://github.com/elixir-nx/explorer/issues/404

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1962,7 +1962,11 @@ defmodule Explorer.Series do
 
   ## Options
 
-    * `:direction` - `:asc` or `:desc`
+    * `:direction` - `:asc` or `:desc`, meaning "ascending" or "descending", respectively.
+      By default it sorts in acending order.
+
+    * `:nils_last` - `true` or `false`. By default it is `true` if direction is `:asc`, and
+      false otherwise.
 
   ## Examples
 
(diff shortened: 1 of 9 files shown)


test/explorer/series_test.exs
```diff
@@ -651,4 +651,73 @@ defmodule Explorer.SeriesTest do
...
+  describe "argsort/2" do
+    test "indices of sorting a series in ascending order" do
+      s1 = Series.from_list([3, 1, nil, 2])
+
+      result = Series.argsort(s1)
+
+      assert Series.to_list(result) == [1, 3, 0, 2]
+    end
+
+    # There is a bug which is not considering "nils first" for descending argsort
```
josevalim, line 698:
```suggestion
    # There is a bug which is not considering "nils first" for descending argsort
    @tag :skip
```

and then uncomment the test? :)

lib/explorer/series.ex
```diff
@@ -1962,7 +1962,11 @@ defmodule Explorer.Series do
 
   ## Options
 
-    * `:direction` - `:asc` or `:desc`
+    * `:direction` - `:asc` or `:desc`, meaning "ascending" or "descending", respectively.
+      By default it sorts in acending order.
+
+    * `:nils_last` - `true` or `false`. By default it is `true` if direction is `:asc`, and
+      false otherwise.
```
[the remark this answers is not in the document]
philss: I think I'm going to change the option, but keep the internals using `nils_last`, because it's easier to work in the backend (with bools). WDYT?

lib/explorer/series.ex
```diff
@@ -1973,24 +1977,70 @@ defmodule Explorer.Series do
...
-    Shared.apply_impl(series, :sort, [opts[:direction] == :desc])
+    opts =
+      Keyword.validate!(opts,
+        direction: :asc,
+        nils:
+          if(K.or(K.is_nil(opts[:direction]), opts[:direction] == :asc), do: :last, else: :first)
+      )
+
+    descending? = opts[:direction] == :desc
+    nils_last? = opts[:nils] == :last
```
josevalim, line 1998:
```suggestion
    opts = Keyword.validate!(opts, [:nils, direction: :asc])
    descending? = opts[:direction] == :desc
    nils_last? = if nils = opts[:nils], do: nils == :last, else: not descending?
```

lib/explorer/series.ex
```diff
@@ -1973,24 +1977,70 @@ defmodule Explorer.Series do
...
-    Shared.apply_impl(series, :argsort, [opts[:direction] == :desc])
+    opts =
+      Keyword.validate!(opts,
+        direction: :asc,
+        nils:
+          if(K.or(K.is_nil(opts[:direction]), opts[:direction] == :asc), do: :last, else: :first)
+      )
+
+    descending? = opts[:direction] == :desc
+    nils_last? = opts[:nils] == :last
```
josevalim, line 2041: Same suggestion here.

josevalim (approved):
Two last comments and ship it!

PR #425, opened 2022-11-24
Update "Exploring Explorer" notebook to v0.4

philss:
This includes some of the new APIs and improvements to the text.

Closes https://github.com/elixir-nx/explorer/issues/421

### DIFF
--- a/notebooks/exploring_explorer.livemd
+++ b/notebooks/exploring_explorer.livemd
@@ -344,30 +344,35 @@ In `Explorer`, like in `dplyr`, we have five main verbs to work with dataframes:
 
 We are going to explore then in this notebook, but first we need to "require"
 the `Explorer.DataFrame` module in order to load the macros needed for these verbs.
-Since we have an alias for that module, we just need to:
+
+I want to take the opportunity to create a shorter alias for the `DataFrame` module,
+called `DF`:
 
 ```elixir
-require DataFrame
+require DataFrame, as: DF
 ```
+
+From now on we are using the shorter version, `DF`, to refer to the required `Explorer.DataFrame`
+module.
 
 ### Select
 
 Let's jump right into it. We can select columns pretty simply.
 
 ```elixir
-DataFrame.select(df, ["year", "country"])
+DF.select(df, ["year", "country"])
 ```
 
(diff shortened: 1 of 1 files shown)


notebooks/exploring_explorer.livemd
```diff
@@ -336,6 +332,24 @@ Or the tail. Let's get a few more values from the tail.
...
+* mutate
+* arrange
+* summarise
+
+We are going to explore then in this notebook, but first we need to "require"
+the `Explorer.DataFrame` module in order to load the macros needed for these verbs.
+Since we have an alias for that module, we just need to:
+
+```elixir
+require DataFrame
```
josevalim, line 350: Let's change `alias Explorer.DataFrame` at the top to `require Explorer.DataFrame, as: DF`. WDYT?

notebooks/exploring_explorer.livemd
```diff
@@ -336,6 +332,24 @@ Or the tail. Let's get a few more values from the tail.
...
+
+* select
+* filter
+* mutate
+* arrange
+* summarise
+
+We are going to explore then in this notebook, but first we need to "require"
+the `Explorer.DataFrame` module in order to load the macros needed for these verbs.
+Since we have an alias for that module, we just need to:
```
josevalim, line 347:
```suggestion
Since we have an alias for that module, we need to:
```

notebooks/exploring_explorer.livemd
josevalim: However, if we do so, should we alias `Explorer.Series` to `S`? Or maybe we start with `alias Explorer.Series` only and introduce the `require Explorer.DataFrame, as: DF` later?

notebooks/exploring_explorer.livemd
```diff
@@ -594,48 +623,58 @@ DataFrame.ungroup(grouped)
...
-more flexible `Explorer.DataFrame.summarise_with/2` function.
+In case you need more complex operations, like calculations inside groups, you can use multiple lines, like this:
 
 ```elixir
-DataFrame.summarise_with(grouped, fn df ->
-  max = Series.max(df["per_capita"])
+DataFrame.summarise(
+  grouped,
+  (
+    ^max = max(per_capita)
```
josevalim, line 663: Hrm, this is very confusing :( We should probably not allow `=` inside queries for now.

I would rewrite this to:

summarize(per_capita_max: max(per_capita), greater_than_9: max(per_capita) > 9.0)

notebooks/exploring_explorer.livemd
philss: I changed to introduce the `DF` alias here (the second option). WDYT?

PR #427, opened 2022-11-24
Add to_iovec

josevalim:


native/explorer/src/encoding.rs
```diff
@@ -324,18 +348,44 @@ pub fn list_from_series(data: ExSeries, env: Env) -> Term {
...
+    match s.dtype() {
+        DataType::Boolean => {
+            let mut bin = OwnedBinary::new(s.len()).unwrap();
+            let slice = bin.as_mut_slice();
+            for (i, v) in s.bool().unwrap().into_iter().enumerate() {
+                slice[i] = v.unwrap() as u8;
+            }
+            [bin.release(env)].encode(env)
+        }
+        DataType::Int32 => series_to_iovec!(resource, s, env, i32, i32),
```
philss, line 381: Would be possible to avoid passing both `resource` and `s`? I don't have a strong opinion about this, but since you can refer to the resource from the series, maybe you can simplify this. WDYT?

philss (approved):
Really cool! :shipit:

native/explorer/src/encoding.rs
josevalim: I tried but then I need to pass the data as a reference or similar and I got stuck :D

philss: Let's go with this and I will try later to remove the duplication (it's easier with the LSP :P).

josevalim:

PR #429, opened 2022-11-25
Refactor `mutate_with/2` and `slice/{2,3}` when using groups

philss:
This is a small refactor to the `PolarsBackend` implementation of DataFrame.

The ideia is to stop retrieving the list of indices of groups to Elixir, and do most of the job at the NIF side.

The only exception is for the `slice/2` with groups and ranges that cannot be made at the NIF side, because Elixir ranges are slightly more complex than the Rust counterpart.


lib/explorer/polars_backend/data_frame.ex
```diff
@@ -478,52 +438,32 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
 
-  defp filter_indices([row_idx | indices], idx, row_indices) do
-    if idx in row_indices do
-      [row_idx | filter_indices(indices, idx + 1, row_indices)]
-    else
-      filter_indices(indices, idx + 1, row_indices)
-    end
+  # Returns a list of lists, where each list is a group of row indices.
+  # TODO: Optimize indices by group so we don't convert it to a list.
+  # Instead, we pass a series to the following operations.
```
josevalim, line 462:
```suggestion
```

josevalim (approved):

PR #433, opened 2022-11-26
Add `Series.shift/2`

sasikumar87:


lib/explorer/series.ex
```diff
@@ -624,6 +624,30 @@ defmodule Explorer.Series do
   @spec last(series :: Series.t()) :: any()
   def last(series), do: Shared.apply_impl(series, :last, [])
 
+  @doc """
+  Returns the series, shifted by given offset.
+  Positive offset shfts from first, negative offset shifts from last.
```
josevalim, line 629:
```suggestion
  Shifts `series` by `offset` with `nil` values.

  Positive offset shifts from first, negative offset shifts from last.
```

josevalim:
Thank you! Apparently data.tables in R support a `shift` function with the same semantics, although dplyr does not and it supports `lead/lag` instead, but I think this one is clearer, so :+1:.

josevalim:

PR #435, opened 2022-11-26
Rename fetch! and take_every to at and at_every

josevalim:
`at` is more likely associated as an index based
operation in Elixir. This will give us more leeway
to add `at` and `at_every` operations to DataFrame
in the future, where `fetch!` could be read as a
"column-based" operation.


cigrainger (approved):
Totally happy with that change. I like it a lot actually!

josevalim:

PR #439, opened 2022-11-28
Add the `:binary` dtype for Series

philss:
This enables the creation of dataframes with binary columns. Useful to work with blob/raw data.

Closes https://github.com/elixir-nx/explorer/issues/182

### DIFF
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -162,15 +162,7 @@ defmodule Explorer.Shared do
   Helper for shared behaviour in inspect.
   """
   def to_string(i, _opts) when is_nil(i), do: "nil"
-
-  def to_string(i, _opts) when is_binary(i) do
-    if String.valid?(i) do
-      "\"#{i}\""
-    else
-      bin_as_list = :binary.bin_to_list(i)
-      "<<#{Enum.join(bin_as_list, ", ")}>>"
-    end
-  end
+  def to_string(i, _opts) when is_binary(i), do: inspect(i)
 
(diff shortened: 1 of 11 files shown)


lib/explorer/shared.ex
```diff
@@ -155,7 +162,16 @@ defmodule Explorer.Shared do
...
-  def to_string(i, _opts) when is_binary(i), do: "\"#{i}\""
+
+  def to_string(i, _opts) when is_binary(i) do
+    if String.valid?(i) do
+      "\"#{i}\""
+    else
+      bin_as_list = :binary.bin_to_list(i)
+      "<<#{Enum.join(bin_as_list, ", ")}>>"
+    end
+  end
```
josevalim, line 173:
```suggestion
  def to_string(i, _opts) when is_binary(i), do: inspect(i)
```

lib/explorer/shared.ex
```diff
@@ -130,7 +133,11 @@ defmodule Explorer.Shared do
   defp type(item, _type) when is_integer(item), do: :integer
   defp type(item, _type) when is_float(item), do: :float
   defp type(item, _type) when is_boolean(item), do: :boolean
-  defp type(item, _type) when is_binary(item), do: :string
+  defp type(item, :binary) when is_binary(item), do: :binary
+
+  defp type(item, _type) when is_binary(item),
+    do: if(String.valid?(item), do: :string, else: :binary)
```
josevalim, line 139: This operation is expensive. I think we should always default to string and users explicitly override if necessary.

josevalim (approved):

PR #440, opened 2022-11-29
Allow Explorer.PolarsBackend.Series.pow/2 on (series, series)

### LINKED ISSUE #375: Allow `Explorer.PolarsBackend.Series.pow/2` on (series, series)
`./lib/explorer/polars_backend/series.ex:216`

```elixir
# TODO: make pow/2 accept series on both sides.
@impl true
def pow(left, exponent) when is_float(exponent),
  do: Shared.apply_series(left, :s_pow, [exponent])

def pow(left, exponent) when is_integer(exponent) and exponent >= 0 do
  cond do
    Series.dtype(left) == :integer -> Shared.apply_series(left, :s_int_pow, [exponent])
    Series.dtype(left) == :float -> Shared.apply_series(left, :s_pow, [exponent / 1])
  end
end
```

nallwhy:
This PR resolves #375.

Because of my lack of Rust knowledge, I need feedback for the `s_pow/2` code.

### DIFF
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -662,6 +662,29 @@ pub fn s_n_distinct(data: ExSeries) -> Result<usize, ExplorerError> {
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
+pub fn s_pow(data: ExSeries, other: ExSeries) -> Result<ExSeries, ExplorerError> {
+    let s = &data.resource.0;
+    let s1 = &other.resource.0;
+
+    let iter1 = s.i64()?.into_iter();
+    match s1.strict_cast(&DataType::UInt32) {
+        Ok(s1) => {
+            let iter2 = s1.u32()?.into_iter();
+
+            let s = iter1
+                .zip(iter2)
+                .map(|(v1, v2)| v1.unwrap().pow(v2.unwrap()))
+                .collect();
+
+            Ok(ExSeries::new(s))
+        }
+        Err(_) => Err(ExplorerError::Other(String::from(
+            "negative exponent with an integer base",
+        ))),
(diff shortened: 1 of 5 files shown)


native/explorer/src/series.rs
```diff
@@ -661,6 +661,29 @@ pub fn s_n_distinct(data: ExSeries) -> Result<usize, ExplorerError> {
...
+    let s1 = &other.resource.0;
+
+    let iter1 = s.i64()?.into_iter();
+    match s1.strict_cast(&DataType::UInt32) {
+        Ok(s1) => {
+            let iter2 = s1.u32()?.into_iter();
+
+            let s = iter1
+                .zip(iter2)
+                .map(|(v1, v2)| v1.unwrap().pow(v2.unwrap()))
```
philss, line 676: Is it possible for one of values to be nil/null? Could you add a test for that if so?

native/explorer/src/series.rs
```diff
@@ -661,6 +661,29 @@ pub fn s_n_distinct(data: ExSeries) -> Result<usize, ExplorerError> {
...
+    let s = &data.resource.0;
+    let s1 = &other.resource.0;
+
+    let iter1 = s.i64()?.into_iter();
+    match s1.strict_cast(&DataType::UInt32) {
+        Ok(s1) => {
+            let iter2 = s1.u32()?.into_iter();
+
+            let s = iter1
+                .zip(iter2)
```
philss, line 675: What happens if the series is different in size?

philss (commented):
This looks like the way to go. Looking good!

philss:
I'm merging this and I should fix the issues that I pointed :)
Thanks!

PR #446, opened 2022-12-04
Add  s_nil_count

ghost:


josevalim:
Thank you @cristineguadelupe! I have a couple questions:

1. Wouldn't it be better to expose it as `Series.null_count` instead of the data frame? I would say it would be better because we would be able to use it in more situations (including queries).

2. Should this be called `null_count` or `nil_count` to mirror Elixir? Series has functions like `drop_nil` and `is_nil`... so I think it should be `nil_count`.

josevalim (approved):

philss:
@cristineguadelupe awesome! Thank you! :purple_heart:

PR #447, opened 2022-12-08
Avoid leaking UInt32 types to Elixir

philss:
We do a cast to `Int64`.

Closes https://github.com/elixir-nx/explorer/issues/431

### DIFF
--- a/native/explorer/src/expressions.rs
+++ b/native/explorer/src/expressions.rs
@@ -6,7 +6,7 @@
 
 use chrono::{NaiveDate, NaiveDateTime};
 use polars::prelude::{col, when, DataFrame, IntoLazy, LiteralValue, SortOptions};
-use polars::prelude::{DataType, Expr, Literal, Series};
+use polars::prelude::{Expr, Literal, Series};
 
 use crate::datatypes::{ExDate, ExDateTime};
 use crate::series::{cast_str_to_dtype, rolling_opts};
@@ -388,7 +388,7 @@ pub fn expr_nil_count(expr: ExExpr) -> ExExpr {
 pub fn expr_n_distinct(expr: ExExpr) -> ExExpr {
     let expr: Expr = expr.resource.0.clone();
 
-    ExExpr::new(expr.n_unique().cast(DataType::Int64))
+    ExExpr::new(expr.n_unique())
 }
 
 #[rustler::nif]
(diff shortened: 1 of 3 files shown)


native/explorer/src/expressions.rs
```diff
@@ -504,7 +504,7 @@ pub fn expr_argsort(expr: ExExpr, descending: bool, nulls_last: bool) -> ExExpr
         nulls_last,
     };
 
-    ExExpr::new(expr.arg_sort(opts).cast(DataType::Int64))
+    ExExpr::new(expr.arg_sort(opts))
```
josevalim, line 507: won't this return uint32?

native/explorer/src/expressions.rs
```diff
@@ -388,7 +388,7 @@ pub fn expr_nil_count(expr: ExExpr) -> ExExpr {
 pub fn expr_n_distinct(expr: ExExpr) -> ExExpr {
     let expr: Expr = expr.resource.0.clone();
 
-    ExExpr::new(expr.n_unique().cast(DataType::Int64))
+    ExExpr::new(expr.n_unique())
```
josevalim, line 391: won't this return uint32?

philss: Yes, it will. And we will cast after the expressions are applied. Do you prefer to keep cast it on both places?

philss: (not that cast will be performed twice. Just once at either places).

josevalim: @philss i am worried that something like `mutate new_column: n_distinct(df["foo"])` will return UInt32, or am I wrong?

josevalim: Oh, ignore me. I see what you are doing. No, I think the current version is better. I think with your version this code will fail:

    mutate new_column: -n_distinct(df["foo"])

josevalim: or maybe we need both. but ideally we should always cast at the expression.

josevalim: Ok, so let me change the question: why do we need this patch? why isn't the previous code enough? It feels like something else is leaking UInt32 :)

philss: It is to fix the bug with `expr_count` that does not cast correctly (the comment from the issue https://github.com/elixir-nx/explorer/issues/431#issuecomment-1328219874).

Maybe this is not the ideal way of doing it, since I'm trying to fix a problem from Polars.
I can close this and try to reproduce the problem with polars and the "count" expr.

josevalim: Oh, I see. let's see if we can fix that in Polars then?

philss: Let's do this.. I'm going to prepare a reproduction example app. In the meanwhile I'm closing this PR :)

Issue #353, opened 2022-09-14
Regex support?


liveforeverx:
First, thank you for great library!

Do you plan (or when do you plan) to support regex manipulations in polars?

Example would be, build a new column, which extracts multiple matches from another column using regex expression and combining output. Is it possible currently archive in another way?

I would be open to contribute them with some guidance.

kimjoaoun:
Isn't it possible doing so using Elixir's already existing regex capabilities?

I don't know if that's exactly what you want, but I think you might write a function that returns the desired value after evaluating a cell with `Regex.match?/2`, and then you could map this function over an existing column, and using `Explorer.DataFrame.mutate` it is possible create a new one.

josevalim:
You can use Series.transform with Elixir regexes but that requires materializing all results to Elixir and back, which is not the most efficient format.

cigrainger:
https://pola-rs.github.io/polars-book/user-guide/howcani/data/strings.html

So this one is certainly possible but opens up a bit of a can of worms. In the Python Polars, there is a string namespace. I think we could totally do that. Something like `Series.String.lengths` or `Series.String.filter`. There's a regex builder re-exported by polars so that's also pretty straightforward. So I guess it's mostly a design decision. As someone who works with a lot of text, I think it'd be very useful. @philss @josevalim what do you think about something like `Series.String`? Is there another option I'm not seeing? The only pause I have is whether we'd want to then namespace other things. E.g. `Series.List` in support of #296?

josevalim:
Honestly, I am not the biggest fan of the namespacing, it feels everything could fit in a single namespace and in the worst case we use `string_` prefixes? It looks like less to manage and closer to SQL too?

cigrainger:
Sure, I'm just a bit wary of confusing verbiage but agreed it's best to give it a try in the single namespace first and if it feels painful look at `string_` prefixes.

philss:
I'm planning to close this issue this week. We now have two new functions to work with regexes: `re_contains/2` and `re_replace/3` - see #894.

I would like to add 4 more functions:
- `Series.count_matches/2` mirroring Python Polars' [`Series.count_matches/2`](https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.str.count_matches.html), but only for literals.
- `Series.re_count_matches/2` which is the same function from above, but with regex support.
[...]

PR #449, opened 2022-12-13
Add Series string functions

cigrainger:
Contributes to #353.

Adds:
- `Series.contains/2`
- ~`Series.extract/2`~
- `Series.trim/1`
- `Series.trim_leading/1`
- `Series.trim_trailing/1`
- `Series.upcase/1`
- `Series.downcase/1`

I had a go at `Series.join/2` (rhyming with `Enum.join/2`, but vectorised) and could make it work for eager but wasn't able to figure out a suitable method for lazy: https://docs.rs/polars/0.25.1/polars/prelude/string/struct.StringNameSpace.html.


lib/explorer/series.ex
```diff
@@ -2797,6 +2797,26 @@ defmodule Explorer.Series do
...
+      iex> s = Explorer.Series.from_list(["abc", "def", "bcd"])
+      iex> Explorer.Series.contains(s, "bc")
+      #Explorer.Series<
+        Polars[3]
+        boolean [true, false, true]
+      >
+  """
+  @doc type: :element_wise
+  @spec contains(Series.t(), String.t()) :: Series.t()
+  def contains(%Series{dtype: :string} = series, pattern),
```
philss, line 2816: Maybe we could have a guard here to prevent anything other then strings/binaries. WDYT?

philss (approved):
Awesome! Just a minor thing and :shipit:

lib/explorer/series.ex
cigrainger: Done 👍

lib/explorer/series.ex
```diff
@@ -2797,6 +2797,155 @@ defmodule Explorer.Series do
...
+      >
+  """
+  @doc type: :element_wise
+  @spec contains(Series.t(), String.t() | Regex.t()) :: Series.t()
+  def contains(%Series{dtype: :string} = series, pattern)
+      when K.or(K.is_binary(pattern), K.is_struct(pattern, Regex)),
+      do: Shared.apply_impl(series, :contains, [pattern])
+
+  def contains(%Series{dtype: :string} = series, literal) when is_binary(literal),
+    do: Shared.apply_impl(series, :contains_literal, [literal])
```
josevalim, line 2828: This is polars specific and I Don't think we should expose it here. :) In fact, given that you match on K.is_binary in the previous clause, I think this whole clause and all mentions to `contains_literal` can be removed?

lib/explorer/series.ex
```diff
@@ -2797,6 +2797,155 @@ defmodule Explorer.Series do
...
+
+      iex> s = Explorer.Series.from_list(["Explorer the library", "is a really", "great library"])
+      iex> Explorer.Series.extract(s, ~r/^(\\w+)\\b/)
+      #Explorer.Series<
+        Polars[3]
+        string ["Explorer", "is", "great"]
+      >
+  """
+  @doc type: :element_wise
+  @spec extract(Series.t(), Regex.t(), pos_integer()) :: Series.t()
```
josevalim, line 2942: This is called `Regex.scan` in Elixir. However if we call it `scan` it may get confused with `Enum.scan`. Thoughts?

lib/explorer/backend/lazy_series.ex
```diff
@@ -549,6 +558,61 @@ defmodule Explorer.Backend.LazySeries do
...
+
+  def contains(series, pattern) when is_binary(pattern) do
+    data = new(:contains_literal, [lazy_series!(series), pattern])
+
+    Backend.Series.new(data, :boolean)
+  end
+
+  @impl true
+  def upcase(series) do
+    data = new(:to_uppercase, [lazy_series!(series)])
```
josevalim, line 576: LazySeries is `polars` agnostic. So we should stick with Explorer names instead of Polars ones. Same applies below!

native/explorer/src/lib.rs
```diff
@@ -280,6 +296,7 @@ rustler::init!(
         s_slice_by_indices,
         s_sort,
         s_standard_deviation,
+        s_strip,
```
josevalim, line 299: I also think we have been naming those functions after the Elixir side. I.e. we should let Elixir dictates the names as much as possible. :)

lib/explorer/polars_backend/series.ex
```diff
@@ -422,6 +422,39 @@ defmodule Explorer.PolarsBackend.Series do
   @impl true
   def unary_not(%Series{} = series), do: Shared.apply_series(series, :s_not, [])
 
+  # Strings
+
+  @impl true
+  def contains(series, %Regex{source: pattern}),
+    do: Shared.apply_series(series, :s_contains, [pattern])
```
josevalim, line 429: You also need to pass the opts field, so you get the modifiers, such as:

    ~r/foo/i

For case insensitive. You may need to check if all Elixir modifiers map to Rust modifiers.

josevalim: Another concern is that regexes in Elixir are not Unicode by default but they seem to be in Polars. I am not even sure if they have a non-Unicode variant. So we should perhaps expect "u" to be set in `opts`? Such as `~r/foo/u` or somehow [disable Unicode](https://docs.rs/regex/latest/regex/#opt-out-of-unicode-support) if "u" is not specified.

josevalim: Perhaps remove the Regex support from this PR we work on the foundation for it in the next one?

lib/explorer/series.ex
cigrainger: Ah I thought I had cleaned that up already! Must have snuck past me when I rebased.

lib/explorer/series.ex
cigrainger: Maybe this is a case where it would make sense to prefix the function name? `string_scan`?

lib/explorer/backend/lazy_series.ex
cigrainger: Good point. I'll fix them up.

native/explorer/src/lib.rs
cigrainger: Ah really? My mental model was the opposite, but I'm not attached to it at all. :) Happy to go that way.

lib/explorer/polars_backend/series.ex
cigrainger: Sure, I'll take it out for now.

native/explorer/src/lib.rs
josevalim: For sure in the Backend we want to have Elixir specific names. Here whatever name is fine but it seems we are mostly aligned to the Elixir side? :)

lib/explorer/series.ex
josevalim: After more thought, I think scan is fine at the end of the day because it is ultimately different signatures? :) and I think the operations you do with scan are all cumulative operations.

cigrainger: So should we keep this function in and just use a `String.t()` for the pattern or should we drop it?

josevalim: I am not sure if it makes sense to scan a static string? It will either return the string itself or nil, so it is basically a contain. So perhaps hold it off until we support regexes which can be next?

cigrainger: Yep 👍

josevalim (approved):

Issue #227, opened 2022-05-17
Lazy Polars backend functionality


cigrainger:
This issue tracks the implementation for full backend functionality for lazy Polars. The intent is to enable splitting things up more easily.

## IO read

- [x] `from_csv` (#451)
- [x] `load_csv` (#451)
- [x] `from_parquet` (#451)
- [x] `load_parquet` (#451)
- [x] `from_ipc` (#451)
- [x] `load_ipc` (#451)
- [x] `from_ndjson` (#451)
- [x] `load_ndjson` (#451)
- [x] `load_ipc_stream` (b326afd48)

PS: `from_ipc_stream` for the Lazy backend is not possible because is not available in Polars. 

## IO write

- [x] Certain IO write operations can be done lazily for streaming: https://genserver.social/notice/AQn5LNxxRIfvNvvkNE - we need to investigate which and break it down
- [x] `to_parquet` (#496) 
- [x] `to_ipc` (#497) - available only in main branch of Polars 
- [x] `to_ipc_stream` (#497) - It's not streamable, even so the name contains "stream". `collect() |> write()`
- [x] `to_csv` (#497) - It's not streamable. `collect() |> write()`
- [x] `to_ndjson` (#497) -  It's not streamable. `collect() |> write()`

## Conversion

- [x] `lazy` (#202)
- [x] `to_lazy` (#202)
- [x] `collect` (#202)

## Introspection

- [x] `names` (#202)
- [x] `dtypes` (#202)
- [x] `shape` (#497)
- [x] `n_rows` (#497)
- [x] `n_columns` (#202)

## Single table verbs

- [x] `head` (#202)
- [x] `tail` (#202)
- [x] `select` (#202)
- [x] `filter` (#476)
- [x] `mutate`
- [x] `arrange`
- [x] `distinct`
- [x] `rename` (#488)
- [x] `dummies` (#497)
- [x] `sample` (#497) - this one may be possible to implement, but not sure how.
- [x] `pull` (#497)
- [x] `slice` (#450)
- [x] `drop_nil` (#488)
- [x] `pivot_wider` (#497)
- [x] `pivot_longer` (#488)
- [x] `group_by` (#202)
- [x] `ungroup` (#202)
- [x] `summarise`

## Two or more table verbs

- [x] `join` (#491)
- [x] `concat_rows` (#492)
- [x] `concat_columns` (#492)

josevalim:
Regarding the IO functions, only half of them need to be implemented, right?

[...]

PR #451, opened 2022-12-16
Add IO operations "from/load" to the LazyFrame implementation

philss:
This is related to #227 

In the first PR for that issue, #202, we [discussed](https://github.com/elixir-nx/explorer/pull/202#discussion_r876379528) a possible `lazy: true` option to control which backend to use. I end up using both `:backend` and `:lazy` to make it easier for clients to choose the backend. This adds the possibility for a backend to be implemented both using an eager (the default) and lazy version. WDYT?

I also choose to "delegate" the load of binaries to the eager implementation, and then converting it to lazy. This is mostly for convenience, because there is not such loaders in the Polars backend.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -394,6 +394,8 @@ defmodule Explorer.DataFrame do
   @spec from_csv(filename :: String.t(), opts :: Keyword.t()) ::
           {:ok, DataFrame.t()} | {:error, term()}
   def from_csv(filename, opts \\ []) do
+    {backend_opts, opts} = Keyword.split(opts, [:backend, :lazy])
+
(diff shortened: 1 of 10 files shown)


cigrainger:
I love the direction. Definitely like having both `:backend` and `:lazy` opts. Yeah I think that also makes sense regarding loading binaries. It's in memory already anyway so you don't lose much compared to loading from file. Definitely should document that clearly though.

josevalim (approved):

philss:
> Definitely should document that clearly though.

@cigrainger You mean about the `:lazy` option or the usage of the eager by the `load_` functions?
Anyway, I'm going to add documentation in a future PR.

PR #453, opened 2022-12-17
Add mode

### LINKED ISSUE #452: compute mode on grouped DataFrame
I have much more experience with R and Python and love the "dataframe" support in Elixir. 

I'm trying to compute the mode in a grouped DataFrame and hit a error with calling `frequencies/1` on a `LazySeries`. (I also can't call `to_list/1` to do the mode in native Elixir.) A minimal working example is below.

I have some ideas for workarounds outside of `Explorer`, but don't want to re-implement the grouping. Or perhaps there is a way to do it within `Explorer` currently.

Any ideas?

```
alias(Explorer.DataFrame, [as: DF])
alias(Explorer.Series, [as: S])
require Explorer.DataFrame

defmodule Stats do
    def mode(series) do
      series
      |> S.frequencies()
      |> DF.arrange(desc: counts, asc: values)
      |> DF.head(1)
      |> DF.to_rows()
      |> Enum.at(0)
      |> Map.get("values")
    end
end

df = DF.new(gp: ["a", "a", "b", "b", "a", "b", "b"], val: [1, 2, 3, 4, 1, 4, 1])
df_group = DF.group_by(df, "gp")

# mode on series works
df["val"] |> Stats.mode()

# summarise works here
#   #Explorer.DataFrame<
#     Polars[2 x 3]
#     gp string ["a", "b"]
#     min integer [1, 1]
#     max integer [2, 4]
#   >
DF.summarise_with(df_group, fn x ->
  min = S.min(x["val"])
  max = S.max(x["val"])
  [min: min, max: max]
end)

# fails
#   ** (RuntimeError) cannot perform frequencies/1 operation on Explorer.Backend.LazySeries.
#   Query operations work on lazy series and those support only a subset of series operations
DF.summarise_with(df_group, fn x ->
  min = S.min(x["val"])
  max = S.max(x["val"])
  mode = Stats.mode(x["val"])
  [min: min, max: max, mode: mode]
end)
```

cigrainger:
~See: #452. This is a WIP because I'm not sure how to handle the fact that you can have no mode or multiple. This gets particularly hairy in a groupby scenario where each group may have different lengths (and thus would cause problems with other summary statistics like mean, median where you know the length will be 1).~

Pandas:
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.mode.html
https://pandas.pydata.org/docs/reference/api/pandas.Series.mode.html#pandas.Series.mode
https://stackoverflow.com/a/54304691

R:
https://stackoverflow.com/questions/66972590/how-to-find-mean-median-mode-based-on-distinctive-groups-in-r
https://cran.r-project.org/web/packages/modeest/modeest.pdf

Okay! Thanks to #725 this is good to go. Closes #452.


lib/explorer/series.ex
```diff
@@ -1241,6 +1241,27 @@ defmodule Explorer.Series do
...
+    * `:float`
+
+  ## Examples
+
+      iex> s = Explorer.Series.from_list([1, 2, 2, 3])
+      iex> Explorer.Series.mode(s)
+      2
+  """
+  @doc type: :aggregation
+  @spec mode(series :: Series.t()) :: float() | nil
```
philss, line 1259: Float or int, right?

```suggestion
  @spec mode(series :: Series.t()) :: number() | nil
```

philss (commented):
> This gets particularly hairy in a groupby scenario where each group may have different lengths

I think this is fine, because I think the computation is going to occur inside the group's context. Maybe adding a test - actually change an existing one with another column - to the `summarise` or `mutate` can clarify a little bit. WDYT?

cigrainger:
> I think this is fine, because I think the computation is going to occur inside the group's context

Yes, between groups, but not within groups. So take the example where you summarise down to mean and mode. If mode is length 0 or >1, then you have a problem where within groups the aggregations are different lengths. Adding a test or two now to clarify as suggested 👍.

cigrainger:
Well it clarified in a different direction: summarise w/ mode returns a single row per group -- a list row! We'll have to pick up #401 anyway.

josevalim (approved):
LGTM although it seems we are waiting on lists?

lib/explorer/series.ex
```diff
@@ -2164,6 +2165,51 @@ defmodule Explorer.Series do
...
+      iex> s = Explorer.Series.from_list(["a", "b", "b", "c"])
+      iex> Explorer.Series.mode(s)
+      #Explorer.Series<
+        Polars[1]
+        string ["b"]
+      >
+  """
+  @doc type: :aggregation
+  @spec mode(series :: Series.t()) :: Series.t() | nil
+  def mode(%Series{dtype: dtype} = series) when is_numeric_or_io_dtype(dtype),
```
josevalim, line 2208: We use iodtype with other meaning, which would include datetimes, time, etc, so I would call it "is_numeric_or_binary_dtype". Also, would categories work here? They are either integers or strings.

lib/explorer/series.ex
```diff
@@ -2164,6 +2165,51 @@ defmodule Explorer.Series do
 
   def mean(%Series{dtype: dtype}), do: dtype_error("mean/1", dtype, [:integer, :float])
 
+  @doc """
+  Gets the mode value of the series.
+
+  ## Supported dtypes
+
+    * `:integer`
+    * `:float`
```
josevalim, line 2174: We need to update this list :)

lib/explorer/series.ex
cigrainger: Good question! I think they would.

cigrainger: Honestly I think it should support any dtype. I'll just have to see what Polars supports and if there's a workaround.

cigrainger: Yep it works for all dtypes! 🎉

lib/explorer/series.ex
```diff
@@ -2164,6 +2164,53 @@ defmodule Explorer.Series do
 
   def mean(%Series{dtype: dtype}), do: dtype_error("mean/1", dtype, [:integer, :float])
 
+  @doc """
+  Gets the most common value of the series.
+
+  ## Supported dtypes
```
josevalim, line 2170: Let’s just say all except lists? Otherwise it is easy for this to go out of date!

lib/explorer/shared.ex
```diff
@@ -9,7 +9,16 @@ defmodule Explorer.Shared do
   within lists inside.
   """
   def dtypes do
-    non_list_dtypes = [
+    non_list_dtypes = non_list_types()
+    list_dtypes = for dtype <- non_list_dtypes, do: {:list, dtype}
+    non_list_dtypes ++ list_dtypes
```
josevalim, line 14: List types are recursive so this is theoretically incomplete. Instead of doing this list, what if we just append `{:list, :any}` instead?

cigrainger: Sure, fine by me 👍

lib/explorer/series.ex
cigrainger: Yep fair 😂

lib/explorer/series.ex
```diff
@@ -2164,6 +2164,44 @@ defmodule Explorer.Series do
 
   def mean(%Series{dtype: dtype}), do: dtype_error("mean/1", dtype, [:integer, :float])
 
+  @doc """
+  Gets the most common value of the series.
```
billylanchantin, line 2168:
```suggestion
  Gets the most common value(s) of the series.

  This function will return multiple values when there's a tie.
```

billylanchantin (approved):
Woo! It's cool to this this PR get across the finish line!

I just had one suggestion about the docs: I think it's worth calling out ties.

lib/explorer/shared.ex
cigrainger: So it looks like this is used for tests? I'm going to leave it how it was (I just wrapped the non-list ones into a separate function) and then I think we should revisit it in a future PR.

josevalim (approved):
Ship it and I can look at the dtypes stuff. :)

PR #454, opened 2022-12-28
Percentiles for describe

cristineguadelupe:


lib/explorer/data_frame.ex
```diff
@@ -4443,10 +4444,10 @@ defmodule Explorer.DataFrame do
   """
   @doc type: :single
   @spec describe(df :: DataFrame.t()) :: DataFrame.t()
-  def describe(df) do
+  def describe(df, percentiles \\ @default_percentiles) do
```
josevalim, line 4447: Let's make this an option, as we will likely have other percentiles in the future!

josevalim: Note we only need to make it an option here. It does not have to be an option in the behaviour.

philss (approved):
:shipit:

josevalim:

PR #456, opened 2022-12-29
Add support for RISCV and Linux on M1/2 with musl

philss:
This adds the precompilation steps and config for those targets.

They are going to be default on `rustler_precompiled` v0.6, but since we may need to support both versions, we are declaring them explicitly.

This also updates the rust toolchain version, and "cross" version.


cigrainger (approved):
:shipit: ❤️

philss:
I'm not sure why it didn't run - the release workflow - for this branch, so I'm going to merge and fix if this breaks `main`.

PR #460, opened 2023-01-02
Allow DataFrame.new/1 to receive the dtypes option

Jhonatannunessilva:
This PR resolves https://github.com/elixir-nx/explorer/issues/458.

### DIFF
--- a/lib/explorer/polars_backend/data_frame.ex
+++ b/lib/explorer/polars_backend/data_frame.ex
@@ -283,10 +283,11 @@ defmodule Explorer.PolarsBackend.DataFrame do
   def from_tabular(tabular, dtypes) do
     {_, %{columns: keys}, _} = reader = init_reader!(tabular)
     columns = Table.to_columns(reader)
+    dtypes = Map.new(dtypes)
 
     keys
     |> Enum.map(fn key ->
-      dtype = dtypes |> Map.new() |> Map.get(key)
+      dtype = Map.get(dtypes, key)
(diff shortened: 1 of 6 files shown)


lib/explorer/polars_backend/data_frame.ex
```diff
@@ -280,15 +280,16 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
 
   @impl true
-  def from_tabular(tabular) do
+  def from_tabular(tabular, dtypes) do
     {_, %{columns: keys}, _} = reader = init_reader!(tabular)
     columns = Table.to_columns(reader)
 
     keys
     |> Enum.map(fn key ->
+      dtype = dtypes |> Map.new() |> Map.get(key)
```
philss, line 289: I think we don't need to create a new map for each key. You could move the `dtypes |> Map.new()` to outside the `Enum.map/2` function.

Alternatively you could use [`List.keyfind/3`](https://hexdocs.pm/elixir/List.html#keyfind/4) instead of creating a map.

philss (commented):
Looking great! Just one comment :)

lib/explorer/polars_backend/data_frame.ex
Jhonatannunessilva: Yes, it makes sense

philss (approved):
:shipit:

philss:
@Jhonatannunessilva thank you! :purple_heart:

Issue #261, opened 2022-06-22
Add categorical/factors dtype


cigrainger:
Pandas: https://pandas.pydata.org/docs/user_guide/categorical.html
R: https://r4ds.had.co.nz/factors.html
Polars: https://docs.rs/polars/0.21.1/polars/docs/performance/index.html#categorical-type

cigrainger:
@jsonbecker Do you have thoughts on ergonomics/ease of use with factors? What can we do to make them easy to work with and intuitive beyond copying R?

jsonbecker:
I have controversial opinions about factors. In my mind, there are two reasons to use factors:

1. Data validation
2. Other functions/packages further down the chain can be smart about discrete values with orders reducing code I have to write later.

I think reason (1) is dumb. Assertive programming of various types are better for explicit validation. On (2), I'm not sure how much the broader Nx ecosystem is going to be "smart" about these things.

I find factors in R _fantastic_ once they exist correctly because much of the ecosystem can then interpret them well. I find the ergonomics of working directly with them to be abysmal.

So how can working with factors be less abysmal? I'd say a few things help.

1. Factors have to be explicitly cast as such-- nothing should attempt to coerce a factor at any time, and it's better to error when supplied a non-factor data type if a factor is expected.
2. Factors should have an easy to work with interface to the allowable values and their order. This should be easily access and updated without having to rewrite the data-- it's really just a map of sequence and allowable values.
3. Unordered factors are pointless. Semantically they exist, but I see no reason to distinguish them at the data type level. Some operations will be order aware, some will not. If the factor is unordered, being aware of an explicit order existing won't have any effect since it's effectively random. It's sort of like setting a seed. Who cares it's consistently in an order that doesn't matter, and you can always randomize the order again yourself easily if that interface is simple to work with.

josevalim:
Categorical types are in, mostly for integration with Nx. We can revisit this issue with more integrated features later on if desired.

PR #464, opened 2023-01-08
Add basic support for categorical dtype

philss:
This is the first step to support categorical series.

There are some limitations to what can be done in the Polars side, so I'm not sure yet how I can manipulate the categories, like adding or removing members, and how to define ordering.

And also maybe we should call it `:category` dtype. WDYT?

This is related to #261


lib/explorer/series.ex
```diff
@@ -9,6 +9,7 @@ defmodule Explorer.Series do
     * `:boolean` - Boolean
     * `:string` - UTF-8 encoded binary
     * `:binary` - Binary
+    * `:categorical` - UTF-8 encoded binary, but as categories.
```
josevalim, line 12:
```suggestion
    * `:categorical` - UTF-8 encoded binary, but as categories
```

:D

Maybe let's sort them alphabetically too?

native/explorer/src/encoding.rs
```diff
@@ -354,6 +387,7 @@ pub fn list_from_series(data: ExSeries, env: Env) -> Result<Term, ExplorerError>
         DataType::Binary => {
             generic_binary_series_to_list(&data.resource, s.binary()?.downcast_iter(), env)
         }
+        DataType::Categorical(Some(mapping)) => categorical_series_to_list(s, env, mapping),
```
josevalim, line 390: This is tricky. I wonder if we should return the categories (as strings) or their indexes in the mapping. And I think `term_from_value` should return the string.

`iovec_from_series` should definitely return indexes though.

josevalim: We don't want to avoid UInt32. But we want to avoid multiple representations for the same dtype. I.e. integers should have only a single representation. Therefore it is fine to return UInt32 for categorical types. :)

PR #465, opened 2023-01-09
Add Series.categories/1

philss:
This function is useful to retrieve only the categories from a categorical series.
The resultant series is of ":string" dtype.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -689,6 +689,33 @@ defmodule Explorer.Series do
   @spec bintype(series :: Series.t()) :: {:s | :u | :f, non_neg_integer()}
   def bintype(series), do: Shared.apply_impl(series, :bintype)
 
+  @doc """
+  Return a series with the category names of a categorical series.
+
+  Each category has the index equal to its position.
+
+  ## Examples
+
+      iex> s = Explorer.Series.from_list(["a", "b", "c", nil, "a", "c"], dtype: :category)
+      iex> Explorer.Series.categories(s)
+      #Explorer.Series<
+        Polars[3]
+        string ["a", "b", "c"]
+      >
+
+      iex> s = Explorer.Series.from_list(["c", "a", "b"], dtype: :category)
+      iex> Explorer.Series.categories(s)
+      #Explorer.Series<
+        Polars[3]
+        string ["c", "a", "b"]
+      >
+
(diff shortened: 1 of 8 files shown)


lib/explorer/series.ex
```diff
@@ -689,6 +689,33 @@ defmodule Explorer.Series do
   @spec bintype(series :: Series.t()) :: {:s | :u | :f, non_neg_integer()}
   def bintype(series), do: Shared.apply_impl(series, :bintype)
 
+  @doc """
+  Return a series with the category names of a categorical series.
+
+  Each category has the index equal to its position.
```
josevalim, line 695: Should we add a note that no order is guaranteed for the categories?

native/explorer/src/series.rs
```diff
@@ -779,6 +779,22 @@ pub fn cast_str_to_dtype(str_type: &str) -> Result<DataType, ExplorerError> {
...
+pub fn s_categories(data: ExSeries) -> Result<ExSeries, ExplorerError> {
+    let s: &Series = &data.resource.0;
+    match s.dtype() {
+        DataType::Categorical(Some(mapping)) => {
+            let size = mapping.len() as u32;
+            let categories: Vec<&str> = (0..size).map(|id| mapping.get(id)).collect();
+            let series = Series::new("categories", &categories);
+            Ok(ExSeries::new(series))
+        }
+        _ => Err(ExplorerError::Other(
```
josevalim, line 792: We should panic because this will never be invoked with something else, right?

lib/explorer/series.ex
philss: I'm not sure. For what I could understand, the ordering is defined by the order the elements appear in the series. But this is not clear.

Internally there is a flag to the the ordering to lexical: https://pola-rs.github.io/polars/polars_core/datatypes/categorical/struct.CategoricalChunked.html#method.set_lexical_sorted

josevalim: I think it is defined like that but mostly as an implementation accident and not a guarantee? Providing fewer guarantees doesn’t hurt anyway :)

josevalim (approved):

lib/explorer/series.ex
philss: Makes sense. Note added :+1:

PR #470, opened 2023-01-12
Normalize numeric dtypes when loading/reading from IO

### LINKED ISSUE #468: Support for more datatypes
When trying to load some of my Parquet encoded data (`Explorer.DataFrame.load_parquet!/1`) that include 16-bit unsigned integers I get an error because of the lack of a function clause matching `Explorer.PolarsBackend.Shared.normalise_dtype("u16")`.

To allow the most basic handling of such data I suppose at the minimum the following definitions would need to be changed:

https://github.com/elixir-nx/explorer/blob/664bf7a2b4b0dbb0c9d13c0cc7c7bc294cfcb9e8/lib/explorer/polars_backend/series.ex#L46-L62

https://github.com/elixir-nx/explorer/blob/664bf7a2b4b0dbb0c9d13c0cc7c7bc294cfcb9e8/lib/explorer/polars_backend/shared.ex#L127-L140

adding e.g.

```elixir
      "u16" -> {:u, 16}
      "u64" -> {:u, 64}
      "i8" -> {:s, 8}
      "i16" -> {:s, 16}
      "f32" -> {:f, 32}
```
and

```elixir
  def normalise_dtype("u16"), do: :integer
  def normalise_dtype("u64"), do: :integer
  def normalise_dtype("i8"), do: :integer
  def normalise_dtype("i16"), do: :integer
  def normalise_dtype("f32"), do: :float
```

respectively. Is the basic import of this data within the scope of the project? Should the columns be transformed into the main supported types on import? Is full support for any of these types within the scope of the project?

philss:
The idea is to keep the numeric dtypes only using Int64 for integers or Float64 for floats.

I tested with some parquet files from this project: https://github.com/PRQL/prql-query and they all worked fine.

 Closes #468

### DIFF
--- a/native/explorer/src/dataframe/io.rs
+++ b/native/explorer/src/dataframe/io.rs
@@ -16,6 +16,7 @@ use std::fs::File;
 use std::io::{BufReader, BufWriter, Cursor};
 use std::result::Result;
 
+use crate::dataframe::normalize_numeric_dtypes;
 use crate::{ExDataFrame, ExplorerError};
 
 // ============ CSV ============ //
@@ -48,7 +49,7 @@ pub fn df_from_csv(
         None => None,
     };
 
-    let df = CsvReader::from_path(filename)?
+    let mut df = CsvReader::from_path(filename)?
         .infer_schema(infer_schema_length)
         .has_header(has_header)
         .with_parse_dates(parse_dates)
@@ -63,7 +64,9 @@ pub fn df_from_csv(
         .with_null_values(Some(NullValues::AllColumns(vec![null_char])))
         .finish()?;
 
-    Ok(ExDataFrame::new(df))
+    let normalized_df = normalize_numeric_dtypes(&mut df)?;
+
+    Ok(ExDataFrame::new(normalized_df))
 }
 
 pub fn schema_from_dtypes_pairs(dtypes: Vec<(&str, &str)>) -> Result<Schema, ExplorerError> {
@@ -157,7 +160,7 @@ pub fn df_load_csv(
 
     let cursor = Cursor::new(binary.as_slice());
 
-    let df = CsvReader::new(cursor)
+    let mut df = CsvReader::new(cursor)
         .infer_schema(infer_schema_length)
         .has_header(has_header)
         .with_parse_dates(parse_dates)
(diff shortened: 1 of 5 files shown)


native/explorer/src/dataframe/io.rs
```diff
@@ -356,11 +370,14 @@ pub fn df_from_ipc_stream(
...
-    let df = IpcStreamReader::new(buf_reader)
+    let mut df = IpcStreamReader::new(buf_reader)
         .with_columns(columns)
         .with_projection(projection)
         .finish()?;
-    Ok(ExDataFrame::new(df))
+
+    let normalized_df = normalize_numeric_dtypes(&mut df)?;
+
+    Ok(ExDataFrame::new(normalized_df))
```
ahnlabb, line 380: Could this part be factored out into a function by using the `SerReader` trait so that future changes to the import throu of SerReader do not miss any of the functions?

ahnlabb (commented):
This is great @philss! The change is still missing for binary data however:

https://github.com/elixir-nx/explorer/blob/25f31bb7397c8eb41f52e81ad6365e0218a43993/native/explorer/src/dataframe/io.rs#L270-L275

Perhaps future misses like that could be avoided by increasing code sharing between the functions a bit?

native/explorer/src/lazyframe.rs
```diff
@@ -7,7 +7,11 @@ pub mod io;
 
 #[rustler::nif(schedule = "DirtyCpu")]
 pub fn lf_collect(data: ExLazyFrame) -> Result<ExDataFrame, ExplorerError> {
-    Ok(ExDataFrame::new(data.resource.0.clone().collect()?))
+    let mut df = data.resource.0.clone().collect()?;
+
+    let normalized_df = crate::dataframe::normalize_numeric_dtypes(&mut df)?;
+
+    Ok(ExDataFrame::new(normalized_df))
```
josevalim, line 14: Doing this change here is too late because all operations in the lazy frame would then act on i32, u32 and so on. We also need to do them immediately after reading the file/binary.

philss: I see. I don't know yet how to apply this change for a LazyFrame, so I decided to revert and keep it just for the  eager frame. I think I'm going to open an issue for that and tackle after the release. WDYT?

native/explorer/src/dataframe/io.rs
philss: Thanks for the suggestion! I'm not sure if I understood correctly, but I added a function to abstract the "finish" process, which is more clear now. Please check again :)

philss:
> This is great @philss! The change is still missing for binary data however:

I forgot the most important :sweat_smile: 
It is fixed now.

josevalim (approved):

native/explorer/src/dataframe/io.rs
philss: I'm merging this in order to add to the new release. Please fell free to open a PR if you think this can be improved.

PR #476, opened 2023-01-18
Add `filter_with/2` for Polars LazyFrame backend

philss:
This function has two versions:

* one that works for ungrouped lazy frames and grouped, but without aggregations.
* another one that aims grouped lazy frames with aggregations in the filter expression.

The second one is more complex, so I opt to add it to a separated function.

### DIFF
--- a/native/explorer/src/lazyframe.rs
+++ b/native/explorer/src/lazyframe.rs
@@ -73,7 +73,7 @@ pub fn lf_slice(data: ExLazyFrame, offset: i64, length: u32) -> Result<ExLazyFra
     Ok(ExLazyFrame::new(lf.slice(offset, length)))
 }
 
-#[rustler::nif(schedule = "DirtyCpu")]
+#[rustler::nif]
 pub fn lf_filter_with(data: ExLazyFrame, ex_expr: ExExpr) -> Result<ExLazyFrame, ExplorerError> {
     let ldf = data.clone_inner();
     let expr = ex_expr.clone_inner();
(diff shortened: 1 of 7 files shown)


native/explorer/src/lazyframe.rs
```diff
@@ -72,3 +72,35 @@ pub fn lf_slice(data: ExLazyFrame, offset: i64, length: u32) -> Result<ExLazyFra
...
+    data: ExLazyFrame,
+    ex_expr: ExExpr,
+    groups: Vec<&str>,
+) -> Result<ExLazyFrame, ExplorerError> {
+    let ldf = data.clone_inner();
+    let aggs: Vec<Expr> = ldf
+        .schema()?
+        .iter_names()
+        .filter(|name| !groups.contains(&name.as_str()))
+        .map(|name| col(name).filter(ex_expr.clone_inner()).list().keep_name())
```
josevalim, line 95: I think we may be able to break it apart.

For example, when we call group in the lazy frame, we emit `ldf.groupby_stable`. Then when we call filter, we transform the aggregations, and so on. And we explode only at the end when we ungroup. Ping me today and we can talk. :)

josevalim (approved):

native/explorer/src/lazyframe.rs
philss: Just to keep track: we decided to not support the groupby with aggregations for this PR. I reverted in https://github.com/elixir-nx/explorer/pull/476/commits/a6f4b761a3b82c0368aea2282794f77ff77455b7

PR #483, opened 2023-01-23
Allow Series.fill_missing/2 to receive date and datetime values

Jhonatannunessilva:
This PR allows the Series.fill_missing/2 function to receive date and datetime values.

Before this PR, if you try to send a date or datetime value to the Series.fill_missing/2 function, an error will be returned:

```elixir
iex(1)> s1 = Explorer.Series.from_list([~D[2023-01-17], ~D[2023-01-18], nil, ~D[2023-01-09]])
#Explorer.Series<
  Polars[4]
  date [2023-01-17, 2023-01-18, nil, 2023-01-09]
>

iex(2)> Explorer.Series.fill_missing(s1, ~D[2023-01-19])
** (CondClauseError) no cond clause evaluated to a truthy value
    (explorer 0.6.0-dev) lib/explorer/polars_backend/series.ex:412: Explorer.PolarsBackend.Series.fill_missing_with_value/2

iex(2)> s2 = Explorer.Series.from_list([~N[2023-01-17 20:00:56.576456], ~N[2023-01-18 20:30:56.576456], nil, ~N[2023-01-09 21:00:56.576456]])
#Explorer.Series<
  Polars[4]
  datetime [2023-01-17 20:00:56.576456, 2023-01-18 20:30:56.576456, nil,
   2023-01-09 21:00:56.576456]
>

iex(3)> Explorer.Series.fill_missing(s2, ~N[2023-01-19 20:00:56.576456])
** (CondClauseError) no cond clause evaluated to a truthy value
    (explorer 0.6.0-dev) lib/explorer/polars_backend/series.ex:412: Explorer.PolarsBackend.Series.fill_missing_with_value/2
```

After this PR the result is:

```elixir
iex(3)> Explorer.Series.fill_missing(s1, ~D[2023-01-19])
#Explorer.Series<
  Polars[4]
  date [2023-01-17, 2023-01-18, 2023-01-19, 2023-01-09]
>

iex(4)> Explorer.Series.fill_missing(s2, ~N[2023-01-19 20:00:56.576456])
#Explorer.Series<
  Polars[4]
  datetime [2023-01-17 20:00:56.576456, 2023-01-18 20:30:56.576456, 2023-01-19 20:00:56.576456,
   2023-01-09 21:00:56.576456]
>
```

### DIFF
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -471,6 +471,20 @@ pub fn s_fill_missing_with_date(
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
+pub fn s_fill_missing_with_datetime(
+    data: ExSeries,
+    strategy: ExDateTime,
+) -> Result<ExSeries, ExplorerError> {
+    let s = data.clone_inner();
+    let s = s
+        .datetime()?
+        .fill_null_with_values(strategy.into())?
+        .cast(&DataType::Datetime(TimeUnit::Microseconds, None))?
+        .into_series();
+    Ok(ExSeries::new(s))
+}
+
+#[rustler::nif(schedule = "DirtyCpu")]
 pub fn s_fill_missing_with_boolean(
     data: ExSeries,
     strategy: bool,
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -476,7 +476,7 @@ pub fn s_fill_missing_with_datetime(
     let s = s
         .datetime()?
         .fill_null_with_values(datetime.into())?
-        .cast(&DataType::Datetime(TimeUnit::Microseconds, None))?
+        .cast(s.dtype())?
         .into_series();
     Ok(ExSeries::new(s))
 }
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -436,35 +436,32 @@ pub fn s_fill_missing_with_atom(data: ExSeries, atom: &str) -> Result<ExSeries, 
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn s_fill_missing_with_int(data: ExSeries, strategy: i64) -> Result<ExSeries, ExplorerError> {
+pub fn s_fill_missing_with_int(data: ExSeries, integer: i64) -> Result<ExSeries, ExplorerError> {
     let s = data.clone_inner();
-    let s = s.i64()?.fill_null_with_values(strategy)?.into_series();
+    let s = s.i64()?.fill_null_with_values(integer)?.into_series();
     Ok(ExSeries::new(s))
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn s_fill_missing_with_float(data: ExSeries, strategy: f64) -> Result<ExSeries, ExplorerError> {
+pub fn s_fill_missing_with_float(data: ExSeries, float: f64) -> Result<ExSeries, ExplorerError> {
     let s = data.clone_inner();
-    let s = s.f64()?.fill_null_with_values(strategy)?.into_series();
+    let s = s.f64()?.fill_null_with_values(float)?.into_series();
     Ok(ExSeries::new(s))
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn s_fill_missing_with_bin(data: ExSeries, strategy: &str) -> Result<ExSeries, ExplorerError> {
+pub fn s_fill_missing_with_bin(data: ExSeries, binary: &str) -> Result<ExSeries, ExplorerError> {
     let s = data.clone_inner();
-    let s = s.utf8()?.fill_null_with_values(strategy)?.into_series();
+    let s = s.utf8()?.fill_null_with_values(binary)?.into_series();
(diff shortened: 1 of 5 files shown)


native/explorer/src/series.rs
```diff
@@ -456,6 +456,34 @@ pub fn s_fill_missing_with_bin(data: ExSeries, strategy: &str) -> Result<ExSerie
     Ok(ExSeries::new(s))
 }
 
+#[rustler::nif(schedule = "DirtyCpu")]
+pub fn s_fill_missing_with_date(
+    data: ExSeries,
+    strategy: ExDate,
```
philss, line 462: It's nitpick, but I would name this "`date: ExDate`" instead of `strategy`.
The same idea for the `s_fill_missing_with_datetime`. WDYT?

philss (approved):
Just a note, but LGTM! :shipit: 

Thank you!

native/explorer/src/series.rs
Jhonatannunessilva: Yes, I agree, but other functions follow this naming pattern, such as: s_fill_missing_with_atom, s_fill_missing_with_int, s_fill_missing_with_float, s_fill_missing_with_bin, and s_fill_missing_with_boolean. WDYT?

philss: I'm more inclined to rename them to be more closed to the data type when it's possible. I guess where we are using `fill_null_with_values` we could rename. WDYT?

native/explorer/src/series.rs
```diff
@@ -456,6 +456,34 @@ pub fn s_fill_missing_with_bin(data: ExSeries, strategy: &str) -> Result<ExSerie
...
+#[rustler::nif(schedule = "DirtyCpu")]
+pub fn s_fill_missing_with_datetime(
+    data: ExSeries,
+    strategy: ExDateTime,
+) -> Result<ExSeries, ExplorerError> {
+    let s = data.clone_inner();
+    let s = s
+        .datetime()?
+        .fill_null_with_values(strategy.into())?
+        .cast(&DataType::Datetime(TimeUnit::Microseconds, None))?
```
josevalim, line 482: I don't think we should cast the series here. This will change the whole precision of the series and I am not sure if it is desired. I believe we need to get the precision from the series and convert the datetime to an integer with the same precision. Check encoding.rs, we have some functions that work around this. :)

Jhonatannunessilva: Do all datetime series have the same precision? (As I understand it, these two functions take the precision of the series: [term_from_value](https://github.com/elixir-nx/explorer/blob/main/native/explorer/src/encoding.rs#L389) and [list_from_series](https://github.com/elixir-nx/explorer/blob/main/native/explorer/src/encoding.rs#L403))

If yes, does this code resolve the datetime cast problem?:
```rust
pub fn s_fill_missing_with_datetime(
    data: ExSeries,
    datetime: ExDateTime,
) -> Result<ExSeries, ExplorerError> {
    let s = data.clone_inner();
    let s = s.datetime()?;
    let s = s
        .fill_null_with_values(datetime.into())?
        .cast(&DataType::Datetime(s.time_unit(), None))?
        .into_series();
    Ok(ExSeries::new(s))
}
```

If I understand correctly, if you remove the cast function, the result will already be an integer:
```rust
pub fn s_fill_missing_with_datetime(
    data: ExSeries,
    datetime: ExDateTime,
) -> Result<ExSeries, ExplorerError> {
    let s = data.clone_inner();
    let s = s
        .datetime()?
        .fill_null_with_values(datetime.into())?
        .into_series();
    Ok(ExSeries::new(s))
}
```
```elixir
iex(1)> s1 = Explorer.Series.from_list([~N[2023-01-17 20:00:56.576456], ~N[2023-01-18 20:30:56.576456], nil, ~N[2023-01-09 21:00:56.576456]])
#Explorer.Series<
  Polars[4]
  datetime [2023-01-17 20:00:56.576456, 2023-01-18 20:30:56.576456, nil,
   2023-01-09 21:00:56.576456]
>
iex(2)> Explorer.Series.fill_missing(s1, ~N[2023-01-19 20:00:56.576456])                    
#Explorer.Series<
  Polars[4]
  integer [1673985656576456, 1674073856576456, 1674158456576456,
   1673298056576456]
>
iex(3)> Explorer.Series.fill_missing(s1, ~N[2023-01-19 20:00:56.576456]) |> Explorer.Series.to_list()
[1673985656576456, 1674073856576456, 1674158456576456, 1673298056576456]
```

philss: > Do all values ​​in a datetime series have the same precision?

@Jhonatannunessilva I would say yes.

I think if we do as the first example you showed, it's going to be fine.
There is a slightly better improvement: you could take the entire `s.dtype()` to cast back

@josevalim for what I could understand, we need to cast back because `fill_null_with_values` is a function of the `ChunkedArray<Int64>`.

josevalim: Ah, perfect. So casting to s.dtype() sounds the like the right way to go to me! But, if `fill_null_with_values` is expecting an integer as input, then we also need to make sure that the DateTime is cast to the input with the right precision.

josevalim:

PR #485, opened 2023-01-24
Simplify release workflow with a new action

philss:
Since we now have https://github.com/philss/rustler-precompiled-action, it's possible to remove a lot of code from the current workflow.

This also have a small fix for the title of each job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,5 @@
 name: Build precompiled NIFs
 
-env:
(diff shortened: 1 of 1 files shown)


philss:
It's taking a while to build because of Rustler 0.27.0 using "release" mode for all environments. I'm sending a commit directly to the main in order to fix this. d579c30a5dc2ffe4ca3bfd84487f248c77573389

josevalim (approved):

PR #487, opened 2023-01-27
Support `time` dtype

sasikumar87:
1. I'm new to rust, pls help me to find the cause of the failing tests.
2. What existing Series functions can support time dtype ?


josevalim:
I would look at where Date and DateTime are supported. :) It is probably going to be a subset of that. :)

native/explorer/src/expressions.rs
```diff
@@ -57,6 +57,15 @@ pub fn expr_date(date: ExDate) -> ExExpr {
     ExExpr::new(expr)
 }
 
+#[rustler::nif]
+pub fn expr_time(_time: ExTime) -> ExExpr {
+    // FIXME: lit() is not available on NaiveTime
```
philss, line 62: I'm not sure if this is possible. We may need to ask Polars team to add (or we open a PR there) this.
See: https://pola-rs.github.io/polars/polars/prelude/trait.Literal.html

lib/explorer/polars_backend/expression.ex
```diff
@@ -173,6 +173,7 @@ defmodule Explorer.PolarsBackend.Expression do
   def to_expr(number) when is_integer(number), do: Native.expr_integer(number)
   def to_expr(number) when is_float(number), do: Native.expr_float(number)
   def to_expr(%Date{} = date), do: Native.expr_date(date)
+  def to_expr(%Time{} = date), do: Native.expr_time(date)
```
philss, line 176: I think we can remove this until we get support from Polars.

philss (commented):
I added two points related to the same thing.

I cannot understand yet whey it's getting an error.

But mostly the code looks good!

philss (approved):
LGTM! :shipit: 

The only thing I would suggest is to add more tests to the `series_test.exs` file. We are lacking some tests there, because in the past we rely only on doctests. But now we want to have a better test suite.

This could be in another PR, but would be cool if you could add tests to the functions you changed (like min/max/cummulative_min, etc).

josevalim:
Let's add the tests now so we build on the practice of adding more tests. Can you please do that @sasikumar87? thank you :heart:

sasikumar87:
@philss @josevalim sure. 

I've already added doctests for impacted functions. Should I remove and add them to `series_test.exs` or keep both?

josevalim:

philss:
@sasikumar87 please keep both.

josevalim:
Keep both please! I thought this was good to merge but the improved coverage will be welcome! :heart:

PR #490, opened 2023-01-30
Simplify `df_describe` with `as_deref()`

philss:
This makes the conversion of `Option<Vec<f64>>` to `Option<&[f64]>` in a cleaner manner.


cristineguadelupe:
Thanks @philss ✈️

josevalim (approved):

PR #492, opened 2023-02-01
Add LazyFrame concat of rows and columns

philss:
~~I'm not entirely sure about the `concat_columns` for LazyFrame, but the new option `check_n_rows` was the way I found to support that function. WDYT?~~ -> The validation was moved to the eager backend.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -4043,10 +4043,10 @@ defmodule Explorer.DataFrame do
   @doc """
   Combine two or more dataframes column-wise.
 
-  Dataframes must have the same number of rows.
-  Due to the fact that lazy dataframes can't have the number of rows calculated,
-  this restriction is not valid for them. But notice that some rows may be removed
-  if the sizes differ. This is because we use an "inner join" to concat the dataframes.
+  This function expects the dataframes to have the same number of rows,
+  otherwise rows may be silently discarded. Eager backends may check
+  whenever this happens and raise instead of silently fail. But this may not
+  be possible for lazy dataframes as the number of rows is not known upfront.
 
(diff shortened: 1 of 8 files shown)


lib/explorer/data_frame.ex
```diff
@@ -4048,6 +4048,19 @@ defmodule Explorer.DataFrame do
   When working with grouped dataframes, be aware that only groups from the first
   dataframe are kept in the resultant dataframe.
 
+  Note that for lazy frames it's not possible to check the number of rows, so
+  you need to pass the option `:check_n_rows` as `false`.
+
+  ## Options
+
+    * `:check_n_rows` - Controls if all dataframes must have the same number of rows.
```
josevalim, line 4056: One option is to move the logic for checking rows inside the backend. So eager will check it and lazy won't. Here we simply document that lazy backends won't check the number of rows and it is up to you to guarantee they match if you don't want data loss. WDYT?

lib/explorer/data_frame.ex
```diff
@@ -4044,6 +4044,9 @@ defmodule Explorer.DataFrame do
   Combine two or more dataframes column-wise.
 
   Dataframes must have the same number of rows.
+  Due to the fact that lazy dataframes can't have the number of rows calculated,
+  this restriction is not valid for them. But notice that some rows may be removed
+  if the sizes differ. This is because we use an "inner join" to concat the dataframes.
```
josevalim, line 4049:
```suggestion
  This function expects the dataframes to have the same number of rows,
  otherwise rows may be silently discarded. Eager backends may check
  whenever this happens and raise instead of silently fail. But this may not
  be possible for lazy dataframes as the number of rows is not known upfront.
```

josevalim (approved):

PR #496, opened 2023-02-06
Add LazyFrame version of `to_parquet/3`

philss:
This is going to save a lazy frame to a parquet file using a stream.
This should keep the memory usage low for bigger files.

### DIFF
--- a/lib/explorer/polars_backend/lazy_frame.ex
+++ b/lib/explorer/polars_backend/lazy_frame.ex
@@ -208,6 +208,19 @@ defmodule Explorer.PolarsBackend.LazyFrame do
   end
 
   @impl true
+  def to_parquet(%DF{groups: []} = df, filename, {compression, level}) do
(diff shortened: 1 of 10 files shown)


lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -207,6 +207,19 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
+  def to_parquet(%DF{groups: []} = df, filename, {compression, level}) do
+    case Native.lf_to_parquet(df.data, filename, Shared.parquet_compression(compression, level)) do
+      {:ok, _} -> :ok
+      {:error, _} = err -> err
+    end
+  end
+
+  @impl true
+  def to_parquet(_df, _filename, _compression) do
+    raise "to_parquet/3 with groups is not supported yet for lazy frames"
```
josevalim, line 220: Do groups matter on these operations? I think the group is just ignored, no?

josevalim (approved):

PR #497, opened 2023-02-07
Raise for unavailable LazyFrame functions explicitly

philss:
This if new functions are added to the behaviour, we should know easily.

### DIFF
--- a/lib/explorer/polars_backend/lazy_frame.ex
+++ b/lib/explorer/polars_backend/lazy_frame.ex
@@ -353,19 +353,36 @@ defmodule Explorer.PolarsBackend.LazyFrame do
     Shared.apply_dataframe(head, out_df, :lf_concat_columns, [Enum.map(tail, & &1.data)])
   end
 
-  # TODO: Make the functions of non-implemented functions
-  # explicit once the lazy interface is ready.
-  funs =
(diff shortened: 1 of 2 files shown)


lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -353,19 +353,36 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
+  ]
+
+  for {fun, arity} <- not_available_funs do
     args = Macro.generate_arguments(arity, __MODULE__)
 
     @impl true
     def unquote(fun)(unquote_splicing(args)) do
-      raise "cannot perform operation on an Explorer.PolarsBackend.LazyFrame"
+      raise "the function `#{unquote(fun)}/#{unquote(arity)}` is not available for the Explorer.PolarsBackend.LazyFrame backend. " <>
+              "Please use Explorer.DataFrame.collect/1 and then call this function upon the resultant dataframe."
```
josevalim, line 385: nitpick: error messages don't end with punctuation

```suggestion
              "Please use Explorer.DataFrame.collect/1 and then call this function upon the resultant dataframe"
```

josevalim (approved):

PR #500, opened 2023-02-08
Fix pow of (series, series) when there is nil

philss:
This prevents an error to raise.

### DIFF
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -785,7 +785,7 @@ pub fn s_pow(data: ExSeries, other: ExSeries) -> Result<ExSeries, ExplorerError>
 
             let s = iter1
                 .zip(iter2)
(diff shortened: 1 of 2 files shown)


native/explorer/src/series.rs
```diff
@@ -785,7 +785,10 @@ pub fn s_pow(data: ExSeries, other: ExSeries) -> Result<ExSeries, ExplorerError>
 
             let s = iter1
                 .zip(iter2)
-                .map(|(v1, v2)| v1.unwrap().pow(v2.unwrap()))
+                .map(|(v1, v2)| {
+                    v1.map(|left| v2.map(|right| left.pow(right)))
+                        .unwrap_or_default()
```
philss, line 790: I had to use `unwrap_or_default()` because this is going to in `Option<Option<i64>>`. So unwrap here is just removing one level of Option.

josevalim (approved):

PR #505, opened 2023-02-14
fixed typo and clarified sentence for macro note

a-alhusaini:
I noticed that the macro note stated with some of the functions like `filter` and `mutate` states that "this is macro" and the correct way to state that would be "this is a macro"

I shortened the sentence too. "Therefore" was redundant.


philss:
@a-alhusaini thank you! :purple_heart:

PR #510, opened 2023-02-17
add max_rows and columns options to from_parquet

mlineen:
Add `max_rows` and `columns` options to `DataFrame.from_parquet/2` -- inspired by the options available in `DataFrame.from_csv/2` and in [polars.read_parquet](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.read_parquet.html) and [ParquetReader](https://pola-rs.github.io/polars/polars/prelude/struct.ParquetReader.html#)

**TODO:** I could use guidance on if/how this should be applied to the lazy backend.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -536,13 +536,30 @@ defmodule Explorer.DataFrame do
 
   @doc """
   Reads a parquet file into a dataframe.
+
+  ## Options
+    * `:max_rows` - Maximum number of lines to read. (default: `nil`)
+    * `:columns` - A list of column names or indexes to keep. If present, only these columns are read into the dataframe. (default: `nil`)
   """
   @doc type: :io
   @spec from_parquet(filename :: String.t(), opts :: Keyword.t()) ::
           {:ok, DataFrame.t()} | {:error, term()}
   def from_parquet(filename, opts \\ []) do
-    backend = backend_from_options!(opts)
-    backend.from_parquet(filename)
+    {backend_opts, opts} = Keyword.split(opts, [:backend, :lazy])
+
+    opts =
+      Keyword.validate!(opts,
+        max_rows: nil,
+        columns: nil
+      )
+
+    backend = backend_from_options!(backend_opts)
+
+    backend.from_parquet(
+      filename,
+      opts[:max_rows],
+      to_columns_for_io(opts[:columns])
+    )
   end
 
   @doc """
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -538,8 +538,11 @@ defmodule Explorer.DataFrame do
   Reads a parquet file into a dataframe.
 
   ## Options
+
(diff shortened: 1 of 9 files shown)


philss:
Hi @mlineen :wave: 

Thank you for the PR!
Regarding the LazyFrame backend, you can only control the `max_row` option by setting the `n_row` field of [`ScanArgsParquet`](https://docs.rs/polars-lazy/0.26.1/polars_lazy/frame/struct.ScanArgsParquet.html) that is the second argument of `scan_parquet`. (you need to change here: https://github.com/elixir-nx/explorer/blob/d7148c9be7c0fe5d8d01de89cac8baae368af487/native/explorer/src/lazyframe/io.rs#L10)

I think we can raise if the person uses the `columns` option for the LazyFrame. WDYT?

mlineen:
> I think we can raise if the person uses the `columns` option for the LazyFrame. WDYT?

I think that sounds okay; does [`select/2`](https://elixir-nx.github.io/explorer/Explorer.DataFrame.html#select/2) ([rust](https://pola-rs.github.io/polars/polars/prelude/struct.LazyFrame.html#method.select) and [python](https://pola-rs.github.io/polars/py-polars/html/reference/lazyframe/api/polars.LazyFrame.select.html#polars.LazyFrame.select)) get applied through the lazy backend and only fetch the columns of interest on huge dataframes? The whole idea would be to efficiently load only a subset of the data.

philss:
> does [select/2](https://elixir-nx.github.io/explorer/Explorer.DataFrame.html#select/2) ([rust](https://pola-rs.github.io/polars/polars/prelude/struct.LazyFrame.html#method.select) and [python](https://pola-rs.github.io/polars/py-polars/html/reference/lazyframe/api/polars.LazyFrame.select.html#polars.LazyFrame.select)) get applied through the lazy backend and only fetch the columns of interest on huge dataframes?

I think so, but I'm not sure if this is propagated to the scan.

philss (approved):
LGTM! :shipit:

lib/explorer/data_frame.ex
```diff
@@ -536,13 +536,30 @@ defmodule Explorer.DataFrame do
 
   @doc """
   Reads a parquet file into a dataframe.
+
+  ## Options
```
josevalim, line 540:
```suggestion
  ## Options

```

lib/explorer/data_frame.ex
```diff
@@ -536,13 +536,30 @@ defmodule Explorer.DataFrame do
...
+    * `:max_rows` - Maximum number of lines to read. (default: `nil`)
+    * `:columns` - A list of column names or indexes to keep. If present, only these columns are read into the dataframe. (default: `nil`)
```
josevalim, line 542:
```suggestion

    * `:columns` - A list of column names or indexes to keep. If present,
      only these columns are read into the dataframe. (default: `nil`)
```

josevalim:

Issue #467, opened 2023-01-11
Consistenly allow :nan, :neg_infinity, and :infinity


josevalim:
They should be allowed everywhere regular Elixir values are allowed, such as:

1. [x] `Series.from_list([:nan, 1.0, :infinity])` (#495)
2. [x] `Series.eq(series, :nan)` and similar
3. [x] `Series.fill_missing(series, :infinity)` (note `:nan` is already allowed here) (#484)

philss:
I believe all of them are covered now. Thank you very much, @Jhonatannunessilva ! :smiley:  :purple_heart:

PR #512, opened 2023-02-20
Allows Series functions to receive nan, infinity, and neg_infinity values

Jhonatannunessilva:
This PR allows Series functions to receive nan, infinity, and neg_infinity values.

This is part of #467

Functions that have been tested and are working:

- [X] Series.equal/2
- [X] Series.not_equal/2
- [X] Series.greater/2
- [X] Series.greater_equal/2
- [X] Series.less/2
- [X] Series.less_equal/2
- [X] Series.in/2
- [X] Series.iotype/2
- [X] Series.sort/2
- [X] Series.argsort/2
- [X] Series.add/2
- [X] Series.subtract/2
- [X] Series.multiply/2
- [X] Series.divide/2
- [X] Series.pow(series, :nan | :infinity | :neg_infinity)
- [X] Series.pow(:nan | :infinity | :neg_infinity, series)
- [ ] Series.pow(float_series_1, float_series_2)

NOTES:

- The Series.pow/2 function only accepts integer series. Is it to allow float series as well?
- In Rust, f64::NAN is not equal to f64::NAN
  - So when Series.equal/2 is called:

    ```elixir
    iex(1)> s1 = Explorer.Series.from_list([1.0, 2.5, :nan, :infinity, :neg_infinity])
    #Explorer.Series<
    Polars[5]
    float [1.0, 2.5, NaN, Inf, -Inf]
    >

    iex(2)> s2 = Explorer.Series.from_list([1.0, 3.0, :nan, :infinity, :neg_infinity])
    #Explorer.Series<
    Polars[5]
    float [1.0, 3.0, NaN, Inf, -Inf]
    >

    iex(3)> Explorer.Series.equal(s1, s2)
    #Explorer.Series<
    Polars[5]
    boolean [true, false, false, true, true]
    >
    ```

  - But with the Series.in/2 function the behavior is different:

    ```elixir
    iex(1)> s1 = Explorer.Series.from_list([1.0, 2.5, :nan, :infinity, :neg_infinity])
    #Explorer.Series<
    Polars[5]
    float [1.0, 2.5, NaN, Inf, -Inf]
    >

    iex(2)> s2 = Explorer.Series.from_list([1.0, 3.0, :nan, :infinity, :neg_infinity])
    #Explorer.Series<
    Polars[5]
    float [1.0, 3.0, NaN, Inf, -Inf]
    >

    iex(3)> Explorer.Series.in(s1, s2)   
    #Explorer.Series<
    Polars[5]
    boolean [true, false, true, true, true]
    >
    ```

### DIFF
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -203,8 +203,11 @@ defmodule Explorer.PolarsBackend.Series do
   def subtract(%Series{} = left, %Series{} = right),
     do: Shared.apply_series(left, :s_subtract, [right.data])
 
-  def subtract(left, right) when is_number(right), do: apply_scalar_on_rhs(:subtract, left, right)
-  def subtract(left, right) when is_number(left), do: apply_scalar_on_lhs(:subtract, left, right)
+  def subtract(left, right) when is_number(right) or right in [:nan, :infinity, :neg_infinity],
+    do: apply_scalar_on_rhs(:subtract, left, right)
+
+  def subtract(left, right) when is_number(left) or left in [:nan, :infinity, :neg_infinity],
+    do: apply_scalar_on_lhs(:subtract, left, right)
 
   @impl true
   def multiply(%Series{} = left, %Series{} = right),
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -252,7 +252,7 @@ defmodule Explorer.PolarsBackend.Series do
   def pow(%Series{} = left, %Series{} = right),
     do: Shared.apply_series(left, :s_pow, [right.data])
 
-  def pow(left, exponent) when is_float(exponent),
+  def pow(left, exponent) when is_float(exponent) or exponent in [:nan, :infinity, :neg_infinity],
     do: Shared.apply_series(left, :s_pow_f_rhs, [exponent])
 
   def pow(left, exponent) when is_integer(exponent) and exponent >= 0 do
@@ -262,8 +262,9 @@ defmodule Explorer.PolarsBackend.Series do
     end
   end
 
-  def pow(exponent, right) when is_float(exponent),
-    do: Shared.apply_series(right, :s_pow_f_lhs, [exponent])
+  def pow(exponent, right)
+      when is_float(exponent) or exponent in [:nan, :infinity, :neg_infinity],
+      do: Shared.apply_series(right, :s_pow_f_lhs, [exponent])
 
   def pow(exponent, right) when is_integer(exponent) and exponent >= 0 do
     cond do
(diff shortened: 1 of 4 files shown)


lib/explorer/polars_backend/series.ex
```diff
@@ -193,29 +193,41 @@ defmodule Explorer.PolarsBackend.Series do
   def add(%Series{} = left, %Series{} = right),
     do: Shared.apply_series(left, :s_add, [right.data])
 
-  def add(left, right) when is_number(right), do: apply_scalar_on_rhs(:add, left, right)
-  def add(left, right) when is_number(left), do: apply_scalar_on_lhs(:add, left, right)
+  def add(left, right) when is_number(right) or right in [:nan, :infinity, :neg_infinity],
```
josevalim, line 196: What do you think about using `defguardp` to define this? I just can't think of a good name :D

    defguardp is_numerical(n) when is_number(n) or n in [:nan, :infinity, :neg_infinity]

Jhonatannunessilva: I agree. I thought of the name is_number?/1, but it might be confused with Elixir's is_number/1. So I think the name is_numerical/1 is probably better.

WDYT?

josevalim: perfect!

Jhonatannunessilva: Should I create this guard in explorer/series.ex or polars_backend/series.ex? (It can be used in [basic_numeric_operation/3](https://github.com/elixir-nx/explorer/pull/512/commits/de1144a7c35b3a6d78f05f78aeb09203c271a53a#diff-2449d4b4a415673b77c0ab0ba4add0310f4a28d33c3e9a9b06f05cd529166436R2214) and [valid_for_bool_mask_operation?/2](https://github.com/elixir-nx/explorer/pull/512/commits/50fa195fe9ee88cbfe76f519c898258b62e7d2cd#diff-2449d4b4a415673b77c0ab0ba4add0310f4a28d33c3e9a9b06f05cd529166436R2558) functions)

josevalim: Explorer.Series. If you want to share, you can create it Utils, but it is small enough that should be fine to duplicate. We should also rename `numeric_dtype?` to `is_numeric_dtype` since guards have the `is_` prefix instead of question marks.

Jhonatannunessilva: So should I rename the [io_dtype?/1](https://github.com/elixir-nx/explorer/blob/main/lib/explorer/series.ex#L81) and [numeric_or_bool_dtype?/1](https://github.com/elixir-nx/explorer/blob/main/lib/explorer/series.ex#L83) guards as well? (is_io_dtype/1 and is_numeric_or_bool_dtype/1)

josevalim: ideally yes, but since there are several, let's do that in another commit to keep this one focused. :)

josevalim:
> The Series.pow/2 function only accepts integer series. Is it to allow float series as well?

Yes, we should support a series of floats too!

josevalim:

josevalim:
I merged this for now so we can unblock work on the upcoming PRs. :)

PR #517, opened 2023-02-23
Add `DF.shuffle/2` and `Series.shuffle/2`

### LINKED ISSUE #506: Sample always return the same dataframe when passing `1.0`
The [`Explorer.DataFrame.sample/3`](https://hexdocs.pm/explorer/Explorer.DataFrame.html#sample/3) function should shuffle the dataframe instead.

philss:
This PR also slightly changes the options of `DF.sample/3` and `Series.sample/3`, introducing
the `:shuffle` option.

We changed the implementation of `Series.sample/3` to use the built-in Polars functions instead of our custom
function. This changed the results of our tests and examples, and should affect users that are using seeds.

Closes #506


lib/explorer/data_frame.ex
```diff
@@ -3395,20 +3398,62 @@ defmodule Explorer.DataFrame do
...
+      >
+
+  """
+  @doc type: :rows
+  @spec shuffle(df :: DataFrame.t(), opts :: Keyword.t()) :: DataFrame.t()
+  def shuffle(df, opts \\ [])
+
+  def shuffle(df, opts) do
+    opts = Keyword.validate!(opts, seed: nil)
+    seed = opts[:seed] || Enum.random(1..1_000_000_000_000)
```
josevalim, line 3454: Shouldn't we just pass it as a nil and let Polars generate a default one?

lib/explorer/series.ex
```diff
@@ -1085,69 +1085,117 @@ defmodule Explorer.Series do
...
+      >
+
+  """
+  @doc type: :shape
+  @spec shuffle(series :: Series.t(), opts :: Keyword.t()) :: Series.t()
+  def shuffle(series, opts \\ [])
+
+  def shuffle(series, opts) do
+    opts = Keyword.validate!(opts, seed: nil)
+    seed = opts[:seed] || Enum.random(1..1_000_000_000_000)
```
josevalim, line 1194: Same here. :)

josevalim (approved):
Beautiful! I have dropped one comment but, in case we have to set the seed upfront, then this looks good to me!

cigrainger (approved):
LGTM! :shipit:

PR #518, opened 2023-02-23
Add `:name` field to `Series` struct

philss:
And fill it when "pull" the series from the dataframe, and when using `get_and_update`.

This is related to https://github.com/elixir-nx/explorer/issues/507

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -3062,7 +3062,7 @@ defmodule Explorer.DataFrame do
   Extracts a single column as a series.
 
   This is equivalent to `df[field]` for retrieving a single field.
-  The resultant series is going to have the column name in the `:name` field.
+  The returned series will have its `:name` field set to the column name.
 
   ## Examples
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -368,7 +368,7 @@ defmodule Explorer.DataFrame do
 
     new_df = put(df, column, new_value)
 
-    {current_value, new_df}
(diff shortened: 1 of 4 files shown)


josevalim (approved):
Only documentation is missing. We can add some docs on top of defstruct saying that both `name` and `dtype` are “public”. And let’s also mention that will be nil unless access through the dataframe!

lib/explorer/series.ex
```diff
@@ -72,8 +72,10 @@ defmodule Explorer.Series do
   @type t :: %Series{data: Explorer.Backend.Series.t(), dtype: dtype()}
   @type lazy_t :: %Series{data: Explorer.Backend.LazySeries.t(), dtype: dtype()}
 
+  # The fields `:dtype` and `:name` are public. But `:name` is always `nil` unless
+  # the series is accessed through the dataframe.
```
josevalim, line 76:
```suggestion
  @doc """
  The Series struct.

  The fields `:dtype` and `:name` are public. `:name` is always `nil` unless
  the series is retrieved from a dataframe.
  """
```

lib/explorer/data_frame.ex
```diff
@@ -3060,6 +3062,7 @@ defmodule Explorer.DataFrame do
   Extracts a single column as a series.
 
   This is equivalent to `df[field]` for retrieving a single field.
+  The resultant series is going to have the column name in the `:name` field.
```
josevalim, line 3065:
```suggestion
  The returned series will have its `:name` field set to the column name.
```

josevalim (approved):

PR #520, opened 2023-02-24
Add Series.format/1 function

Jhonatannunessilva:
This PR resolves https://github.com/elixir-nx/explorer/issues/297.

Thanks for the help @philss!

After this PR:

```elixir
iex> s1 = Explorer.Series.from_list(["a", "b", "c"])
iex> s2 = Explorer.Series.from_list(["d", "e", "f"])
iex> s3 = Explorer.Series.from_list(["g", "h", "i"])
iex> Explorer.Series.format([s1, s2, s3])
#Explorer.Series<
  Polars[3]
  string ["adg", "beh", "cfi"]
>

iex> s1 = Explorer.Series.from_list(["a", "b", "c", "d"])
iex> s2 = Explorer.Series.from_list([1, 2, 3, 4])
iex> s3 = Explorer.Series.from_list([1.5, :nan, :infinity, :neg_infinity])
iex> Explorer.Series.format([s1, "/", s2, "/", s3])
#Explorer.Series<
  Polars[4]
  string ["a/1/1.5", "b/2/NaN", "c/3/inf", "d/4/-inf"]
>

iex> s1 = Explorer.Series.from_list([<<1>>, <<239, 191, 19>>], dtype: :binary)
iex> s2 = Explorer.Series.from_list([<<3>>, <<4>>], dtype: :binary)
iex> Explorer.Series.format([s1, s2])
** (RuntimeError) External error: invalid utf-8 sequence

# With DataFrame:

iex> df = Explorer.DataFrame.new([a: ["a", "b", "c"], b: ["d", "e", "f"]])                                
#Explorer.DataFrame<
  Polars[3 x 2]
  a string ["a", "b", "c"]
  b string ["d", "e", "f"]
>

iex> Explorer.DataFrame.mutate_with(df, fn df -> [c: Explorer.Series.format([df["a"], "/", df["b"]])] end)
#Explorer.DataFrame<
  Polars[3 x 3]
  a string ["a", "b", "c"]
  b string ["d", "e", "f"]
  c string ["a/d", "b/e", "c/f"]
>

iex> ldf = Explorer.DataFrame.new([a: ["a", "b", "c"], b: ["d", "e", "f"]], lazy: true)                   
#Explorer.DataFrame<
  LazyPolars[??? x 2]
  a string ["a", "b", "c"]
  b string ["d", "e", "f"]
>

iex> Explorer.DataFrame.mutate_with(ldf, fn ldf -> [c: Explorer.Series.format([ldf["a"], "/", ldf["b"]])] end)
#Explorer.DataFrame<
  LazyPolars[??? x 3]
  a string ["a", "b", "c"]
  b string ["d", "e", "f"]
  c string ["a/d", "b/e", "c/f"]
>
```

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -39,6 +39,7 @@ defmodule Explorer.Backend.LazySeries do
     pow: 2,
     fill_missing_with_value: 2,
     fill_missing_with_strategy: 2,
+    format: 2,
     concat: 2,
     coalesce: 2,
     cast: 2,
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -39,7 +39,7 @@ defmodule Explorer.Backend.LazySeries do
     pow: 2,
     fill_missing_with_value: 2,
     fill_missing_with_strategy: 2,
-    format: 2,
+    format: 1,
     concat: 1,
     coalesce: 2,
     cast: 2,
@@ -408,6 +408,14 @@ defmodule Explorer.Backend.LazySeries do
       end
 
     Backend.Series.new(data, dtype)
+  end
+
+  @impl true
+  def format(list) do
+    series_list = Enum.map(list, &series_or_lazy_series!/1)
+    data = new(:format, [series_list], aggregations?(series_list))
+
+    Backend.Series.new(data, :string)
   end
 
   @impl true
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -118,6 +118,9 @@ defmodule Explorer.PolarsBackend.Series do
   def at(series, idx), do: Shared.apply_series(series, :s_at, [idx])
 
   @impl true
+  def format(s1, s2), do: Shared.apply_series(matching_size!(s1, s2), :s_format, [s2.data])
+
+  @impl true
   def concat(s1, s2), do: Shared.apply_series(s1, :s_concat, [s2.data])
 
   @impl true
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -123,7 +123,7 @@ defmodule Explorer.PolarsBackend.Series do
 
     case Explorer.PolarsBackend.Native.s_format(polars_series) do
       {:ok, %__MODULE__{} = new_series} -> Shared.create_series(new_series)
-      {:error, error} -> raise "Cannot concat series with Polars. Reason: #{inspect(error)}"
+      {:error, error} -> raise "cannot format series with Polars, reason: #{inspect(error)}"
     end
   end
 
@@ -133,7 +133,7 @@ defmodule Explorer.PolarsBackend.Series do
 
     case Explorer.PolarsBackend.Native.s_concat(polars_series) do
       {:ok, %__MODULE__{} = new_series} -> Shared.create_series(new_series)
-      {:error, error} -> raise "Cannot concat series with Polars. Reason: #{inspect(error)}"
+      {:error, error} -> raise "cannot concat series with Polars, reason: #{inspect(error)}"
     end
   end
 
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -118,7 +118,14 @@ defmodule Explorer.PolarsBackend.Series do
   def at(series, idx), do: Shared.apply_series(series, :s_at, [idx])
 
   @impl true
-  def format(s1, s2), do: Shared.apply_series(matching_size!(s1, s2), :s_format, [s2.data])
+  def format(list) do
+    polars_series = for s <- list, do: s.data
+
+    case Explorer.PolarsBackend.Native.s_format(polars_series) do
+      {:ok, %__MODULE__{} = new_series} -> Shared.create_series(new_series)
+      {:error, error} -> raise "Cannot concat series with Polars. Reason: #{inspect(error)}"
+    end
+  end
 
   @impl true
   def concat([%Series{} | _] = series) do
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1401,22 +1401,15 @@ defmodule Explorer.Series do
   @spec format([
           Series.t() | number() | Date.t() | Time.t() | NaiveDateTime.t() | boolean() | String.t()
         ]) :: Series.t()
-  def format([%Series{} = h | t] = _list), do: Enum.reduce(t, h, &format_reducer/2)
+  def format([_ | _] = list), do: Shared.apply_series_impl(:format, [cast_to_string(list)])
 
-  def format([h | t] = _list),
-    do: Enum.reduce(t, from_list([h], dtype: :string), &format_reducer/2)
-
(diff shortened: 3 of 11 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -39,6 +39,7 @@ defmodule Explorer.Backend.LazySeries do
     pow: 2,
     fill_missing_with_value: 2,
     fill_missing_with_strategy: 2,
+    format: 2,
```
josevalim, line 42: You still need to implement the callback in this module. I think you can look at `concat`, the implementation will be similar.

lib/explorer/series.ex
```diff
@@ -1369,6 +1369,55 @@ defmodule Explorer.Series do
...
+      ** (RuntimeError) External error: invalid utf-8 sequence
+  """
+  @doc type: :shape
+  @spec format([
+          Series.t() | number() | Date.t() | Time.t() | NaiveDateTime.t() | boolean() | String.t()
+        ]) :: Series.t()
+  def format([%Series{} = h | t] = _list), do: Enum.reduce(t, h, &format_reducer/2)
+
+  def format([h | t] = _list),
+    do: Enum.reduce(t, from_list([h], dtype: :string), &format_reducer/2)
```
josevalim, line 1407: Instead of formatting one by one, I think you can pass the whole list at once. The implementation will be something like:

```elixir
def format([_ | _] = list) do
  Shared.apply_series_impl(:format, cast_to_string(list))
end
```

where cast_to_string will call `from_list` on a string and cast non-string types. Then you will do the reduce in Rust (or in the Polars backend!). In other words, the callback will be `format/1`.

lib/explorer/backend/lazy_series.ex
josevalim: Actually, `concat/2` is not a good example, because it does it two by two, we likely want to pass the whole list to Rust at once (it may be able to better optimize it). I will open up a separate issue for that.

lib/explorer/series.ex
Jhonatannunessilva: Yes, it makes sense! (https://github.com/elixir-nx/explorer/pull/520/commits/42c429f6e8b07483df881afdb87f807b84c4e7e0)

lib/explorer/backend/lazy_series.ex
Jhonatannunessilva: https://github.com/elixir-nx/explorer/pull/520/commits/d6f8564791942ee0db11584718e51e47add63916

lib/explorer/polars_backend/series.ex
```diff
@@ -117,6 +117,16 @@ defmodule Explorer.PolarsBackend.Series do
   @impl true
   def at(series, idx), do: Shared.apply_series(series, :s_at, [idx])
 
+  @impl true
+  def format(list) do
+    polars_series = for s <- list, do: s.data
+
+    case Explorer.PolarsBackend.Native.s_format(polars_series) do
+      {:ok, %__MODULE__{} = new_series} -> Shared.create_series(new_series)
+      {:error, error} -> raise "Cannot concat series with Polars. Reason: #{inspect(error)}"
```
philss, line 126:
```suggestion
      {:error, error} -> raise "Cannot format series with Polars. Reason: #{inspect(error)}"
```

native/explorer/src/series.rs
```diff
@@ -161,6 +161,16 @@ pub fn s_slice(series: ExSeries, offset: i64, length: usize) -> Result<ExSeries,
     Ok(ExSeries::new(series.slice(offset, length)))
 }
 
+#[rustler::nif(schedule = "DirtyCpu")]
+pub fn s_format(list: Vec<ExSeries>) -> Result<ExSeries, ExplorerError> {
+    Ok(list
+        .into_iter()
+        .reduce(|acc, s2| {
+            ExSeries::new(acc.utf8().unwrap().concat(s2.utf8().unwrap()).into_series())
```
philss, line 169: I think we could avoid the additional allocations if we first map all the chunked arrays (the `series.utf8()` part), and then `concat` in the reduce, and after that we create one series in the end. The way you did we are creating intermediate `ExSeries` and `Series` that are going to be dropped for most of the operations.

Maybe it's just cheaper to go with the approach from `s_concat`, that uses a `for` loop. That way you can also rely on the `?` to return errors, instead of using `unwrap()`. WDYT?

philss (commented):
Awesome! :rocket: 

Just a suggestion added :)

lib/explorer/polars_backend/series.ex
josevalim: Messages start with lowercase. We should also fix this in concat:

```suggestion
      {:error, error} -> raise "cannot format series with Polars, reason: #{inspect(error)}"
```

lib/explorer/series.ex
```diff
@@ -1369,6 +1369,48 @@ defmodule Explorer.Series do
...
+      >
+
+      iex> s1 = Explorer.Series.from_list([<<1>>, <<239, 191, 19>>], dtype: :binary)
+      iex> s2 = Explorer.Series.from_list([<<3>>, <<4>>], dtype: :binary)
+      iex> Explorer.Series.format([s1, s2])
+      ** (RuntimeError) External error: invalid utf-8 sequence
+  """
+  @doc type: :shape
+  @spec format([
+          Series.t() | number() | Date.t() | Time.t() | NaiveDateTime.t() | boolean() | String.t()
```
josevalim, line 1402:
```suggestion
          Series.t() | String.t()
```

lib/explorer/series.ex
```diff
@@ -1369,6 +1369,48 @@ defmodule Explorer.Series do
...
+        ]) :: Series.t()
+  def format([_ | _] = list), do: Shared.apply_series_impl(:format, [cast_to_string(list)])
+
+  defp cast_to_string(list) do
+    Enum.map(list, fn
+      %Series{dtype: :string} = s -> s
+      %Series{} = s -> cast(s, :string)
+      value -> from_list([value], dtype: :string)
```
josevalim, line 1410:
```suggestion
      value when is_binary(value) -> from_list([value], dtype: :string)
      other -> raise ArgumentError, "format/1 expects a list of series or strings, got: #{inspect(other)}"
```

josevalim: I don't think we need to allow all data types. Only series and strings, that should simplify/remove some of the tests too!

Jhonatannunessilva: https://github.com/elixir-nx/explorer/pull/520/commits/6226df4589db02f568930592b34d12c3424b88a4

lib/explorer/series.ex
Jhonatannunessilva: https://github.com/elixir-nx/explorer/pull/520/commits/6226df4589db02f568930592b34d12c3424b88a4

native/explorer/src/series.rs
Jhonatannunessilva: https://github.com/elixir-nx/explorer/pull/520/commits/8655649cbe8ec80e84c82d3397157a387f9c953a

lib/explorer/polars_backend/series.ex
Jhonatannunessilva: https://github.com/elixir-nx/explorer/pull/520/commits/90b72f3471805a3ec5029c8adf90c79e62781c90

philss (approved):

josevalim:

PR #525, opened 2023-02-28
Add support for a list of filters in DF.filter/2

philss:
It is going to convert the list to a single "AND" expression.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1936,7 +1936,7 @@ defmodule Explorer.DataFrame do
   Picks rows based on `Explorer.Query`.
 
   The query is compiled and runs efficiently against the dataframe.
-  The query must return a boolean expression.
+  The query must return a boolean expression or a list of boolean expressions.
 
   > #### Notice {: .notice}
   >
@@ -1966,6 +1966,14 @@ defmodule Explorer.DataFrame do
         Polars[1 x 2]
         col1 string ["b"]
         col2 integer [2]
+      >
+
+      iex> df = Explorer.DataFrame.new(col1: [5, 4, 3], col2: [1, 2, 3])
+      iex> Explorer.DataFrame.filter(df, [col1 > 3, col2 < 3])
+      #Explorer.DataFrame<
+        Polars[2 x 2]
+        col1 integer [5, 4]
+        col2 integer [1, 2]
       >
 
   Returning a non-boolean expression errors:
@@ -2017,7 +2025,7 @@ defmodule Explorer.DataFrame do
   Picks rows based on a callback function.
 
   The callback receives a lazy dataframe. A lazy dataframe does
-  hold any values, instead it stores all operations in order to
+  not hold any values, instead it stores all operations in order to
   execute all filtering performantly.
 
   This is a callback version of `filter/2`.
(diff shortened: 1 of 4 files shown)


lib/explorer/data_frame.ex
```diff
@@ -1936,7 +1936,7 @@ defmodule Explorer.DataFrame do
   Picks rows based on `Explorer.Query`.
 
   The query is compiled and runs efficiently against the dataframe.
-  The query must return a boolean expression.
+  The query must return a boolean expression or a list of boolean expressions.
```
josevalim, line 1939:
```suggestion
  The query must return a boolean expression or a list of boolean expressions.
  When a list is returned, they are joined as `and` expressions.
```

lib/explorer/data_frame.ex
```diff
@@ -2078,6 +2086,24 @@ defmodule Explorer.DataFrame do
...
+            Enum.reduce(lazy_series, fn x, acc ->
+              Explorer.Backend.LazySeries.binary_and(acc, x)
+            end)
+
+          Shared.apply_impl(df, :filter_with, [df, series.data])
+        else
+          raise ArgumentError,
+                "expecting the function to return a single or a list of boolean LazySeries, " <>
+                  "but instead it contains:\n#{inspect(first_non_boolean)}\n" <>
+                  "that is of dtype #{inspect(first_non_boolean.dtype)}"
```
josevalim, line 2105: I would remove the rest because if may not be a series at all?

```suggestion
                  "but instead it contains:\n#{inspect(first_non_boolean)}"
```

lib/explorer/data_frame.ex
```diff
@@ -2078,6 +2086,24 @@ defmodule Explorer.DataFrame do
...
+        end
+
       other ->
         raise ArgumentError,
               "expecting the function to return a LazySeries, but instead it returned " <>
```
josevalim, line 2110: We need to update this message too? Perhaps we should encapsulate the error message in a function?

lib/explorer/data_frame.ex
```diff
@@ -2078,6 +2086,24 @@ defmodule Explorer.DataFrame do
...
 
+      [%Series{dtype: :boolean, data: %LazySeries{}} | _rest] = lazy_series ->
+        first_non_boolean =
+          Enum.find(lazy_series, &(!match?(%Series{dtype: :boolean, data: %LazySeries{}}, &1)))
+
+        if is_nil(first_non_boolean) do
+          series =
+            Enum.reduce(lazy_series, fn x, acc ->
+              Explorer.Backend.LazySeries.binary_and(acc, x)
+            end)
```
josevalim, line 2098:
```suggestion
          series = Enum.reduce(lazy_series, &Explorer.Backend.LazySeries.binary_and/2)
```

lib/explorer/series.ex
```diff
@@ -2299,6 +2299,13 @@ defmodule Explorer.Series do
   defp basic_numeric_operation(operation, %Series{dtype: dtype}, _),
     do: dtype_error("#{operation}/2", dtype, [:integer, :float])
 
+  defp basic_numeric_operation(operation, left, right)
+       when K.and(is_numerical(left), is_numerical(right)) do
+    raise ArgumentError,
+          "expected at least a series for the #{inspect(operation)} operation, " <>
+            "but instead got two numbers: #{operation}(#{inspect(left)}, #{inspect(right)})"
```
josevalim, line 2306:
```suggestion
            "instead got two numbers: #{inspect(left)} and #{inspect(right)}"
```

PR #531, opened 2023-03-02
Add support for slicing by series

### LINKED ISSUE #434: Allow series to be given to both Series.slice and DF.slice

philss:
This adds the feature of slicing a series by another series, or a dataframe by a series.
It works inside queries too (lazy series).

Closes #434

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -59,8 +59,8 @@ defmodule Explorer.Backend.LazySeries do
     sort: 3,
     distinct: 1,
     unordered_distinct: 1,
+    slice: 2,
     slice: 3,
-    slice_by_indices: 2,
     sample_n: 5,
     sample_frac: 5,
     head: 2,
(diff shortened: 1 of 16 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -60,6 +60,7 @@ defmodule Explorer.Backend.LazySeries do
     distinct: 1,
     unordered_distinct: 1,
     slice: 3,
+    slice_by_indices: 2,
```
josevalim, line 63: Shouldn’t this still be `slice: 2`? I thought that we try to keep the same names as the callbacks as much as possible.

josevalim (approved):
One tiny comment and ship it!

lib/explorer/backend/lazy_series.ex
philss: @josevalim The problem is that I cannot have a function with different arities because we are defining functions in the NIF module with this.
But I think I can control that elsewhere. Let me try..

philss: I think this solves the problem: https://github.com/elixir-nx/explorer/pull/531/commits/2b14c2946f5408991d4e15055aa2d5e108a6aebd

PR #535, opened 2023-03-07
Fix pivot wider if the series of column names contains nil

philss:
This is going to correct name the column as "nil" instead of breaking.

There is a potential caveat that is when a dataframe contains already a "null" column, and is going to create another "null" column. This is going go break.
Since that case is less plausable, we are ignoring it.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -3848,10 +3848,45 @@ defmodule Explorer.DataFrame do
         | C        | 15        |           | 10        | 14        |           |
         +----------+-----------+-----------+-----------+-----------+-----------+
 
+  Pivot wider can create unpredictable column names, and sometimes they can conflict with ID columns.
+  In that scenario, we add a number as suffix to duplicated column names. Here is an example:
+
+      iex> df = Explorer.DataFrame.new(
+      iex>   product_id: [1, 1, 1, 1, 2, 2, 2, 2],
+      iex>   property: ["product_id", "width_cm", "height_cm", "length_cm", "product_id", "width_cm", "height_cm", "length_cm"],
+      iex>   property_value: [1, 42, 40, 64, 2, 35, 20, 40]
+      iex> )
+      iex> Explorer.DataFrame.pivot_wider(df, "property", "property_value")
+      #Explorer.DataFrame<
+        Polars[2 x 5]
+        product_id integer [1, 2]
+        product_id_1 integer [1, 2]
+        width_cm integer [42, 35]
+        height_cm integer [40, 20]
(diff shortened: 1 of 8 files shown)


philss:
cc @josevalim @cristineguadelupe

lib/explorer/data_frame.ex
```diff
@@ -3922,28 +3922,7 @@ defmodule Explorer.DataFrame do
...
-    out_dtypes =
-      df.dtypes
-      |> Map.drop([names_from, values_from])
-      |> Map.merge(new_dtypes)
-
-    out_groups = df.groups -- [names_from, values_from]
-
-    out_df = %{df | dtypes: out_dtypes, names: out_columns, groups: out_groups}
-
     Shared.apply_impl(df, :pivot_wider, [
```
josevalim, line 3925: Do we need to do something with groups? I assume this will always ungroup it now, right? But I think the previous version was wrong in regards to groups (it was keeping groups that no longer exist in the dataframe).

josevalim: Ignore me.

lib/explorer/data_frame.ex
```diff
@@ -3898,7 +3898,7 @@ defmodule Explorer.DataFrame do
             Keyword.t()
         ) :: DataFrame.t()
   def pivot_wider(df, names_from, values_from, opts \\ []) do
-    opts = Keyword.validate!(opts, id_columns: 0..-1//1, names_prefix: "")
+    opts = Keyword.validate!(opts, id_columns: 0..-1//1, names_prefix: nil)
```
josevalim, line 3901: We can also treat prefix as empty string, no?

josevalim:
```suggestion
    opts = Keyword.validate!(opts, id_columns: &(&1.dtype != :float), names_prefix: "")
```

test/explorer/data_frame_test.exs
```diff
@@ -2001,6 +2016,34 @@ defmodule Explorer.DataFrameTest do
...
+      df = DF.new(id: [1, 1], variable: ["a", "b"], value: [1, 2], b: [4, 5])
+      df1 = DF.pivot_wider(df, "variable", "value")
+
+      assert DF.names(df1) == ["id", "b", "a", "b_1"]
+
+      assert DF.to_columns(df1, atom_keys: true) == %{
+               id: [1, 1],
+               b: [4, 5],
+               a: [1, nil],
+               b_1: [nil, 2]
```
josevalim, line 2029: Document.

josevalim (approved):

PR #537, opened 2023-03-08
Improve README by adding a short guide

philss:
This also improves the main page of our docs, by using a section of the README.


philss:
@cigrainger @josevalim I wanted to show a little bit of the API in our README, so people can get an idea of what a code using Explorer looks like. This is also useful for our main docs page.

You can see the new README in my fork: https://github.com/philss/explorer/tree/ps-improve-readme-2023-03-07
WDYT?

josevalim (approved):

PR #541, opened 2023-03-10
Add Series.ewm_mean/2 for calculating exponentially weighted moving average

sasikumar87:
Closes https://github.com/elixir-nx/explorer/issues/527


lib/explorer/series.ex
```diff
@@ -3149,6 +3149,48 @@ defmodule Explorer.Series do
...
+      >
+  """
+  @doc type: :window
+  def ewm_mean(series, opts \\ []),
+    do: Shared.apply_impl(series, :ewm_mean, [ewm_opts_with_defaults(opts)])
+
+  defp ewm_opts_with_defaults(opts) do
+    defaults = [alpha: 0.5, adjust: true, min_periods: 1, ignore_nulls: true]
+
+    Keyword.merge(defaults, opts, fn _key, _left, right -> right end)
```
philss, line 3191: You can archive almost the same by using [`Keyword.validate!/2`](https://hexdocs.pm/elixir/Keyword.html#validate!/2), that sets default values and checks if you are using only the provided keys.

I think you were based in the `window_*`, right? I think we need to update them to use `Keyword.validate!/2` (in another PR).

test/explorer/series_test.exs
```diff
@@ -3074,4 +3074,85 @@ defmodule Explorer.SeriesTest do
       assert Enum.slice(enum2, 1..2) == ["b", "c"]
     end
   end
+
+  describe "ewm_mean/2" do
+    test "returns calculated ewm values" do
```
philss, line 3079: I would split the tests for each option in different test cases. WDYT?

lib/explorer/backend/series.ex
```diff
@@ -139,6 +139,16 @@ defmodule Explorer.Backend.Series do
...
 
+  # Exponentially weighted windows
+
+  @type ewm_option() ::
+          {:alpha, float()}
+          | {:adjust, boolean()}
+          | {:min_periods, integer()}
+          | {:ignore_nulls, boolean()}
+
+  @callback ewm_mean(s, [ewm_option()]) :: s
```
philss, line 150: I liked the approach, and I think we could use more in our functions, instead of requiring each option to be an argument :+1:

lib/explorer/series.ex
```diff
@@ -3149,6 +3149,48 @@ defmodule Explorer.Series do
...
+    * `:alpha` - Optional smoothing factor which specifies the imporance given
+      to most recent observations. It is a value such that, 0 < alpha <= 1. Defaults to 0.5.
+
+    * `:adjust` - If set to true, it corrects the bias introduced by smoothing process.
+      Defaults to `true`.
+
+    * `:min_periods` - The number of values in the window that should be non-nil
+      before computing a result. Defaults to `1`.
+
+    * `:ignore_nulls` - If set to true, it ignore nulls in the calculation. Defaults to `true`.
```
philss, line 3166: This option could be named `ignore_nils`, since we are talking from the Elixir user perspective.

philss (commented):
This looks really good!! :D

I added some comments.

lib/explorer/backend/series.ex
sasikumar87: Sorry, I couldn't get what you suggest here.

philss: Oh, it's just a suggestion that we should do more of what you did here :)
I think I'm going to change some of our callbacks to this style. Specially the callbacks related to IO.

philss (approved):
:shipit:

lib/explorer/backend/series.ex
josevalim: There is a downside in this approach: if we add new options in the future, then existing implementations may forget to process said new options. Because of that, I am thinking we should perhaps go with the existing approach of passing each option as argument. Thoughts @philss?

lib/explorer/series.ex
```diff
@@ -3149,6 +3149,47 @@ defmodule Explorer.Series do
...
+      Defaults to `true`.
+
+    * `:min_periods` - The number of values in the window that should be non-nil
+      before computing a result. Defaults to `1`.
+
+    * `:ignore_nils` - If set to true, it ignore nulls in the calculation. Defaults to `true`.
+
+  ## Examples
+
+      iex> s = 1..10 |> Enum.to_list() |> Explorer.Series.from_list()
```
josevalim, line 3170: To simplify the examples, let's use a range from `1..5`?

lib/explorer/backend/series.ex
philss: @josevalim you got me. Yeah, this downside is a problem. But on the other hand, it would give more flexibility and compatibility between versions. For now I think we could stick to the existing approach, and revisit this when we have another backend planned.

@sasikumar87 could you change this callback to use arguments instead of keyword args?

sasikumar87: @philss @josevalim Updated.
But `window_*` functions still use this kind of approach, should change them too?

josevalim: Yes, a PR that changes that will be welcome. Maybe instead of:

    do: Shared.apply_impl(series, :window_max, [window_size, window_opts_with_defaults(opts)])

We could do:

    do: Shared.apply_impl(series, :window_max, [window_size | window_args(opts)])

and `window_args` returns a list.

josevalim:

PR #542, opened 2023-03-11
Use Keyword.validate!/2 instead of Keyword.merge/3

sasikumar87:
This along with setting defaults, also checks if only the provided keys are used and raises error, if not.


philss:
@sasikumar87 thank you! :purple_heart:

PR #549, opened 2023-03-24
Add log/2 and exponential/1 to Series

### LINKED ISSUE #548: Add support for log and exp operations in DF.mutate
Currently, there is no other way to calculate the logarithm or exponent of the DataFrame column than separating it from DataFrame, calculating log/exp and putting it back.

philss:
May close #548 

This version only accepts `log(series, number)`, since Polars implements the equivalent for series.
Do you think we need `log(series, series)` and `log(number, series)`? If so, I can implement in upcoming PRs.

Also, I don't know much about `exponential`, so please help me documenting that.

cc/ @msluszniak @kimjoaoun


lib/explorer/series.ex
```diff
@@ -2244,6 +2244,51 @@ defmodule Explorer.Series do
...
+      iex> Explorer.Series.log(s, 2.0)
+      #Explorer.Series<
+        Polars[3]
+        float [3.0, 4.0, 5.0]
+      >
+
+  """
+  @doc type: :element_wise
+  @spec log(argument :: Series.t(), base :: number()) :: Series.t()
+  def log(argument, base \\ 2) when is_number(base) do
```
josevalim, line 2278: For compatibility with Nx, the default should be a natural logarithm (where the base is `e`). My suggestion is to have `log/1` and `log/2` as distinct functions.

lib/explorer/series.ex
```diff
@@ -2244,6 +2244,51 @@ defmodule Explorer.Series do
...
+
+    basic_numeric_operation(:log, argument, base)
+  end
+
+  @doc """
+  Calculate the exponential of all elements.
+  """
+  @doc type: :element_wise
+  @spec exponential(Series.t()) :: Series.t()
+  def exponential(%Series{} = s), do: Shared.apply_impl(s, :exponential)
```
josevalim, line 2290: Let's call it `exp` for consistency with Nx too. :)

lib/explorer/series.ex
```diff
@@ -2244,6 +2244,51 @@ defmodule Explorer.Series do
...
+  @spec log(argument :: Series.t(), base :: number()) :: Series.t()
+  def log(argument, base \\ 2) when is_number(base) do
+    if base <= 0, do: raise(ArgumentError, "base must be a positive, non-zero, number")
+    if base == 1, do: raise(ArgumentError, "base cannot be equal to 1")
+
+    basic_numeric_operation(:log, argument, base)
+  end
+
+  @doc """
+  Calculate the exponential of all elements.
```
josevalim, line 2286:
```suggestion
  Calculates the exponential of all elements.
```

lib/explorer/series.ex
```diff
@@ -2244,6 +2244,51 @@ defmodule Explorer.Series do
   @spec pow(left :: Series.t() | number(), right :: Series.t() | number()) :: Series.t()
   def pow(left, right), do: basic_numeric_operation(:pow, left, right)
 
+  @doc """
+  Calculate the exponent of a logarithm.
```
josevalim, line 2248:
```suggestion
  Calculates the logarithm on a given base.
```

lib/explorer/series.ex
```diff
@@ -2244,6 +2244,51 @@ defmodule Explorer.Series do
...
+
+  The resultant series is going to be of dtype :integer
+  if the series and the base are both integers. Otherwise
+  the it is going to be a float series.
```
msluszniak, line 2252: I guess the log function should always return float series. For Example $log_2 (3) \approx 1.58496$

josevalim: Agreed on always returning floats!

lib/explorer/series.ex
```diff
@@ -2244,6 +2244,71 @@ defmodule Explorer.Series do
...
+      #Explorer.Series<
+        Polars[3]
+        float [3.0, 4.0, 5.0]
+      >
+
+  """
+  @doc type: :element_wise
+  @spec log(argument :: Series.t(), base :: number()) :: Series.t()
+  def log(argument, base) when is_number(base) do
+    if base <= 0, do: raise(ArgumentError, "base must be a positive number")
```
msluszniak, line 2297: I think that it will be better, in general, to return negative infinity for 0 in the logarithm. However, it looks like there is no support for infinity in Explorer Series/DataFrame.

image

josevalim: @msluszniak we do support it, it seems the error is elsewhere. Likely Kino.Explorer.

msluszniak: Yes, probably the aggregate functions like min, max, and mean don't work properly. I'll send an issue to Kino.Explorer.

josevalim: The bug is in Explorer as well. Reported here: https://github.com/elixir-nx/explorer/issues/550

lib/explorer/series.ex
```diff
@@ -2244,6 +2244,71 @@ defmodule Explorer.Series do
...
+      iex> s = Explorer.Series.from_list([1, 2, 3, nil, 4])
+      iex> Explorer.Series.log(s)
+      #Explorer.Series<
+        Polars[5]
+        float [0.0, 0.6931471805599453, 1.0986122886681098, nil, 1.3862943611198906]
+      >
+
+  """
+  @doc type: :element_wise
+  @spec log(argument :: Series.t()) :: Series.t()
```
josevalim, line 2271: Instead of having a default value, I think we should really call `log/1`. This will help us avoid precision issues if we support multiple float types in the future and I believe the natural log may be more optimized. For example, it seems Rust implements all other logs in favor of `ln`: https://doc.rust-lang.org/src/std/f64.rs.html#428-430

josevalim (approved):

PR #551, opened 2023-03-24
Fix encoding of special float values for Series

philss:
This commit fixes the encoding of the return of some aggregation functions for the Series module.

We don't have the same problem for expressions because the result of a lazy series evaluation is always a series.

Closes https://github.com/elixir-nx/explorer/issues/550


philss:
There is a problem I chose not to fix in this PR: the specs of most of the `Series` are missing the option to return a lazy series. I should fix it after the merge of this PR.

PR #557, opened 2023-04-01
Fix `nif_not_loaded` error when `Series.ewm_mean/2` is called from query

sasikumar87:


philss:
@sasikumar87 thank you! :purple_heart:

josevalim:
I will push some tests. we should probably add tests for all new expressions we add. :)

PR #559, opened 2023-04-03
Read parquet dataframe from S3

philss:
This is a work-in-progress PR that aims to read dataframes from AWS S3 (and compatible services).

Currently only parquet is supported by the Polars, AFAIK.

I don't know yet how I'm going to test this, but I'm investigating.


josevalim:
For testing, perhaps we can use Plug to start a webserver running on localhost and then pass the path of a parquet file which will read from disk and send it with `Plug.Conn.sendfile`?

philss:
I'm closing this for now until we decide about the API.

PR #567, opened 2023-04-11
Add experimental IO writer that talks to a process

philss:
This is an experiment to prove a concept about sending the chunks of a dataframe to Elixir, in a given format - this case we are using IPC - in order to send them via web.

It's feasible, although we need to obey some restrictions: https://docs.rs/rustler/latest/rustler/env/struct.Env.html#method.send

PS: open a PR just to keep the reference. It's going to be closed.

### DIFF
--- a/lib/explorer/polars_backend/experimental_io.ex
+++ b/lib/explorer/polars_backend/experimental_io.ex
@@ -1,0 +1,29 @@
+defmodule Explorer.PolarsBackend.ExperimentalIo do
+  @moduledoc false
+
+  # To try it:
+  #
+  # iex> df = Explorer.Datasets.wine()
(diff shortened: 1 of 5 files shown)


josevalim:
Thank you @philss, so it seems we will need to (1) start a thread from the Rust side and then (2) use something for concurrency control (perhaps a mutex will suffice). We can focus on 1 next and then tackle 2. Also please ping the Rustler folks, maybe they will have ideas.

philss:
@josevalim I'm going to close this for now.

Talking with the guys from Rustler, they suggest an slightly different approach of using that `ResourceArc::make_binary` (from https://github.com/rusterlium/rustler/pull/487) function and kind of go "pooling" the data that is arriving, making an iterator. This would probably avoid the need to send data via processes, and we could have zero-copy. I'm not entirely sure of how it would look like, but I plan to try when we need it.

Issue #569, opened 2023-04-11
DataFrame.iter_rows method?


mlineen:
For larger dataframes, I've found the pattern below to be significantly faster than waiting for and iterating over the result of `DataFrame.to_rows/1`

Should Explorer implement something like [DataFrame.iter_rows](https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.iter_rows.html#polars-dataframe-iter-rows) with a similar warning?

> Row iteration is not optimal as the underlying data is stored in columnar form

```
{n_rows, _} = Explorer.DataFrame.shape(df)
chunk_size = 1_000

# calculate offsets
Range.new(0, n_rows - 1, chunk_size)
|> Stream.map(fn offset -> Explorer.DataFrame.slice(df, offset, chunk_size) end)
|> Stream.map(&Explorer.DataFrame.to_rows/1)
# process each row
|> Enum.each(&IO.inspect/1)
```

I'm also considering adding `Task.async_stream` into this pipeline for concurrency, which is why I'm not sure if the project would prefer this type of processing not be core to the Explorer module.

josevalim:
I believe we could add a `to_rows_stream`, which chunk size as option. A PR is welcome. We can copy the disclaimer from `to_rows` itself. @cigrainger/@philss any other suggestions on the name?

cigrainger:
@josevalim I like the name. And I think this is a great feature to have. :)

PR #571, opened 2023-04-12
add Explorer.DataFrame.to_rows_stream method

mlineen:
PR for #569

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1525,6 +1525,40 @@ defmodule Explorer.DataFrame do
     Shared.apply_impl(df, :to_rows, [opts[:atom_keys]])
   end
 
+  @doc """
+  Converts a dataframe to a stream of maps (rows).
+
+  > #### Warning {: .warning}
+  >
+  > This is an expensive operation since data is stored in a columnar format.
+  > Prefer to use the operations in this module rather than the ones in `Enum`
+  > whenever possible, as this module is optimized for large series.
+
+  ## Options
+
+    * `:atom_keys` - Configure if the resultant maps should have atom keys. (default: `false`)
+    * `:buffer_size` - The number of rows that are buffered internally while iterating over the data. (default: `1000`)
+
+  ## Examples
+
+      iex> df = Explorer.DataFrame.new(floats: [1.0, 2.0], ints: [1, nil])
+      iex> Explorer.DataFrame.to_rows_stream(df) |> Enum.map(& &1)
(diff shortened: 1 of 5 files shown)


philss:
@mlineen thank you! :purple_heart:  We are trying to have tests for all the functions, so would you mind to add a test covering this new one?

And also, for what I understood, this is going to work only for eager dataframes, right? Maybe we could have a better error message in case a lazy frame is in use.

mlineen:
Sure, it wasn't clear how you wanted to test; I think `to_rows` is only tested via doctests right now which is what I patterned it off of. I'll take a look at the lazy frame issue too.

lib/explorer/data_frame.ex
```diff
@@ -1525,6 +1525,40 @@ defmodule Explorer.DataFrame do
...
+  > #### Warning {: .warning}
+  >
+  > This is an expensive operation since data is stored in a columnar format.
+  > Prefer to use the operations in this module rather than the ones in `Enum`
+  > whenever possible, as this module is optimized for large series.
+
+  ## Options
+
+    * `:atom_keys` - Configure if the resultant maps should have atom keys. (default: `false`)
+    * `:buffer_size` - The number of rows that are buffered internally while iterating over the data. (default: `1000`)
```
josevalim, line 1540: Can we call it chunk_size? I don't think buffer size is correct because buffer means it is somehow being accumulated somewhere (in disk or in memory) but we are really just traversing it using a chunk.

lib/explorer/data_frame.ex
```diff
@@ -1525,6 +1525,40 @@ defmodule Explorer.DataFrame do
...
+      [%{"floats" => 1.0, "ints" => 1}, %{"floats" => 2.0 ,"ints" => nil}]
+
+      iex> df = Explorer.DataFrame.new(floats: [1.0, 2.0], ints: [1, nil])
+      iex> Explorer.DataFrame.to_rows_stream(df, atom_keys: true) |> Enum.map(& &1)
+      [%{floats: 1.0, ints: 1}, %{floats: 2.0, ints: nil}]
+  """
+  @doc type: :conversion
+  @spec to_rows_stream(df :: DataFrame.t(), Keyword.t()) :: Stream.t()
+  def to_rows_stream(df, opts \\ []) do
+    opts = Keyword.validate!(opts, atom_keys: false, buffer_size: 1_000)
```
josevalim, line 1555:
```suggestion
    opts = Keyword.validate!(opts, atom_keys: false, buffer_size: 1_000)
    atom_keys = opts[:atom_keys]
    buffer_size = opts[:buffer_size]
```

So we don't access it every time. :)

PR #572, opened 2023-04-12
Update Rust Nightly version to "2023-02-14"

philss:
This makes it even with the version Polars is using.

PS: it's also an attempt to fix the build for ARM 32 bits.


josevalim (approved):

philss:
There is a problem happening on CI and I can't reproduce locally.
I'm going to hold this change for now, and try another time :/

PR #576, opened 2023-04-19
fix: handle datetimes with other resolutions than MS for series aggregation

### LINKED ISSUE #575: Min not working for datetime[NS]
Hi. Thank you for a great lib.

When working with some parquet data with a timestamp of precesion NS `datetime[NS]`, the Series.min fails with a panic error. Looking at the code it looks like we only handle the case where the datetime precision is MS.
[https://github.com/elixir-nx/explorer/blob/bcf150595adedb592e028659f435ba3b169a5e93/native/explorer/src/series.rs#L708](https://github.com/elixir-nx/explorer/blob/bcf150595adedb592e028659f435ba3b169a5e93/native/explorer/src/series.rs#L708)

This is also the case with other aggregation functions such as max, is there are gracefull way to handle both NS,MS

Hanspagh:
As discussed in the issue when deal with this by using `encode_datetime`, my rust is a bit rusty, so let me know if there is better ways to do this, maybe we want to handle errors from `encode_datetime` atm I assume it would never fail.

fix #575


philss (commented):
Cool! It looks good! :+1: 

But would be nice to add some tests covering the functions that are using those NIFs (in the `series_test.exs`). I think we don't have tests for datetimes there. Since it's not possible to use NS in Elixir, I would add a test loading a small parquet file with a series containing datetimes with NS and then call `min`. WDYT?

Hanspagh:
Yea, sure I can do this. Was wondering what the best way to get some NS date-times

Hanspagh:
So while adding the test, I notice one more function that does not handle the `ns` namely the to `day_of_week`. I am unsure how we want to handle this, do we extend the `timestamp_to_datetime` to handle different formats or do you have a better suggestion?

josevalim:
@Hanspagh I looked at the code and it seems Polars supports weekday now. So that issue can be solved like this it seems:

```rust
pub fn s_day_of_week(s: ExSeries) -> Result<ExSeries, ExplorerError> {
    let s2 = s
            .weekday()?
            .cast(&DataType::Int64)?;

    Ok(ExSeries::new(s2))
}
```

Can you please send a PR with a test? :)

josevalim:
I will go ahead and push the changes above cause I am working on something around this area. :)

Issue #577, opened 2023-04-19
Largish CSV hangs on from_csv/load_csv


jeregrine:
I have a ~90mb CSV with many columns. When I attempt to load it explorer just hangs, I waited 200s before giving up. 

When I run it through NimbleCsv I get a parse error. I suspect it's erroring and not being handled? 

```elixir
Mix.install([
  {:req, "~> 0.3.6"},
  {:explorer, "~> 0.5.6"}
])

mprop_file = "/tmp/mprop.csv"
unless File.exists?(mprop_file) do
  Req.get!("https://data.milwaukee.gov/dataset/562ab824-48a5-42cd-b714-87e205e489ba/resource/0a2c7f31-cd15-4151-8222-09dd57d5f16d/download/mprop.csv", output: mprop_file)
end

IO.puts("Loading #{mprop_file}")
IO.puts("File Size: #{File.stat!(mprop_file).size}")

mprop_explorer = Explorer.DataFrame.from_csv!(mprop_file)

IO.inspect(mprop_explorer)
```

treebee:
@jeregrine is the result in `/tmp/mprop-nd.json` valid nd-json ? I would assume you need another `jq` step to ensure that every json object is on a single line:  `jq -c '.elements | .[]' | jq -c . > /tmp/mprop-nd.json`

josevalim:
I believe they are a couple action items here:

1. Can we make the delimiter be a whole string instead of a single character?
2. Can we reproduce the issue where Polars takes forever in Rust itself?
3. Can we make sure that invalid outputs return a better error rather than panicking?

jeregrine:
The data I linked in the scripts are public data so feel free to download. https://data.milwaukee.gov/dataset/mprop

jeregrine:
This was more of a "lets play around with explorer using a known dataset" and less of a serious blocker project for me ATM. NimbleCSV also gave a fairly cryptic error on line 1 till I did this step, even with proper newline configured, and a list of all 3 possible options. 

I did the ole "open in excel, export to csv" trick and was able use NimbleCSV to open this, but explorer did give a better error message this time!

```
[...]

PR #578, opened 2023-04-19
End of Line For CSV Loading/Parsing

jeregrine:
The Polars CSVReader lets you define an eol_char. I can't test this due to the mac build bug so I am opening a WIP PR. 

Related to #577

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -41,7 +41,8 @@ defmodule Explorer.Backend.DataFrame do
               max_rows :: option(integer()),
               columns :: columns_for_io(),
               infer_schema_length :: option(integer()),
-              parse_dates :: boolean()
+              parse_dates :: boolean(),
+              eol_char :: option(String.t())
             ) :: result(df)
   @callback to_csv(df, filename :: String.t(), header? :: boolean(), delimiter :: String.t()) ::
               ok_result()
@@ -58,7 +59,8 @@ defmodule Explorer.Backend.DataFrame do
               max_rows :: option(integer()),
               columns :: columns_for_io(),
               infer_schema_length :: option(integer()),
-              parse_dates :: boolean()
+              parse_dates :: boolean(),
+              eol_char :: option(String.t())
             ) :: result(df)
 
   # IO: Parquet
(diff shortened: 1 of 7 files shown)


philss:
@jeregrine please run `mix ci` to fix the issues.

philss:
@jeregrine please try with an older version of nightly. We are using `nightly-2023-02-14` in the CI.

jeregrine:
Hey that worked. Pushed up the format

jeregrine:
I'll work on adding a couple tests probably tomorrow.

jeregrine:
I just get "Argument Error" whenever I try to set this. 

Also I suspect I will need to fix this upstream in polars since u8 isn't enough to handle \r\n.

treebee:
@jeregrine the `Argument Error` comes from passing `eol_char` as a string, a utf8 encoded binary.
You can check how it's done for the `delimiter` option.

jeregrine:
> @jeregrine the `Argument Error` comes from passing `eol_char` as a string, a utf8 encoded binary. You can check how it's done for the `delimiter` option.

A single character of a utf8 encoded binary is a 8 bitunsigned number. 

Rust:

    delimiter_as_byte: u8,

Elixir:
        delimiter: ",",

The only real difference I can see is the pattern match before the call. 

        <<delimiter::utf8>>,

You can understand my confusion then
Screenshot 2023-04-24 at 12 28 27 PM

jeregrine:
I'm going to close this cause I think we're better spent fixing this upstream AND THEN making all the required changes.

philss:
> The only real difference I can see is the pattern match before the call.
>
>    <<delimiter::utf8>>,

@jeregrine Yeah, we pass that delimiter, assuming that is one char long, as an `u8` number (`delimiter` is going to be a number in this pattern matching).
I think we want to be able to set ` \r\n` if needed, so would be cool to be able to do that in Polars.

But, for the case that you mentioned in #577, I think this change would suffice. I'm OK to support a single char, and then we add support for multiple when fixed upstream.

josevalim:
:+1: for `eol_delimiter` as a string which we still enforce to be single byte.

philss:
@jeregrine would you mind to just add a test? I think something around `load_csv` is enough. You can generate a string that uses, for example, `\r` as the `eol_delimiter`.

Also, I think we can change to `eol_delimiter` like José suggested. WDYT?

jeregrine:
I need to update the spots to use u8 and I need to swap it out. But yea Ive got tests locally I havent pushed up. On Apr 24, 2023, at 4:32 PM, Philip Sampaio ***@***.***> wrote:﻿
@jeregrine would you mind to just add a test? I think something around load_csv is enough. You can generate a string that uses, for example, \r as the eol_delimiter.
Also, I think we can change to eol_delimiter like José suggested. WDYT?

jeregrine:
@philss I added the changes you requested.

philss (commented):
I think `eol_delimiter` is a better name.
But overall it looks good! :ship:

jeregrine:
accepted your changes

philss (approved):
:ship:

josevalim:

jeregrine:
<3 Happy to help anyway I can.

Issue #568, opened 2023-04-11
Allow diff of date and datetime series


sasikumar87:
How should the API be like?

josevalim:
Can you expand what you mean?

sasikumar87:
Assuming a dataframe like `DataFrame.new(a: [~D[2023-04-11], ~D[2023-04-10]], b: [~D[2023-04-06], ~D[2023-04-05]])`, I need to find the diff between `a` and `b`.

josevalim:
I see! How is it called in Pandas and Polars? What about dplyr? Those can provide some initial reference points.

sasikumar87:
Pandas seems to support `-` for date and datetime. For conversion to different units like day, hour etc, they allow this subtracted value to be divided by [timedelta64](https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime-and-timedelta-arithmetic) type, which can specify the unit of measurement.

Polars also seems to support `-` through [sub](https://docs.rs/polars/latest/polars/export/chrono/struct.Date.html#method.sub), returning Duration type. From duration, we can calculate different units of measurement with functions like `num_days`, `num_hours` etc.

dplyr seems to have [difftime](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/difftime) which can support both date and datetime with `units` option to return as weeks, days, hours, etc.

In Elixir, `Date.diff/2` returns number of days, while `NaiveDateTime.diff/3` returns number of seconds / milliseconds etc based on third argument, which cannot be greater than a second.

Given this, I'm thinking of `Series.diff/3`, with third argument as an atom that supports `weeks`, `days`, `hours`, `minutes`, `seconds`, `milliseconds`, `microseconds`, `nanoseconds`, all the values that can be converted from polars Duration type. And we can also make third argument optional with `days` as default if the the series is of date type and `seconds` for datetime type, to maintain Elixir's default convention.

Will this be ok?

sasikumar87:
One problem is `diff` sounds like an arithmetic operation which supports numbers, but here we are using it only for date and datetime. Can it be `Series.diff_date/3` and `Series.diff_datetime/3` ?

josevalim:
[...]

PR #581, opened 2023-04-24
feat: add date_diff for date series

Hanspagh:
Started on #568


native/explorer/src/series.rs
```diff
@@ -1215,6 +1215,24 @@ pub fn s_to_time(s: ExSeries) -> Result<ExSeries, ExplorerError> {
...
+    let s1 = other.clone_inner();
+
+    let res = (s - s1)
+        .duration()?
+        .into_iter()
+        .map(|d| match d {
+            Some(d) => Some(d / 86400000),
+            None => None,
+        })
+        .collect::<Int64Chunked>()
```
josevalim, line 1230: Do you need to iterate? I believe you should be able to call `/ 86400000` or similar on a series to compute the division.

Issue #585, opened 2023-04-25
[idea] Allow pivot_wider on categorical values


dkuku:
In machine learning there is often need for sparse columns created from categorical values. The easiest way to it I found this, but it would be nice to just specify a single column like that and fill the defaults with ones (and maybe the rest with zeros ?)
`Explorer.DataFrame.pivot_wider("g")`
```
DF.new(%{id: [11,12,13], g: ~w/m f m/, gid: [1,1,1]})
|> Explorer.DataFrame.pivot_wider("g", "gid")
|> Explorer.DataFrame.table

+--------------------------------------------+
| Explorer DataFrame: [rows: 3, columns: 3]  |
+--------------+--------------+--------------+
|      id      |      m       |      f       |
|  <integer>   |  <integer>   |  <integer>   |
+==============+==============+==============+
| 11           | 1            |              |
+--------------+--------------+--------------+
| 12           |              | 1            |
+--------------+--------------+--------------+
| 13           | 1            |              |
+--------------+--------------+--------------+

```

josevalim:
It should be possible today. Maybe we just need to relax the checks. Can you give it a try please?

dkuku:
Sorry - I clicked enter to early, I was updating the query but it's not livebook - here it just sends the message.

josevalim:
Sorry, I am confused then. Can you please provide a more complete example?

dkuku:
I'm thinking that when we know that this is a categorical value then we don't need to create another column just to create the ones and zeros.
Explorer.DataFrame.pivot_wider(df, "g") should be enough. Maybe with an option to fill empty values with zeros. I think this is a common usecase when preparing data for further processing.

dkuku:
also instead of having second column as a single value would be an option
`Explorer.DataFrame.pivot_wider(df, "g", 1)`

josevalim:
Let’s go in parts then. :) A PR to support categorical values is welcome. It is probably a small change and adding categories to the list of supports columns. Do you want to give it a try? :)

dkuku:
I may try, I never looked at the code but when it's elixir then I should be fine.

josevalim:
Yes, my assumption is that we only need to change the elixir bits

dkuku:
We just need to align how it should look. I think that creating additional column is not needed here. In other places we also accept either column or scalar value.

dkuku:
[...]

PR #586, opened 2023-04-28
Allow pivot_wider to use category fields

dkuku:
Currently, a data frame can't be created when passing a series directly - I had to update it in tests.
 #585 
 
``` 
DF.new(category: Series.from_list(["a", "b", "a"], dtype: :category))
```


josevalim:

PR #595, opened 2023-05-11
Update Polars to v0.29

philss:
I'm leave as draft because I'm investigating if it's possible to make the Parquet compression with Gzip work again.
For some reason it is giving a segfault with recent versions of Rust Nightly. I'm also going to check if this works with older versions of Nightly.

There are two breaking changes:

- the way `pivot_wider` now works when there are multiple columns in `values_from` is to repeat the original column in the columns created.
- since casting a `:time` series to `:string` is now better supported, I noticed that we were working with the wrong time unit for that type of series. Polars is using `nanoseconds` to represent time series, and we were using `microseconds`. I believe we should stick to nanoseconds in order to prevent some operations from breaking (like casting, or arithmetic ops).


josevalim:
Can you send a separate PR only for the nanoseconds change? I am doing the same change in another branch and then I will merge it and resolve the conflict locally.

philss:
> Can you send a separate PR only for the nanoseconds change? I am doing the same change in another branch and then I will merge it and resolve the conflict locally.

@josevalim sure! 1 min...

josevalim (approved):

PR #596, opened 2023-05-11
Fix representation of time series from us to ns

philss:
Since Polars represents time series as nanoseconds internally, we should also do the same.
This is because some conversions, like casting to string, cannot work properly if we treat the time as microseconds.

### DIFF
--- a/native/explorer/src/datatypes.rs
+++ b/native/explorer/src/datatypes.rs
@@ -301,24 +301,31 @@ pub struct ExTime {
     pub microsecond: (u32, u32),
 }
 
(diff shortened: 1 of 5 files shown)


native/explorer/src/datatypes.rs
```diff
@@ -301,24 +301,31 @@ pub struct ExTime {
     pub microsecond: (u32, u32),
 }
 
-pub use polars::export::arrow::temporal_conversions::time64us_to_time as timestamp_to_time;
+// pub use polars::export::arrow::temporal_conversions::time64us_to_time as timestamp_to_time;
```
josevalim, line 304:
```suggestion
```

josevalim (approved):

PR #598, opened 2023-05-17
WIP: Add Series to_arrow/from_arrow

philss:
This is still a proof of concept.

### DIFF
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -1214,20 +1214,32 @@ pub fn s_atan(s: ExSeries) -> Result<ExSeries, ExplorerError> {
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn s_to_arrow(env: Env, s: ExSeries) -> Result<Binary, ExplorerError> {
+pub fn s_to_arrow(env: Env, s: ExSeries) -> Result<i64, ExplorerError> {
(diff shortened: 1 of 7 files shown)


native/explorer/src/series.rs
```diff
@@ -1210,3 +1212,42 @@ pub fn s_atan(s: ExSeries) -> Result<ExSeries, ExplorerError> {
...
+    let transmuted: &[u8] =
+        unsafe { slice::from_raw_parts(array_ptr as *const u8, mem::size_of::<ffi::ArrowArray>()) };
+
+    let bin = unsafe { res.make_binary_unsafe(env, |_| transmuted) };
+
+    // We tell the compiler to forget about this box, and therefore don't
+    // call the destructor (drop). This will keep the array in memory, without
+    // causing a segfault. But it's probably considered a leak.
+    // See: https://doc.rust-lang.org/std/mem/fn.forget.html
+    mem::forget(c_array);
```
josevalim, line 1234: I am assuming `c_array` is freshly allocated by this function and it is not a reference to the Series. If so, this is 100% a leak. In this case, we have two options:

* Copy the binary

* If there is a `ffi` function that tells us the byte size of the Arrow Array, then we can ask Erlang to allocate a NewBinary and have the `ffi::export_array_to_c` write to this new binary

PR #600, opened 2023-05-23
Fix infer_schema_length: nil option

awerment:
Currently, setting `infer_schema_length: nil` when creating a `DataFrame` from CSV has no effect, because there is a redundant fallback to the default value of 1000 rows if the `max_rows` option is not set (or set to `nil`). 

Effectively, it is not possible to use all rows for schema inference without setting `infer_schema_length` (or `max_rows`) to a large enough "guess". I believe this does not match the documented behavior:
> `:max_rows` - Maximum number of lines to read. (default: nil)
> `:infer_schema_length` Maximum number of rows read for schema inference. Setting this to nil will do a full table scan and will be slow (default: 1000).

This PR removes the fallback to the default.

I'm not sure if the changes here are actually preferable to changing the docs, as scanning all rows will be slow for larger files (as stated), but ran into an issue myself with the following situation:
- the files' line count is not known upfront (guessing a large enough `infer_schema_length` / `max_rows` value felt wrong)
- first `N`k lines contain `0`, later lines contain float values. (an issue with the source of the files, but fixing that is not possible)

### DIFF
--- a/test/explorer/data_frame/csv_test.exs
+++ b/test/explorer/data_frame/csv_test.exs
@@ -132,6 +132,8 @@ defmodule Explorer.DataFrame.CSVTest do
   end
 
   describe "from_csv/2 options" do
+    @default_infer_schema_length 1_000
+
     @tag :tmp_dir
     test "delimiter", config do
       csv =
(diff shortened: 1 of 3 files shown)


test/explorer/data_frame/csv_test.exs
```diff
@@ -211,6 +213,84 @@ defmodule Explorer.DataFrame.CSVTest do
...
+      csv =
+        tmp_csv(config.tmp_dir, """
+        a
+        #{1..10 |> Enum.join("\n")}
+        1.0
+        """)
+
+      assert_raise RuntimeError, ~r/from_csv failed:/, fn ->
+        DF.from_csv!(csv, infer_schema_length: nil, max_rows: 10)
+      end
```
awerment, line 291: The reason why this second case causes a `RuntimeError` is that there are no guarantees with `max_rows`, as documented [here](https://docs.rs/polars/latest/polars/prelude/struct.LazyCsvReader.html#method.with_n_rows).

josevalim:

PR #605, opened 2023-05-30
WIP: try to improve caching on CI

philss:
Introduces the https://github.com/Swatinem/rust-cache Action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,26 +26,14 @@ jobs:
     name: test
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/cache@v3
-        with:
-          path: |
-            deps
-            _build
-          key: ${{ runner.os }}-mix-${{ matrix.elixir_version }}-${{ hashFiles('**/mix.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-mix-
-
-      - uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            native/explorer/target/
-            priv/native/
-            _build/test/lib/explorer
-          key: ${{ runner.os }}-cargo-${{ env.RUST_TOOLCHAIN_VERSION }}-${{ hashFiles('**/Cargo.lock') }}
+      # - uses: actions/cache@v3
+      #   with:
+      #     path: |
+      #       deps
+      #       _build
+      #     key: ${{ runner.os }}-mix-${{ matrix.elixir_version }}-${{ hashFiles('**/mix.lock') }}
+      #     restore-keys: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -20,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        nif: ["2.17", "2.16", "2.15"]
```
philss, line 23: We don't need to build for `2.17` yet, because we don't use anything specifically from that version. Rustler Precompiled will select the `2.16`.

.github/workflows/ci.yml
```diff
@@ -26,26 +26,18 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/cache@v3
-        with:
-          path: |
-            deps
-            _build
-          key: ${{ runner.os }}-mix-${{ matrix.elixir_version }}-${{ hashFiles('**/mix.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-mix-
```
philss, line 36: I chose to remove this cache because it's much faster to rebuild the deps then restoring the cache. The only exception is the build of the native code.

.github/workflows/release.yml
```diff
@@ -48,6 +53,13 @@ jobs:
           toolchain: nightly-2023-02-14
           targets: ${{ matrix.job.target }}
 
+      - uses: Swatinem/rust-cache@v2
```
philss, line 56: This Action makes a lot easier to cache if we are only dealing with a Rust project. Since the precompilation is entirely done by cargo, and we have control over the place the target is saved, I decided to use this here.

philss: Another thing I didn't mention is that this caching should reduce the precompilation time in half, which is great! :rocket:

.github/workflows/rust-ci.yml
```diff
@@ -24,25 +24,18 @@ jobs:
...
-          key: ${{ runner.os }}-cargo-${{ env.RUST_TOOLCHAIN_VERSION }}-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-
 
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: "${{ env.RUST_TOOLCHAIN_VERSION }}"
           components: rustfmt, clippy
 
+      - uses: Swatinem/rust-cache@v2
```
philss, line 33: Here we use the "rust-cache" action as well, because we only deal with `cargo`.

philss (commented):
I think it's better now.

The cached Elixir build is now taking around 2min30s, against the non-cached that is taking around 8 min. The simplification in the cached paths makes harder to have a miss. So we should have quicker suite.

The precompilation workflow now takes around 18 minutes, compared to the 30 minutes from the version without any cache. This one is harder to improve, but I should iterate on this in another moment.

PR #611, opened 2023-06-12
Add cumulative product to Series

guarilha:
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.cumprod.html


philss (commented):
Hey @guarilha :wave: 

This is looking great!

Just one thing: could you please add a column with this function to this test case? => https://github.com/elixir-nx/explorer/blob/7ef2bbb83933f3f42e905799a7ecec8122b70339/test/explorer/data_frame_test.exs#L1018

We basically cover the lazy version there.

Thank you! :purple_heart:

philss (approved):

PR #612, opened 2023-06-12
add Series.abs/1

guarilha:
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.abs.html


josevalim (approved):

philss (commented):
@guarilha I think we could have a test like you did in the previous PR.

philss:
Thank you! :purple_heart:

PR #614, opened 2023-06-14
Add Series.skew/2

guarilha:
https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.skew.html

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1867,7 +1867,11 @@ defmodule Explorer.Series do
   @doc """
   Compute the sample skewness of a data set.
 
-  For normally distributed data, the skewness should be about zero. For unimodal continuous distributions, a skewness value greater than zero means that there is more weight in the right tail of the distribution. The function skewtest can be used to determine if the skewness value is close enough to zero, statistically speaking.
+  For normally distributed data, the skewness should be about zero.
+  For unimodal continuous distributions, a skewness value greater
+  than zero means that there is more weight in the right tail of the
+  distribution. The function skewtest can be used to determine if
+  the skewness value is close enough to zero, statistically speaking.
 
   See scipy.stats for more information.
 
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1868,12 +1868,10 @@ defmodule Explorer.Series do
   Compute the sample skewness of a series.
 
   For normally distributed data, the skewness should be about zero.
+
   For unimodal continuous distributions, a skewness value greater
   than zero means that there is more weight in the right tail of the
-  distribution. The function skewtest can be used to determine if
-  the skewness value is close enough to zero, statistically speaking.
-
-  See scipy.stats for more information.
+  distribution.
 
   ## Supported dtypes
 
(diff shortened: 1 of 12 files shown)


lib/explorer/series.ex
```diff
@@ -1864,6 +1864,53 @@ defmodule Explorer.Series do
   def quantile(%Series{dtype: dtype}, _),
     do: dtype_error("quantile/2", dtype, [:integer, :float, :date, :time, :datetime])
 
+  @doc """
+  Compute the sample skewness of a data set.
+
+  For normally distributed data, the skewness should be about zero. For unimodal continuous distributions, a skewness value greater than zero means that there is more weight in the right tail of the distribution. The function skewtest can be used to determine if the skewness value is close enough to zero, statistically speaking.
```
josevalim, line 1870:
```suggestion
  For normally distributed data, the skewness should be about zero.
  For unimodal continuous distributions, a skewness value greater
  than zero means that there is more weight in the right tail of the
  distribution. The function skewtest can be used to determine if
  the skewness value is close enough to zero, statistically speaking.
```

lib/explorer/series.ex
```diff
@@ -1864,6 +1864,53 @@ defmodule Explorer.Series do
...
+
+  See scipy.stats for more information.
```
josevalim, line 1872: We should not point to scipy :D

lib/explorer/series.ex
```diff
@@ -1864,6 +1864,57 @@ defmodule Explorer.Series do
   def quantile(%Series{dtype: dtype}, _),
     do: dtype_error("quantile/2", dtype, [:integer, :float, :date, :time, :datetime])
 
+  @doc """
+  Compute the sample skewness of a data set.
```
josevalim, line 1868:
```suggestion
  Compute the sample skewness of a series.
```

lib/explorer/series.ex
```diff
@@ -1864,6 +1864,57 @@ defmodule Explorer.Series do
...
+
+  For normally distributed data, the skewness should be about zero.
+  For unimodal continuous distributions, a skewness value greater
+  than zero means that there is more weight in the right tail of the
+  distribution. The function skewtest can be used to determine if
```
josevalim, line 1873: What is the `skewtest` function?

lib/explorer/series.ex
guarilha: haha my bad, copied the description from polars and forgot to edit it. I'll change it and the ref to skewtest.

lib/explorer/series.ex
guarilha: Removed the reference to `skewtest` as it seems to just be available on [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.skewtest.html#scipy.stats.skewtest).

josevalim:
Looking good now! Can you please look at the CI now? It seems we are missing `skew` in the lazy backend (and we need to add a test for it like in the previous PRs).

philss (approved):

josevalim:

PR #615, opened 2023-06-15
Traverse categories once and reuse terms in categorial series

### LINKED ISSUE #562: Improvements to categorical to series Rust implementation
1. Today we use a hash map but a vector would suffice
2. We build a binary for each category, instead, we should build the binary just once

The idea is to have a vector of Erlang binaries to access and return them directly. Creating the binary for each category only once.

thehabbos007:
Resolves #562 

This PR changes from a `HashMap` to a `Vector` when converting a categorical series to a list. 
The vector is as big as the the cardinality of the category, which means we're reusing the vector across the category indices.

I ran some very unscientific micro benchmarks with benchee and got the following results
```
Name                  ips        average  deviation         median         99th %
old_to_list        468.96        2.13 ms    ±10.91%        2.02 ms        2.71 ms
new_to_list        1.19 K      838.64 μs    ±19.62%      857.04 μs     1330.65 μs
```
Keeping in mind that the bench is very unscientific, the speedup seems reasonable!

As it's an internal change I expect the existing test suite to cover the cases of the changes.

<details><summary>Bench code</summary>

```elixir
alias Explorer.Series

gen_cat = fn x ->  for _ <- 1..x, into: "", do: <<Enum.random('0123456789abcdef')>> end

Benchee.run(
  %{
    "to_list" => fn series ->
      Series.to_list(series)
    end,
  },
  before_each: fn _ ->
      categories = Series.from_list(Enum.map(0..10_000, fn _ ->  gen_cat.(5) end), dtype: :category)
      indexes = Series.from_list(Enum.map(0..100_000, fn _ -> :rand.uniform(10_000 - 1) end))
      Series.categorise(indexes, categories)
  end
)
```

<details>

### DIFF
--- a/native/explorer/src/encoding.rs
+++ b/native/explorer/src/encoding.rs
@@ -359,14 +359,14 @@ fn categorical_series_to_list<'b>(
     let cat_size = mapping.len();
     let mut terms: Vec<NIF_TERM> = vec![nil_as_c_arg; cat_size];
 
-    for index in 0..cat_size {
+    for (index, term) in terms.iter_mut().enumerate() {
         if let Some(existing_str) = mapping.get_optional(index as u32) {
             let mut binary = NewBinary::new(env, existing_str.len());
             binary.copy_from_slice(existing_str.as_bytes());
 
             let binary_term: Term = binary.into();
 
-            terms[index] = binary_term.as_c_arg();
+            *term = binary_term.as_c_arg();
         }
     }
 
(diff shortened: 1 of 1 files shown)


philss:
@thehabbos007 awesome! :D 

Please run the formatter with `mix ci` from the root of the project, or `cargo fmt` inside crate path.

thehabbos007:
Thanks, @philss. Have applied the change :)

josevalim:

PR #616, opened 2023-06-15
Improve flake.nix

viniciusmuller:
This pull request fixes a deprecation message that is shown when entering the development shell and also adds `cmake` to it, as it is required when building the rust project


cigrainger (approved):
Thanks! Yeah they updated `fenix` recently.

PR #618, opened 2023-06-15
Add precompiled NIFs for freebsd

maltekrupa:
Hi,

as discussed in https://github.com/elixir-nx/explorer/issues/617, this will add precompiled NIFs for freebsd.

The solution was to use a more recent version of https://github.com/cross-rs/cross. Since this failed on `ubuntu-20.04`, the OS had to be set to a newer version too (`ubuntu-22.04`).

I left the other targets untouched but it'd probably make sense to update the OS for them too. Either here or in a new PR. What do you think?

Best
Malte


philss:
> I left the other targets untouched but it'd probably make sense to update the OS for them too. Either here or in a new PR. 

I would keep the `ubuntu-20.04` for all the other targets for compatibility issues. If we build in a newer OS, it may brake systems running on older versions.

@maltekrupa another important thing is that we may need to change [`rustler_precompiled`](https://github.com/philss/rustler_precompiled) to normalize the target name. Your error message got the target ` "amd64-portbld-freebsd13.1"`, but this name should match the target name we are building to.

maltekrupa:
Thank you for the quick response.

> I would keep the ubuntu-20.04 for all the other targets for compatibility issues. If we build in a newer OS, it may brake systems running on older versions.

Ok, what about the other targets that use cross? Should they stay on the old cross+ubuntu version too?

> another important thing is that we may need to change [rustler_precompiled](https://github.com/philss/rustler_precompiled) to normalize the target name

Cool, I'll check that next.

philss:
@maltekrupa I think we can keep the newer version of cross, unless it's not compatible with most of the targets.

philss:
@maltekrupa thanks for the updates! I think we need one more thing: add the target to the list of targets here: 
https://github.com/elixir-nx/explorer/blob/2005b9048c184a93023cbe0d95a690f22725300f/lib/explorer/polars_backend/native.ex#L15

And also, I decided to not force this new target as default for now (in `rustler_precompiled`), so it makes easier for maintainers to decide to adopt or not.

maltekrupa:
> thanks for the updates! 

Thank you for taking the time to guide me through everything!

> I think we need one more thing: add the target to the list of targets here:

Done 👍 

> And also, I decided to not force this new target as default for now (in rustler_precompiled), so it makes easier for maintainers to decide to adopt or not.

Sounds good to me. :)

leonardorame:
Hi, I'm trying to use Explorer 0.8.2 on FreeBSD 14 (in a LiveBook) and I'm getting this error:

```
[warning] The on_load function for module Elixir.Explorer.PolarsBackend.Native returned:
{:error,
 {:load_failed,
  ~c"Failed to load NIF library: '/usr/home/leonardo/.cache/mix/installs/elixir-1.16.2-erts-14.2.3/5f580a480fb9239855394fbd02c8fe03/_build/dev/lib/explorer/priv/native/libexplorer-v0.8.2-nif-2.15-x86_64-unknown-freebsd.so: No space available for static Thread Local Storage'"}}
```

Is the `libexplorer-v0.8.2-nif-2.15-x86_64-unknown-freebsd.so` compiled for FreeBSD 14 or for an older version?.

Issue #603, opened 2023-05-28
Support relocating columns


josevalim:
Reference: https://dplyr.tidyverse.org/reference/relocate.html

PR #619, opened 2023-06-15
Implement `relocate` inspired by dplyr

thehabbos007:
My attempt at resolving #603

I've added tests and an implementation for Polars (non-lazy). Any suggestions and pointers are welcome :)


native/explorer/src/dataframe.rs
```diff
@@ -455,6 +455,35 @@ pub fn df_from_series(columns: Vec<ExSeries>) -> Result<ExDataFrame, ExplorerErr
...
+        .partition(|series| !columns.contains(&series.name()));
+
+    let (second_series, second_to_relocate): (Vec<Series>, Vec<Series>) = second_series
+        .into_iter()
+        .partition(|series| !columns.contains(&series.name()));
+
+    let series = first_series
+        .into_iter()
+        .chain(first_to_relocate.into_iter())
+        .chain(second_to_relocate.into_iter())
```
thehabbos007, line 478: This may be wrong, actually. This wouldn't re-insert the columns in the correct order
```elixir
df = Explorer.DataFrame.new(col_a: [1, 2], col_b: [5.1, 5.2], col_c: [4,5], col_d: ["yes", "no"], col_e: [4, 1])
#  .. 
df2 = Explorer.DataFrame.relocate(df, [4, 0], before: 2)
> #Explorer.DataFrame<
  Polars[2 x 5]
  col_b float [5.1, 5.2]
  col_e integer [4, 1]
  col_a integer [1, 2]
  col_c integer [4, 5]
  col_d string ["yes", "no"]
>
Explorer.DataFrame.dump_csv(df2)
> {:ok, "col_b,col_a,col_e,col_c,col_d\n5.1,1,4,4,yes\n5.2,2,1,5,no\n"}
```
So while the implementation in Elixir orders it appropriately, it isn't ordered correctly here for the underlying dataframe

thehabbos007: Okay, should be addressed in 234c61c1d551c3146b766dd7b810b46017aedf84 :)

lib/explorer/data_frame.ex
```diff
@@ -2973,6 +2973,126 @@ defmodule Explorer.DataFrame do
...
+    out_df = %{df | names: new_names}
+
+    Shared.apply_impl(df, :relocate, [out_df, columns, col_index])
+  end
+
+  defp relocate_columns({direction, :first}, df, columns_to_relocate),
+    do: relocate_columns({direction, 0}, df, columns_to_relocate)
+
+  defp relocate_columns({direction, :last}, df, columns_to_relocate),
+    do: relocate_columns({direction, -1}, df, columns_to_relocate)
```
thehabbos007, line 3072: I suppose this "special casing" is somewhat redundant when 0 and -1 can be provided as the index. Should I omit this in favor for consistency with other functions that don't use `:first` and `:last` in this way?

josevalim: I like the readability of using :first | :last, so I would keep it. :)

josevalim: Actually, there is an ambiguity if you have a column name first or last, so we should probably remove it. :)

josevalim (commented):
Just one comment on first/last but it looks great to me, thank you! Also note that we have to support this operation on lazy data frames. :)

lib/explorer/data_frame.ex
thehabbos007: Yeah that's true. Although you'd be able to specify the column with a string instead. I'm leaning on removing it still, though

thehabbos007: Removed in 09d2bf718a11ee9d670135a3c62fe4646e25ba8c, will look at the lazy data frames now!

lib/explorer/data_frame.ex
```diff
@@ -2973,6 +2973,112 @@ defmodule Explorer.DataFrame do
...
+  Relocates columns.
+
+  Change column order within a DataFrame. The `before` and `after` options are mutually exclusive.
+  Providing no options will relocate the columns to beginning of the DataFrame.
+
+  ## Options
+
+    * `:before` - Specifies to relocate before the given column.
+
+    * `:after` - Specifies to relocate after the given column.
```
josevalim, line 2986:
```suggestion
    * `:before` - Specifies to relocate before the given column.
      It can be an index or a column name.

    * `:after` - Specifies to relocate after the given column.
      It can be an index or a column name.
```

josevalim:
Great call on using `select`! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

thehabbos007:
@josevalim I was about to add a couple test cases for `lazy` relocation, do you want me to open a separate PR with those? :)

josevalim:
Feel free to submit a PR, but I would only add a simple smoke test, given it is built on top of select. :)

PR #620, opened 2023-06-16
Add lazy data frame test case for `relocate`

thehabbos007:
Sorry, should have pushed the changes in bulk!

I also typo'd the arity, so just drive-by changing that. Thank you!


josevalim:

PR #621, opened 2023-06-17
Add rank to Series

guarilha:
Also found a bug when using with skew on lazy series and added a new test case for it. 

ref: https://pola-rs.github.io/polars/py-polars/html/reference/series/api/polars.Series.rank.html


josevalim (approved):

philss (approved):
Awesome! :rocket: 

Thank you! :purple_heart:

PR #623, opened 2023-06-18
Update the way Rust returns errors

Jhonatannunessilva:
This PR resolves https://github.com/elixir-nx/explorer/issues/584

After this PR, all errors returned from Rust will be strings instead of tuples

### DIFF
--- a/lib/explorer/polars_backend/shared.ex
+++ b/lib/explorer/polars_backend/shared.ex
@@ -150,7 +150,6 @@ defmodule Explorer.PolarsBackend.Shared do
   def internal_from_dtype(:string), do: "str"
 
   defp error_message(error) when is_binary(error), do: error
-  defp error_message(error), do: inspect(error)
 
   def parquet_compression(nil, _), do: :uncompressed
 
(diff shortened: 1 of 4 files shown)


lib/explorer/polars_backend/shared.ex
```diff
@@ -149,7 +149,7 @@ defmodule Explorer.PolarsBackend.Shared do
   def internal_from_dtype(:integer), do: "i64"
   def internal_from_dtype(:string), do: "str"
 
-  defp error_message({_err_type, error}) when is_binary(error), do: error
+  defp error_message(error) when is_binary(error), do: error
   defp error_message(error), do: inspect(error)
```
josevalim, line 153: Will this branch ever be called?

josevalim (approved):

lib/explorer/polars_backend/shared.ex
Jhonatannunessilva: I don't think so, but if I remove this branch the function will be:

```elixir
defp error_message(error), do: error
```

So is it better to remove it?

josevalim: Remove the branch but keep the guard in the previous one?

Jhonatannunessilva: https://github.com/elixir-nx/explorer/pull/623/commits/cafb699044a648a8a3d07c650cecaafc309d6514

philss:
yey! :tada: 
thank you, @Jhonatannunessilva! :purple_heart:

PR #648, opened 2023-07-10
ADBC support in from_query

josevalim:


native/explorer/Cargo.toml
```diff
@@ -73,3 +73,6 @@ version = "0.29"
 
 [dependencies.polars-algo]
 version = "0.29"
+
+[patch.crates-io]
+arrow2 = { git = "https://github.com/jorgecarleitao/arrow2.git" }
```
philss, line 78: I think we may want to remove this patch, since a new version was launched. But I'm not sure if the fix we want is there.

lib/explorer/data_frame.ex
```diff
@@ -391,6 +392,95 @@ defmodule Explorer.DataFrame do
...
+  def from_query(conn, query, params, opts \\ [])
+
+  if Code.ensure_loaded?(Adbc) do
+    def from_query(conn, query, params, opts)
+        when is_binary(query) and is_list(params) and is_list(opts) do
+      backend = backend_from_options!(opts)
+      backend.from_query(conn, query, params)
+    end
+  else
+    def from_query(conn, query, params, opts) do
```
josevalim, line 461:
```suggestion
    def from_query(_conn, _query, _params, _opts) do
```

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -20,6 +20,18 @@ defmodule Explorer.PolarsBackend.DataFrame do
 
   # IO
 
+  @impl true
```
josevalim, line 23:
```suggestion
  @compile {:no_warn_undefined, Adbc.Connection}
  @impl true
```

philss (approved):
Awesome work! :heart: :rocket:

lib/explorer/data_frame.ex
```diff
@@ -96,6 +96,7 @@ defmodule Explorer.DataFrame do
   - [Arrow IPC](https://arrow.apache.org/docs/format/Columnar.html#ipc-file-format)
   - [Arrow Streaming IPC](https://arrow.apache.org/docs/format/Columnar.html#ipc-streaming-format)
   - [Newline Delimited JSON](http://ndjson.org)
+  - Databases via `ADBC` in `from_query/3`
```
josevalim, line 99:
```suggestion
  - Databases via `Adbc` in `from_query/3`
```

native/explorer/Cargo.toml
josevalim: I checked and it has not been released yet. :(

lib/explorer/data_frame.ex
```diff
@@ -391,6 +392,95 @@ defmodule Explorer.DataFrame do
...
+  in your supervision tree.
+
+  First, add `:adbc` as a dependency in your `mix.exs`:
+
+      {:adbc, "~> 0.1"}
+
+  Now, in your config/config.exs, configure the drivers you are going to use
+  (see `Adbc` module docs for more information on supported drivers):
+
+      config :adbc, :drivers, [:sqlite3]
```
sasikumar87, line 414:
```suggestion
      config :adbc, :drivers, [:sqlite]
```

josevalim:

PR #645, opened 2023-07-07
Starting FSS abstraction

philss:
This is work-in-progress.

PR #649, opened 2023-07-12
Add FSS abstraction to remaining `from_*` IO functions

philss:
This is a continuation of #645 

PS: I left out the introduction of `endpoint / base_url` to another PR.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1061,16 +1061,23 @@ defmodule Explorer.DataFrame do
 
     * `:lazy` - force the results into the lazy version of the current backend.
 
+    * `:config` - An optional config struct or map, normally associated with remote
+      file systems. See [IO section](#module-io-operations) for more details. (default: `nil`)
+
   """
   @doc type: :io
-  @spec from_ipc_stream(filename :: String.t()) :: {:ok, DataFrame.t()} | {:error, term()}
+  @spec from_ipc_stream(filename :: String.t() | fs_entry()) ::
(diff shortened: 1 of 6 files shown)


lib/explorer/data_frame.ex
```diff
@@ -682,34 +690,40 @@ defmodule Explorer.DataFrame do
...
+    {:ok, %S3.Entry{url: entry, config: config}}
   end
 
-  defp normalise_entry("file://" <> path, _config), do: %Local.Entry{path: path}
+  defp normalise_entry("file://" <> path, _config), do: {:ok, %Local.Entry{path: path}}
 
   defp normalise_entry(filepath, _config) when is_binary(filepath) do
     if File.exists?(filepath) do
-      %Local.Entry{path: filepath}
+      {:ok, %Local.Entry{path: filepath}}
```
josevalim, line 721: What if we always return it as `%Local.Entry{path: filepath}`? Then if it doesn't exist, Polars will fail to read anyway and say the file does not exist. This way we don't need to fiddle with ArgumentError below. Thoughts?

josevalim (approved):
One tiny comment and ship it!

lib/explorer/data_frame.ex
philss: But you mean, remove the result from the `normalise_entry/2` entirely? Or just always return `{:ok, %Local.Entry{path: filepath}}` for this function clause?

josevalim: Don't check if the file exists, alwaus return `{:ok, ...}` for the function clause.

PR #650, opened 2023-07-12
Fix citation for UCI datasets (wine and iris)

firefly-cpp:
I recommend reconciling citations for UCI datasets according to the guidelines provided by [UCI](https://archive.ics.uci.edu/citation). Hence, accompanying a DOI for each dataset would also be very nice.


josevalim:

Issue #646, opened 2023-07-10
Normalise errors to return `Exception.t()` from the backend


philss:
The idea is to return something like `{:error, Exception.t()`} instead of `{:error, String.t()}` that we have today.

josevalim:
We will simply call `RuntimeError.message(...)` to convert them. :)

Jhonatannunessilva:
I don't know if I understood correctly, but the idea is to change this:

https://github.com/elixir-explorer/explorer/blob/e29d77e0b2a3b27b9dfd5375ea5ef64f0b0320b5/lib/explorer/polars_backend/shared.ex#L152

for this?:

```elixir
defp error_message(error) when is_binary(error), do: RuntimeError.message(%{message: error})
```

josevalim:
Yes, more precisely, you can do:

    defp error_message(error) when is_binary(error), do: RuntimeError.message(error)

Jhonatannunessilva:
Using `RuntimeError.message(error)` returns this error:

```elixir
iex(1)> RuntimeError.message("test")
** (KeyError) key :message not found in: "test"

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (elixir 1.15.0) lib/exception.ex:801: RuntimeError.message/1
    iex:1: (file)
```

That's why I did it like this `RuntimeError.message(%{message: error})`

josevalim:
I am very sorry, I meant this: `RuntimeError.exception(error)`. The goal is to make it so it returns an Exception struct. We would also need to update the typespecs. :)

PR #651, opened 2023-07-15
Return exception instead of string

Jhonatannunessilva:
This PR resolves https://github.com/elixir-explorer/explorer/issues/646

After this PR, the error_message/1 function will return an exception instead of a string.


Jhonatannunessilva:
Which typespecs should change?

Because from what I saw the functions error_message/1, apply/2, apply_series/3, apply_dataframe/3, and apply_dataframe/4 do not have typespecs.

josevalim:
I think entries like this: https://github.com/elixir-explorer/explorer/blob/main/lib/explorer/data_frame.ex#L537

Should now say `{:error, Exception.t}`, no?

josevalim:

PR #652, opened 2023-07-18
Read parquet files from AWS S3

philss:
This feature enables reading [Parquet files](https://parquet.apache.org/) directly from services like the AWS S3.
It's also compatible with other services that implements the S3 API (like minIO, localstack, DO spaces, etc).

This is using [Polars'](https://github.com/pola-rs/polars) "scan_parquet" feature, that is lazy and also works for reading multiple files at once.

```elixir
Explorer.DataFrame.from_parquet("s3://test-bucket/*.parquet")
```


lib/fss.ex
```diff
@@ -13,19 +13,99 @@ defmodule FSS do
 
   defmodule S3 do
     defmodule Config do
-      defstruct [:region, :access_key_id, :secret_access_key]
+      defstruct [
+        :access_key_id,
+        :region,
+        :secret_access_key,
+        {:endpoint, "amazonaws.com"},
```
wojtekmach, line 20: do you foresee supporting different endpoints for S3-compatible services?

lib/fss.ex
```diff
@@ -13,19 +13,99 @@ defmodule FSS do
...
+                {:error, "host is required"}
+
+              %{path: nil} ->
+                {:error, "path to the resource is required"}
+            end
+
+          uri.scheme == "http" or uri.scheme == "https" ->
+            parts =
+              case String.split(uri.host, ".") do
+                ["s3", region | tail] ->
```
jonatanklosko, line 72: So this is specific to AWS, is it specific to AWS on Polars side, or could we support other compatible storages?

cigrainger: Good point. We use Minio and it'd be great to be able to use it with this.

Qqwy: Looks like Polars itself only supports the s3://-style URLs without specifying the region in the URL (source and the related CloudLocation and CloudType structs). (You can pass the region as a CloudOption though).

The 'virtual-hosted style' URLs seem to be AWS-specific as far as I can find. Some other storages (including Minio) have opt-in support for them, though the exact format of the resulting URLs might be too different to cover with this particular pattern-match.

lib/fss.ex
```diff
@@ -13,19 +13,102 @@ defmodule FSS do
...
+                 bucket: parts[:bucket],
+                 key: parts[:key],
+                 port: uri.port,
+                 config: config
+               }}
+            end
 
-      @type t :: %__MODULE__{url: String.t(), config: Config.t()}
+          true ->
+            {:error, :not_supported}
```
josevalim, line 109:
```suggestion
            {:error, :invalid_format}
```

lib/fss.ex
philss: Since Polars only supports the `s3://` version - thanks for checking that, @Qqwy ! -, I'm going to remove the handle of `http(s)` URLs for now.

lib/fss.ex
philss: I added a message for this error. WDYT?
https://github.com/elixir-explorer/explorer/pull/652/files#diff-0cb4ec47f6ec5659387509052e0673b6204edecf2e060b49a141551c6eb7b22bR72

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -146,24 +146,31 @@ defmodule Explorer.PolarsBackend.LazyFrame do
   defp char_byte(<<char::utf8>>), do: char
 
   @impl true
-  def from_parquet(%S3.Entry{}, _max_rows, _columns) do
-    raise "S3 is not supported yet"
+  def from_parquet(%S3.Entry{} = entry, max_rows, columns) do
+    if columns, do: raise_columns_presense_error()
```
josevalim, line 150:
```suggestion
    if columns, do: raise_columns_presence_error()
```

lib/fss.ex
```diff
@@ -13,19 +13,65 @@ defmodule FSS do
...
+
+            %{host: nil} ->
+              {:error, "host is required"}
 
-      @type t :: %__MODULE__{url: String.t(), config: Config.t()}
+            %{path: nil} ->
+              {:error, "path to the resource is required"}
+          end
+        else
+          {:error, "only s3:// URIs are supported for now"}
```
josevalim, line 72: What if we use `ArgumentError.exception("only s3:// URIs are supported for now")` and we also start adhering to `{:error, Exception.t}` here too?

lib/explorer/data_frame.ex
```diff
@@ -711,7 +711,7 @@ defmodule Explorer.DataFrame do
 
   defp normalise_entry("s3://" <> _rest = entry, config) do
     config = s3_config(config)
```
josevalim, line 713: Now that we have a S3 config from env, should we lazy load it from the environment?

lib/fss.ex
```diff
@@ -13,19 +13,65 @@ defmodule FSS do
...
+              config: Config.t()
+            }
+
+      def config_from_system_env() do
+        %FSS.S3.Config{
+          access_key_id: System.get_env("AWS_ACCESS_KEY_ID"),
+          secret_access_key: System.get_env("AWS_SECRET_ACCESS_KEY"),
+          region: System.get_env("AWS_REGION", System.get_env("AWS_DEFAULT_REGION")),
+          token: System.get_env("AWS_SESSION_TOKEN")
+        }
```
josevalim, line 49: What happens if any of those env vars are missing? Should we try to raise if any of them is missing?

native/explorer/src/datatypes.rs
```diff
@@ -400,3 +404,51 @@ impl TryFrom<ExParquetCompression> for ParquetCompression {
         Ok(compression)
     }
 }
+
+// =========================
+// ====== FSS Structs ======
+// =========================
+
+#[derive(NifStruct, Clone, Debug)]
```
josevalim, line 412: Beautiful.

lib/fss.ex
```diff
@@ -13,19 +13,65 @@ defmodule FSS do
...
+      end
+
+      def parse(url, opts \\ []) do
+        opts =
+          opts
+          |> Keyword.validate!([:config])
+          |> Keyword.put_new_lazy(:config, fn -> config_from_system_env() end)
```
josevalim, line 56: here is one idea:

```suggestion
        config =
          opts
          |> Keyword.validate!([config: nil])
          |> Keyword.fetch!(:config)
          |> case do
            nil -> config_from_system_env()
            config when is_list(config) -> struct!(config_from_system_env(), config)
            %Config{} -> config
          end
          |> validate_config!() # check there are no nils
```

This way we allow no config to be set, %Config{}, or a keyword list to update config. Then we validate no field is missing due to env vars (or default values). WDYT?

lib/fss.ex
philss: Yeah, agreed: [174094a](https://github.com/elixir-explorer/explorer/pull/652/commits/174094a3c270fb2873d1b432daaf76d3e7f07231)

lib/explorer/data_frame.ex
philss: I removed this function and delegate all the config to the `parse/2` function.

lib/fss.ex
philss: We should raise now :boom:

lib/fss.ex
philss: Implemented in: [174094a](https://github.com/elixir-explorer/explorer/pull/652/commits/174094a3c270fb2873d1b432daaf76d3e7f07231)
Thanks! :purple_heart:

philss:
I'm having a bad time implementing the integration test using minIO. It seams that Polars (or Reqwest) does not accept connections that are not using TLS - only HTTPs endpoints are accepted.

My first attempt to have localhost certs was to generate them with [mkcert](https://github.com/FiloSottile/mkcert).
I exposed the certs to the minio server, but it didn't work for me :/

Then I tested using a [ngrok](https://ngrok.com) tunnel - that uses TLS - pointing to my minIO instance on localhost and it works :tada:

Unfortunately this is not a valid solution, so I'm considering running a minIO instance in a private server, or use the AWS S3 service, but with instructions on how to reproduce the environment.

If any of you have ideas, please let me know!

josevalim:
Did you try calling this in the object_store configuration? https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.with_allow_http

The docs also mention localstack testing, so it may be worth looking at object_store own tests in case they use something similar?

lib/fss.ex
```diff
@@ -13,19 +13,95 @@ defmodule FSS do
...
+
+            %{host: nil} ->
+              {:error, ArgumentError.exception("host is required")}
+
+            %{path: nil} ->
+              {:error, ArgumentError.exception("path to the resource is required")}
+          end
+        else
+          {:error, ArgumentError.exception("only s3:// URIs are supported for now")}
+        end
```
wojtekmach, line 90: I think you can remove some nesting if you push this scenario up,

```elixir
case uri do
  %{scheme: "s3", host: host, ...} ->
    ...

  %{scheme: scheme} when scheme != "s3" ->
    ...

  %{host: nil} ->
    ...

  ...
end
```

it's a bit redundant, but I think less nesting would be a win.

Also, nitpick:

> ```elixir
> %{host: nil} ->
>   {:error, ArgumentError.exception("host is required")}
> ```

while technically it is _host_, conceptually it is _bucket_. Maybe worth changing?

This gave me an idea, I think it could simply be:

```elixir
case URI.parse(url) do
  %{scheme: "s3", host: host, path: "/" <> key} when is_binary(host) ->
    ...

  url ->
    {:error,
     ArgumentError.exception("expected s3://<bucket>/<key> URL, got: " <> URI.to_string(url))}
end
```

but I might be trying too hard, I think it's fine either way. :)

Qqwy: > while technically it is host, conceptually it is bucket. Maybe worth changing?

Maybe it makes sense to follow the field naming which Polars itself is using for this?
https://github.com/pola-rs/polars/blob/main/polars/polars-io/src/cloud/glob.rs#L76-L85

philss: @wojtekmach @Qqwy thanks for the suggestions! I applied most of them :)

I think I'm following more or less the naming which Polars is using, but I think they have more concepts that don't apply here.

philss:
> Did you try calling this in the object_store configuration? https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.with_allow_http

> The docs also mention localstack testing, so it may be worth looking at object_store own tests in case they use something similar?

@josevalim thank you! I didn't try this config. I cannot access the internal builder that this method is from, so I'm trying the way of using the `(key, value)` configs, like we are doing today. But for that, I need to [update Polars to v0.31](https://github.com/elixir-explorer/explorer/pull/659), since the key pair for allowing http is only available after that version (they updated object_store, so we can use).

philss:
@josevalim @wojtekmach @Qqwy @jonatanklosko I think it's ready for another pass, if you can review again :)

PS: sorry for the amount of failed builds :sweat:

lib/explorer/data_frame.ex
```diff
@@ -107,11 +107,18 @@ defmodule Explorer.DataFrame do
       # path to a file in disk
       Explorer.DataFrame.from_parquet("/path/to/file.parquet")
 
-      # path to a URL schema (with optional configuration) (coming soon)
-      Explorer.DataFrame.from_parquet("s3://bucket/file.parquet", config: FSS.S3.config(credentials))
+      # path to a URL schema (with optional configuration)
+      Explorer.DataFrame.from_parquet("s3://bucket/file.parquet", config: FSS.S3.Entry.config_from_system_env())
```
josevalim, line 111: Maybe something got lost in translation but this should be this (functions should be in FSS.S3):

```suggestion
      Explorer.DataFrame.from_parquet("s3://bucket/file.parquet", config: FSS.S3.config_from_system_env())
```

But in practice, `config` is optional.

lib/fss.ex
```diff
@@ -13,19 +13,91 @@ defmodule FSS do
...
+      defstruct [:bucket, :key, :port, :config]
+
+      @type t :: %__MODULE__{
+              bucket: String.t(),
+              key: String.t(),
+              port: pos_integer(),
+              config: Config.t()
+            }
+
+      def config_from_system_env() do
```
josevalim, line 43: Just for clarify: this functijon should be in FSS.S3 and not FSS.S3.Entry.

lib/explorer/data_frame.ex
```diff
@@ -107,11 +107,18 @@ defmodule Explorer.DataFrame do
...
 
-      # path to a filesystem specification entry (coming soon)
-      Explorer.DataFrame.from_parquet(FSS.S3.entry("s3://bucket/file.parquet", credentials))
+      # it's possible to configure using keyword lists
+      Explorer.DataFrame.from_parquet("s3://bucket/file.parquet", config: [access_key_id: "my-key", secret_access_key: "my-secret"])
+
+      # a FSS entry with its config inside
+      Explorer.DataFrame.from_parquet(%FSS.S3.Entry{config: %FSS.S3.Config{}})
```
josevalim, line 117:
```suggestion
      # a FSS entry (it already includes its config)
      Explorer.DataFrame.from_parquet(FSS.S3.parse("s3://bucket/file.parquet"))
```

lib/fss.ex
```diff
@@ -13,19 +13,91 @@ defmodule FSS do
...
+
+      defp validate_config!(%Config{} = config) do
+        access = config.access_key_id
+        secret = config.secret_access_key
+        region = config.region
+
+        if is_nil(access) or is_nil(secret) or is_nil(region) or
+             access == "" or secret == "" or region == "" do
+          raise "missing configuration keys or region to access the S3 API"
+        end
```
josevalim, line 97:
```suggestion
        check!(config, :access_key_id, "AWS_...")
        check!(config, :secret_access_key, "AWS_...")
        check!(config, :region, "AWS_...")
```

where check is:

```elixir
def check!(config, key, env) do
  if Map.fetch!(config, key) in ["", nil] do
    raise ArgumentError, "missing #{inspect(key)} for FSS.S3 (set the key or the #{env} env var)"
  end
end
```

josevalim (approved):
Amazing work, just some minor final nits. :) Also please add to the README how to run the cloud tests. :)

Issue #337, opened 2022-09-08
Explorer.Series.min(left, right)


petrkozorezov:
I'm trying to find minimum between two columns, and was surprised that there is no such function (like Nx.min/2).
```elixir
alias Explorer.DataFrame, as: EDF
alias Explorer.Series, as: ES

EDF.new(a: [1, 2, 3], b: [3, 2, 1])
|> EDF.mutate(df, [c: &ES.min(&1[:a], &1[:b])])
```
Maybe I do not see something and there is a better way to achieve this than filter + join?
```elixir
EDF.join(
  EDF.filter(df, &ES.less_equal(&1[:a], &1[:b])) |> EDF.mutate(min: &(&1[:a])),
  EDF.filter(df, &ES.less_equal(&1[:b], &1[:a])) |> EDF.mutate(min: &(&1[:b])),
  how: :outer
)
```

josevalim:
@petrkozorezov it seems we need to expose the underlying operation from polars, based on min expression here: https://stackoverflow.com/questions/69888178/calculating-rowwise-minimum-of-two-series

philss:
I think we can use the new `Explorer.Series.select/3` to perform this:

```elixir
require Explorer.DataFrame, as: EDF
alias Explorer.Series, as: ES

df = EDF.new(a: [1, 2, 3], b: [3, 2, 1])
EDF.mutate(df, [c: select(a < b, a, b)])
```

sasikumar87:
@philss @josevalim Is there a way to handle `nil` values in comparison?

```elixir
require Explorer.DataFrame, as: EDF
alias Explorer.Series, as: ES

df = EDF.new(a: [1, 2, nil], b: [3, 2, 1])
EDF.mutate(df, [c: select(a < b, a, b)])
```
gives
```
#Explorer.DataFrame<
  Polars[3 x 3]
  a integer [1, 2, nil]
  b integer [3, 2, 1]
  c integer [1, 2, 1]
>
```

I expect any comparison with nil to return nil .

philss:
@sasikumar87 I'm not sure yet how to handle this. But I think you are correct. If predicate is `nil`, we may want to return `nil`. 

For a quick work around, please try the following:

```elixir
EDF.mutate(df, [c: select(is_nil(a) or is_nil(b), select(is_nil(a), a, b), select(a < b, a, b))])
```

josevalim:
Is this a polars behavior? Maybe it is worth bringing this up there?

PR #658, opened 2023-07-19
Test case for select/3 with nil in predicate

sasikumar87:
@philss Added this test case, per our discussion at https://github.com/elixir-explorer/explorer/issues/337#issuecomment-1341694303 .

I also found https://github.com/pola-rs/polars/pull/9142 in polars 0.31 . I am new to rust, but seemed this could fix the issue. Adding this draft PR to see if this issue is really fixed once we upgrade to polars version 0.31 .

### DIFF
--- a/test/explorer/series_test.exs
+++ b/test/explorer/series_test.exs
@@ -2794,6 +2794,15 @@ defmodule Explorer.SeriesTest do
       s3 = Series.select(Series.less_equal(s, 2), -1, s)
       assert Series.to_list(s3) == [-1, -1, 3]
     end
+
+    test "select handles nil, nil on predicate returns nil" do
+      s1 = Series.from_list([1, 2, 3, 4, nil])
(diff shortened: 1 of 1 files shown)


sasikumar87:
@philss nil values are ignored on purpose it seems https://github.com/pola-rs/polars/blob/main/polars/polars-core/src/chunked_array/ops/zip.rs#L16-L17 .  WDYT?

josevalim:
I think it is fine to treat `nil`s as false and it would also mirror Elixir semantics, so we can follow Polars here. :)

sasikumar87:
@josevalim I agree to treat nil as false by default. But from developers perspective, we would have more easy control, if there is an option to override this behaviour in select/3. What do you think?

josevalim:
People can always handle it on top by doing `select(is_nil(a) or is_nil(b), nil, result)` but I don't think we need to handle it by default. :)

PR #659, opened 2023-07-19
Update Polars to v0.31

philss:
With the update, we had to also bump the nightly version that we use to the same version Polars is using.

Some tests were updated to reflect changes in Polars. I chose to copy some of the examples they provide to make sure our results match their documentation.


native/explorer/src/series.rs
```diff
@@ -297,10 +302,15 @@ pub fn s_argsort(
     descending: bool,
     nulls_last: bool,
 ) -> Result<ExSeries, ExplorerError> {
+    // Make this an option
```
josevalim, line 305:
```suggestion
    // TODO: Make this an option
```

native/explorer/src/series.rs
```diff
@@ -283,10 +283,15 @@ pub fn s_sort(
...
```
josevalim, line 286:
```suggestion
    // TODO: Make this an option
```

native/explorer/src/series.rs
```diff
@@ -331,43 +341,63 @@ pub fn s_frequencies(series: ExSeries) -> Result<ExDataFrame, ExplorerError> {
...
 #[rustler::nif(schedule = "DirtyCpu")]
 pub fn s_qcut(
     series: ExSeries,
     quantiles: Vec<f64>,
-    labels: Option<Vec<&str>>,
+    labels: Option<Vec<String>>,
     break_point_label: Option<&str>,
     category_label: Option<&str>,
-    maintain_order: bool,
+    _maintain_order: bool,
```
josevalim, line 375: It seems this is no longer an option right, right? So we need to remove it from the API here and in the function above.

josevalim (approved):
Awesome, minor comments and ship it :)

PR #661, opened 2023-07-21
Add necessary cargo features

philss:
These features are useful to control which NIF version we are going to compile
the project to, and which paths are going to be disabled due to incompatibility
with some targets we support.

This is using most of the work that is present in https://github.com/philss/rustler-precompiled-action/pull/4
But with an additional "cargo-args" that is experimental.


native/explorer/src/lazyframe/io.rs
```diff
@@ -54,6 +55,18 @@ pub fn lf_from_parquet_cloud(
     Ok(ExLazyFrame::new(lf))
 }
 
+#[cfg(not(feature = "aws"))]
```
josevalim, line 58: Why do we want to have this optional? Some targets do not support it as well?

josevalim (approved):

native/explorer/src/lazyframe/io.rs
philss: Yeah, the problem is that `object_store` has a dependency - called [ring](https://github.com/briansmith/ring) - that has some low-level code in C and Assembly which does not compile for RISCV computers. In my research I found that there were some attempts to fix the issue, but they didn't succeed. I was considering keep a fork and try some patches, but I think that would require some great amount of work.

philss:
I'm going to merge this one to fix the build, but I should go back to change the version once I release an official version of the `rustler-precompiled-action`.

Issue #663, opened 2023-07-25
Add binding to str.slice and/or str.split and/or add trim/2


DeemoONeill:
Currently doing data sanitizing (specifically strings) with explorer can be frustrating,

Currently the only string ops exposed are
    [contains/2](https://hexdocs.pm/explorer/Explorer.Series.html#contains/2)
    [downcase/1](https://hexdocs.pm/explorer/Explorer.Series.html#downcase/1)
    [trim/1](https://hexdocs.pm/explorer/Explorer.Series.html#trim/1)
    [trim_leading/1](https://hexdocs.pm/explorer/Explorer.Series.html#trim_leading/1)
    [trim_trailing/1](https://hexdocs.pm/explorer/Explorer.Series.html#trim_trailing/1)
    [upcase/1](https://hexdocs.pm/explorer/Explorer.Series.html#upcase/1)

It would be good if we could get slice or split into explorer, so things like trimming the currency symbol from financial data can be done in explorer rather than iterating over the data in elixir.

Another nice to have could be trim/2 which takes in which characters to trim (similar to the py-polars str.strip). Currently there is trim/1 which just does white space. 

I don't which ones go best with "constraining your options" philosophy, whilst still giving tools to do more effecient string manipulation.

If I could be tagged in any PR that comes from this, that would be great, i'd love to see how the bindings are added so I can help contribute in future.

josevalim:
Hi @DeemoONeill! Thanks for opening the issue. We could at least add trim/2, trim_leading/2 and trim_trailing/2 where the second argument is the string to trim. PRs from the overall community are very welcome!

Here is a template one can follow to add said functions: https://github.com/elixir-explorer/explorer/pull/627 :)

josevalim:
`slice/3` would also be very convenient, agreed! :+1:

DeemoONeill:
While working on trim, i think i've found a bug in the original implementation. the doctest is 

```elixir
      iex> s = Explorer.Series.from_list(["  abc", "def  ", "  bcd"])
      iex> Explorer.Series.trim(s)
      #Explorer.Series<
        Polars[3]
        string ["abc", "def", "bcd"]
      >
 ```
 
 and this passes.

Changing this to 
```elixir
      iex> s = Explorer.Series.from_list(["  abc", "def  ", "  bcd   "])
      iex> Explorer.Series.trim(s)
      #Explorer.Series<
        Polars[3]
        string ["abc", "def", "bcd"]
      >
```

[...]

PR #664, opened 2023-07-25
implement trim/2

DeemoONeill:
added trim/2 which allows passing a string to allow more than just whitespace to be trimmed from strings

fixed a bug in the trim implementation which only trimmed the leftmost whitespace originally (and only leftmost other patterns following changes)


lib/explorer/backend/series.ex
```diff
@@ -238,6 +238,7 @@ defmodule Explorer.Backend.Series do
   @callback upcase(s) :: s
   @callback downcase(s) :: s
   @callback trim(s) :: s
+  @callback trim(s, String.t()) :: s
```
philss, line 241: I think we can keep only this callback, and remove the `trim(s)`. For what I could understand, you always pass an empty `""` by default, right? WDYT?

lib/explorer/polars_backend/series.ex
```diff
@@ -616,7 +616,11 @@ defmodule Explorer.PolarsBackend.Series do
 
   @impl true
   def trim(series),
-    do: Shared.apply_series(series, :s_trim)
+    do: Shared.apply_series(series, :s_trim, [""])
```
philss, line 619: For completeness, we would remove this `trim/1` if we remove the callback.

lib/explorer/series.ex
```diff
@@ -4276,19 +4276,39 @@ defmodule Explorer.Series do
...
   """
   @doc type: :string_wise
   @spec trim(Series.t()) :: Series.t()
   def trim(%Series{dtype: :string} = series),
-    do: apply_series(series, :trim)
+    do: apply_series(series, :trim, [""])
+
+  @doc type: :string_wise
+  @spec trim(Series.t(), String.t()) :: Series.t()
+  def trim(%Series{dtype: :string} = series, string),
```
philss, line 4307: Here we could have a guard:

```suggestion
  def trim(%Series{dtype: :string} = series, string) when is_binary(string),
```

lib/explorer/series.ex
```diff
@@ -4276,19 +4276,39 @@ defmodule Explorer.Series do
...
+    do: apply_series(series, :trim, [string])
 
-  def trim(%Series{dtype: dtype}), do: dtype_error("trim/1", dtype, [:string])
+  # def trim(%Series{dtype: dtype}), do: dtype_error("trim/1", dtype, [:string])
```
philss, line 4310:
```suggestion
```

philss (commented):
LGTM! Only that suggestion of removing the `trim/1`.

lib/explorer/backend/series.ex
DeemoONeill: Ahh ok, I wasn't sure if i would break it.

I pass an empty string but is that the best way? 

I don't really see how else i could encode it on the rust side unless it was an enum with an atom type and a string type.

like 

```rust
enum OptionalString {
     Atom(TermType::Atom),
     Str(String)
}
```
(not sure if this would work)

 and then pass in `:nil` when there is no string passed?

lib/explorer/polars_backend/series.ex
DeemoONeill: ahh, i see. I'll give that a go

lib/explorer/backend/series.ex
philss: Oh, we can have this as an optional value as well.
The callback could be like `trim(s, String.t() | nil)` and we could have a signature like this in the Rust side:

```rust
pub fn expr_trim(expr: ExExpr, string: Option<String>) -> ExExpr
```

Then you can call the native function with `nil` or a string, that Rustler is going to auto convert it to an `Option<String>` (you don't need a custom type for that). WDYT?

DeemoONeill: That's exactly what i was hoping you would say 😆

DeemoONeill: That's implemented now

philss: Cool! Now just run `mix ci` to fix the issues and we are good to go!

lib/explorer/polars_backend/expression.ex
```diff
@@ -121,6 +121,7 @@ defmodule Explorer.PolarsBackend.Expression do
     # Strings
     contains: 2,
     trim: 1,
+    trim: 2,
```
philss, line 124: You can remove `trim: 1` from here as well.

lib/explorer/backend/series.ex
DeemoONeill: getting a load of doctest failures, probably due to being on windows and the difference between \r\n and \n.

I'll get it running in a linux environment tomorrow and sort the remaining issues

philss: It's fine :)
If this helps: I ran the command here and the diff necessary to fix the issues is the following: 

```diff
diff --git a/lib/explorer/backend/lazy_series.ex b/lib/explorer/backend/lazy_series.ex
index 5a488d10..c33b6b52 100644
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -113,7 +113,6 @@ defmodule Explorer.Backend.LazySeries do
     contains: 2,
     trim_leading: 1,
     trim_trailing: 1,
-    trim: 1,
     trim: 2,
     upcase: 1,
     downcase: 1,
diff --git a/lib/explorer/polars_backend/expression.ex b/lib/explorer/polars_backend/expression.ex
index 8e8aa781..d8c62c15 100644
--- a/lib/explorer/polars_backend/expression.ex
+++ b/lib/explorer/polars_backend/expression.ex
@@ -120,7 +120,6 @@ defmodule Explorer.PolarsBackend.Expression do
 
     # Strings
     contains: 2,
-    trim: 1,
     trim: 2,
     trim_leading: 1,
     trim_trailing: 1,
diff --git a/test/explorer/data_frame_test.exs b/test/explorer/data_frame_test.exs
index 7d6ba05a..60c8d1b5 100644
--- a/test/explorer/data_frame_test.exs
+++ b/test/explorer/data_frame_test.exs
@@ -1515,8 +1515,6 @@ defmodule Explorer.DataFrameTest do
     end
 
     test "trim characters from string" do
-      IO.puts("running trim characters")
-
       df =
         DF.new(
           a: ["£2", "3£", "£200£", "£££20"],
diff --git a/test/explorer/series_test.exs b/test/explorer/series_test.exs
index e140dadb..7225001d 100644
--- a/test/explorer/series_test.exs
+++ b/test/explorer/series_test.exs
@@ -3972,6 +3972,7 @@ defmodule Explorer.SeriesTest do
 
       assert Series.trim(series) |> Series.to_list() == ["123", "2", "20$"]
     end
+
     test "trim/2" do
       series = Series.from_list(["£1£23", "2£", "£20£"])
 
```

test/explorer/series_test.exs
```diff
@@ -3966,6 +3966,19 @@ defmodule Explorer.SeriesTest do
     end
   end
 
+  describe "trim/1 and trim/2" do
+    test "trim/1" do
+      series = Series.from_list(["  123   ", "       2   ", "    20$    "])
+
+      assert Series.trim(series) |> Series.to_list() == ["123", "2", "20$"]
+    end
```
josevalim, line 3974:
```suggestion
    end

```

lib/explorer/series.ex
```diff
@@ -4276,19 +4276,38 @@ defmodule Explorer.Series do
...
+      >
   """
   @doc type: :string_wise
   @spec trim(Series.t()) :: Series.t()
   def trim(%Series{dtype: :string} = series),
-    do: apply_series(series, :trim)
+    do: apply_series(series, :trim, [nil])
+
+  @doc type: :string_wise
+  @spec trim(Series.t(), String.t() | nil) :: Series.t()
```
josevalim, line 4306:
```suggestion
  @spec trim(Series.t(), String.t()) :: Series.t()
```

And we also need to provide documentation for this function. Elixir documents each name/arity pair. :) You can move the examples you added above to here. :)
AND WE AL

lib/explorer/backend/series.ex
```diff
@@ -237,7 +237,7 @@ defmodule Explorer.Backend.Series do
   @callback contains(s, String.t()) :: s
   @callback upcase(s) :: s
   @callback downcase(s) :: s
-  @callback trim(s) :: s
+  @callback trim(s, String.t()) :: s
```
josevalim, line 240:
```suggestion
  @callback trim(s, String.t() | nil) :: s
```

DeemoONeill:
I've made the changes, will get a docker container up after work and get `mix ci` run and make sure i've not broken any formatting

josevalim:
It is flawless, thank you @DeemoONeill! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #653, opened 2023-07-18
Example to dump a dataframe in CSV format to a S3-compatible object store

Qqwy:
This PR is a draft on how to dump a dataframe to a S3-compatible object store.
In this particular example, the CSV format is used; using JSON, Partque or IPC format should only require minimal changes.
Also, to make trying out easy, it currently uses the hard-coded `object_store::LocalFileSystem` as object store.

I've used the following to test the functionality manually for now:

```elixir
iex> mountains = Explorer.DataFrame.new(name: ["Everest", "K2", "Aconcagua"], elevation: [8848, 8611, 6962])
iex> Explorer.PolarsBackend.DataFrame.to_csv_writer_sample(mountains, "/tmp/mountains.csv", false, ",")
iex> File.read!("/tmp/mountains.csv")
"Everest,8848\nK2,8611\nAconcagua,6962\n"
```

Once properly integrated (main blocker: how to set up and pass in the configuration of the object store?), this could be turned into a full test.

PR #665, opened 2023-07-26
Write parquet from eager dataframe to S3

philss:
This is the first of the "write" operations using the abstraction from #653.
It implements the `to_parquet/3`.

I wanted to open this PR first to evaluate if the structure looks good.

### DIFF
--- a/native/explorer/Cargo.toml
+++ b/native/explorer/Cargo.toml
@@ -19,9 +19,11 @@ rand_pcg = "0.3"
 rustler = { version = "0.29", default-features = false, features = ["derive"] }
 thiserror = "1"
 smartstring = "1"
-tokio = { version = "1.29.1", features = ["rt"] }
-tokio-util = { version = "0.7.8", features = ["io", "io-util"] }
-object_store = "0.6.1"
+
+# Deps necessary for cloud features.
+tokio = { version = "1.29.1", default-features = false, features = ["rt"], optional = true }
+tokio-util = { version = "0.7.8", default-features = false, features = ["io", "io-util"], optional = true }
+object_store = { version = "0.6.1", default-features = false, optional = true }
 
 # MiMalloc won´t compile on Windows with the GCC compiler.
 # On Linux with Musl it won´t load correctly.
@@ -79,7 +81,7 @@ version = "0.31"
 [features]
 default = ["ndjson", "cloud", "nif_version_2_15"]
 
-cloud = ["aws"]
+cloud = ["object_store", "tokio", "tokio-util", "aws"]
 ndjson = ["polars/json"]
 aws = ["polars/async", "polars/aws"]
 
--- a/native/explorer/Cargo.toml
+++ b/native/explorer/Cargo.toml
@@ -21,9 +21,9 @@ thiserror = "1"
 smartstring = "1"
 
 # Deps necessary for cloud features.
-tokio = { version = "1.29.1", default-features = false, features = ["rt"], optional = true }
-tokio-util = { version = "0.7.8", default-features = false, features = ["io", "io-util"], optional = true }
-object_store = { version = "0.6.1", default-features = false, optional = true }
+tokio = { version = "1.29", default-features = false, features = ["rt"], optional = true }
+tokio-util = { version = "0.7", default-features = false, features = ["io", "io-util"], optional = true }
+object_store = { version = "0.6", default-features = false, optional = true }
 
 # MiMalloc won´t compile on Windows with the GCC compiler.
 # On Linux with Musl it won´t load correctly.
--- a/native/explorer/src/cloud_writer.rs
+++ b/native/explorer/src/cloud_writer.rs
@@ -29,6 +29,8 @@ impl CloudWriter {
     /// which bridges the sync writing process with the async ObjectStore multipart uploading.
     pub fn new(object_store: Box<dyn ObjectStore>, path: Path) -> Self {
         let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_time()
+            .enable_io()
             .build()
             .unwrap();
         let (multipart_id, writer) =
--- a/native/explorer/src/dataframe/io.rs
+++ b/native/explorer/src/dataframe/io.rs
@@ -515,6 +515,7 @@ pub fn df_load_ndjson(
     )))
 }
 
+#[cfg(feature = "cloud")]
 #[rustler::nif(schedule = "DirtyIo")]
 pub fn df_to_csv_writer_sample(
     data: ExDataFrame,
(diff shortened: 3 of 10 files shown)


philss:
cc @josevalim @Qqwy

native/explorer/Cargo.toml
```diff
@@ -19,9 +19,11 @@ rand_pcg = "0.3"
...
 thiserror = "1"
 smartstring = "1"
-tokio = { version = "1.29.1", features = ["rt"] }
-tokio-util = { version = "0.7.8", features = ["io", "io-util"] }
-object_store = "0.6.1"
+
+# Deps necessary for cloud features.
+tokio = { version = "1.29.1", default-features = false, features = ["rt"], optional = true }
+tokio-util = { version = "0.7.8", default-features = false, features = ["io", "io-util"], optional = true }
+object_store = { version = "0.6.1", default-features = false, optional = true }
```
josevalim, line 26:
```suggestion
tokio = { version = "1.29", default-features = false, features = ["rt"], optional = true }
tokio-util = { version = "0.7", default-features = false, features = ["io", "io-util"], optional = true }
object_store = { version = "0.6", default-features = false, optional = true }
```

test/explorer/data_frame/parquet_test.exs
```diff
@@ -227,4 +227,28 @@ defmodule Explorer.DataFrame.ParquetTest do
...
+        region: "us-east-1"
+      }
+
+      path = "s3://test-bucket/test-writes/wine-#{System.monotonic_time()}.parquet"
+
+      assert :ok = DF.to_parquet(df, path, config: config)
+
+      saved_df = DF.from_parquet!(path, config: config)
+
+      assert DF.to_columns(saved_df) == DF.to_columns(Explorer.Datasets.wine())
```
josevalim, line 251: Absolutely beautiful.

josevalim (approved):

native/explorer/Cargo.toml
```diff
@@ -79,7 +81,7 @@ version = "0.31"
 [features]
 default = ["ndjson", "cloud", "nif_version_2_15"]
 
-cloud = ["aws"]
+cloud = ["object_store", "tokio", "tokio-util", "aws"]
 ndjson = ["polars/json"]
 aws = ["polars/async", "polars/aws"]
```
Qqwy, line 86: Which way around do we want these features to depend?

Do we want the `cloud` feature to depend on all possible provider features (e.g. `aws`/`gcp`/`azure`/`...`), or have `cloud` be the 'base' feature and features like `aws` depend on it?

Polars itself does the former.

native/explorer/Cargo.toml
```diff
@@ -79,7 +81,7 @@ version = "0.31"
 [features]
 default = ["ndjson", "cloud", "nif_version_2_15"]
 
-cloud = ["aws"]
+cloud = ["object_store", "tokio", "tokio-util", "aws"]
```
Qqwy, line 84: Which way around do we want these features to depend?

Do we want the cloud feature to depend on all possible provider features (e.g. aws/gcp/azure/...), or have cloud be the 'base' feature and features like aws depend on it?

Polars itself does the former.

native/explorer/src/cloud_writer.rs
```diff
@@ -29,6 +29,8 @@ impl CloudWriter {
     /// which bridges the sync writing process with the async ObjectStore multipart uploading.
     pub fn new(object_store: Box<dyn ObjectStore>, path: Path) -> Self {
         let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_time()
+            .enable_io()
```
Qqwy, line 33: Great work! Apologies for missing these when I copied the example over from the Polars repository 🙈

native/explorer/src/dataframe/io.rs
```diff
@@ -230,6 +231,59 @@ pub fn df_to_parquet(
...
+
+#[cfg(not(feature = "aws"))]
+#[rustler::nif]
+pub fn df_to_parquet_cloud(
+    _data: ExDataFrame,
+    _ex_entry: ExS3Entry,
+    _ex_compression: ExParquetCompression,
+) -> Result<(), ExplorerError> {
+    Err(ExplorerError::Other(format!(
+        "AWS reads and writes are not enabled for this machine"
```
Qqwy, line 283: Not sure what we do in other cases, but maybe it is helpful to mention that Explorer itself is compiled without support for it on this machine? The current message seems rather vague to me, i.e. difficult to search for online, and not immediately clear from seeing the message what the cause is and how to fix it.

native/explorer/src/dataframe/io.rs
```diff
@@ -230,6 +231,59 @@ pub fn df_to_parquet(
...
+
+    let compression = ParquetCompression::try_from(ex_compression)?;
+
+    ParquetWriter::new(&mut cloud_writer)
+        .with_compression(compression)
+        .finish(&mut data.clone())?;
+    Ok(())
+}
+fn object_store_to_explorer_error(error: impl std::fmt::Debug) -> ExplorerError {
+    ExplorerError::Other(format!("Internal Object Store error: #{error:?}"))
```
Qqwy, line 272: I'd recommend writing `ObjectStore` here without the space in-between, to make it easier to search for online.

native/explorer/Cargo.toml
josevalim: I don't think we expect anyone to ever use them. It is only used internally to exclude those projects in certain targets.

Qqwy:
I know this abstraction was already built in the last PR, but I just want to mention that I really really love the abstraction of `FSS.Local.Entry` vs `FSS.S3.Entry` and how this is parsed and then pattern-matched on in `to_parquet`/`from_parquet`.
This is really Elixir at its finest, in my opinion! 🎉

native/explorer/Cargo.toml
philss: Yeah, it's only used to restrict some features to some targets (mostly, RISCV and ARM based targets).

But, since you mentioned, I think a better way to organize would be to have `cloud` as a base, and we have `aws` as the feature we enable / disable. Perhaps we could only have the `aws` feature for now - without `cloud`-, since we don't have any other cloud provider.  WDYT?

native/explorer/src/cloud_writer.rs
philss: No problem! :D
I couldn't understand why they are needed, but without them, we get panics.

native/explorer/src/dataframe/io.rs
philss: Makes sense! I'm going to improve the messages related to this.

native/explorer/Cargo.toml
philss: I think I won't change this for now. Going to postpone a little bit.

philss:
I'm merging this. If you have any more comments, I can fix from the main branch. Thanks!

PR #667, opened 2023-07-27
Use `nil_values` instead of `null_character` in IO operations

### LINKED ISSUE #666: `DataFrame.from_csv` incorrectly reads "NA" as `nil`
With the following CSV, the default `null_character` configuration of `"NA"` causes `Explorer.DataFrame.from_csv` to incorrectly read the abbreviation as `nil`.

```csv
name,abbreviation
Namibia,NA
```
ISO: https://www.iso.org/obp/ui/#iso:code:3166:NA

```
iex(1)> Explorer.DataFrame.from_csv("./tmp/sample.csv")
{:ok,
 #Explorer.DataFrame<
   Polars[1 x 2]
   name string ["Namibia"]
   abbreviation string [nil]
 >}
```

cnpryer:
Closes #666

~~Updates `null_character` defaults used in csv reading operations to use `nil` instead of `"NA"`.~~ See https://github.com/elixir-explorer/explorer/pull/667#discussion_r1275687546

Updates usage of `null_character` to `null_values` allowing multiple missing value interpretation in IO operations. This also changes the default from "NA" to `nil`.

Result:

```
iex(8)> Explorer.DataFrame.load_csv("name,abbreviation\nNamibia,NA")
{:ok,
 #Explorer.DataFrame<
   Polars[1 x 2]
   name string ["Namibia"]
   abbreviation string ["NA"]
 >}
```


notebooks/exploring_explorer.livemd
```diff
@@ -28,7 +28,7 @@ For CSV, your 'usual suspects' of options are available:
 * `dtypes` - A keyword list of `[column_name: dtype]`. If a type is not specified for a column, it is imputed from the first 1000 rows. (default: `[]`)
 * `header` - Does the file have a header of column names as the first row or not? (default: `true`)
 * `max_rows` - Maximum number of lines to read. (default: `nil`)
-* `null_character` - The string that should be interpreted as a nil value. (default: `"NA"`)
+* `null_character` - The string that should be interpreted as a nil value. (default: `nil`)
```
cnpryer, line 31: My only concern atm is if `nil` is interpreted as `"nil"`, but this is only a documentation problem. Same comment applies to generated docs as well.

lib/explorer/data_frame.ex
```diff
@@ -516,7 +516,7 @@ defmodule Explorer.DataFrame do
 
     * `:max_rows` - Maximum number of lines to read. (default: `nil`)
 
-    * `:null_character` - The string that should be interpreted as a nil value. (default: `"NA"`)
+    * `:null_character` - The string that should be interpreted as a nil value. (default: `nil`)
```
philss, line 519: We could also rename the option to `:nil_values` and accept a list of values. 

```suggestion
    * `:nil_values` - A string or a list of strings that should be interpreted as nil values.
       By default, all empty cells are nil, but it's common to have nil values declared as "NA",
       so this option is useful for that. (default: `nil`) 
```

notebooks/exploring_explorer.livemd
philss: Yeah, I agree :+1:

native/explorer/src/dataframe/io.rs
```diff
@@ -77,7 +77,7 @@ pub fn df_from_csv(
         .with_encoding(encoding)
         .with_columns(column_names)
         .with_dtypes(schema)
-        .with_null_values(Some(NullValues::AllColumns(vec![null_char])))
+        .with_null_values(null_char.map(|x| NullValues::AllColumns(vec![x])))
```
philss, line 80: Nice :+1: 

Since we can have a vec here, we can give the option to pass more than one string. WDYT?

philss (commented):
It makes sense to me :+1: 

I added some suggestions.

lib/explorer/data_frame.ex
cnpryer: Read my mind :)

Only clarification needed is I interpret "A string or a list of strings" as list | string | nil. Is that correct? I could just assume list | nil for now.

native/explorer/src/dataframe/io.rs
cnpryer: Agreed f9b34bba148dc710fdf6f5db024e4b54401a33de

lib/explorer/backend/data_frame.ex
```diff
@@ -43,7 +43,7 @@ defmodule Explorer.Backend.DataFrame do
               entry :: fs_entry(),
               dtypes,
               delimiter :: String.t(),
-              null_character :: String.t(),
+              nil_values :: option(list(String.t())),
```
cnpryer, line 46: I also modified this to use `option`.

notebooks/exploring_explorer.livemd
cnpryer: I don't want to mark this as resolved, but it's probably a little bit more obvious in f9b34bba148dc710fdf6f5db024e4b54401a33de. New call-out [here](https://github.com/elixir-explorer/explorer/pull/667#discussion_r1276193406) though.

cnpryer:
I didn't use "nil" in Rust since I figured it's Elixir-specific (I'm very new to Elixir). I can change that as well if you'd like.

lib/explorer/data_frame.ex
```diff
@@ -516,7 +516,7 @@ defmodule Explorer.DataFrame do
 
     * `:max_rows` - Maximum number of lines to read. (default: `nil`)
 
-    * `:null_character` - The string that should be interpreted as a nil value. (default: `"NA"`)
+    * `:nil_values` - A string or a list of strings that should be interpreted as a nil values. (default: `nil`)
```
josevalim, line 519: Given it is a list, do we need to support `nil`? `nil` and empty list should be the same.  So my suggestion would be to require this to be a list. WDYT?

josevalim: Also, are empty entries always `nil`? Can we turn that off?

philss: >  So my suggestion would be to require this to be a list.

I agree! The API should accept only string or a list of strings - and the default is an empty list -, but I think we can pass down as an Option to the NIF, so it's slightly more efficient there. WDYT?

lib/explorer/data_frame.ex
philss: Yeah, it was correct. But I think we can follow José's suggestion and only accept `string | list(string)` for now. The default would be `[]`. I think we can still treat this as an option down in the NIF. WDYT?

lib/explorer/data_frame.ex
josevalim: Nah, I don't think the empty list vs an option makes a large difference in terms of efficiency. I would only allow lists (not even just strings) and nothing more. :) Especially as the name is `nil_values` (plural).

cnpryer: 8349ff32ac257ce8ebd2e7d431d9e16f388ea01f.

It looks like there's maybe some room for more consistency here. I noticed that, for example, `dtypes` can't be `nil` but to the NIF they're `Option` and `columns` can be `list` | `nil`.

Usually I like to scan for high-impact contribs, but if this is something you'd like a PR for I'm happy to send it as well. I can probably be more useful this way while I learn Elixir :)

notebooks/exploring_explorer.livemd
cnpryer: I think the new verbiage in 8349ff32ac257ce8ebd2e7d431d9e16f388ea01f completely resolves this.

lib/explorer/data_frame.ex
josevalim: @cnpryer for columns it makes sense to be list | nil because an empty list would be no columns (even though it makes no sense, at least semantically it has a different meaning). I think though we can remove the Option from dtypes. :) PRs are always welcome!

josevalim:

PR #668, opened 2023-07-27
implement trim_trailing/2 and trim_leading/2

DeemoONeill:
tidy up trim/2 rust implementation


lib/explorer/series.ex
```diff
@@ -4332,9 +4332,27 @@ defmodule Explorer.Series do
   @doc type: :string_wise
   @spec trim_leading(Series.t()) :: Series.t()
   def trim_leading(%Series{dtype: :string} = series),
-    do: apply_series(series, :trim_leading)
+    do: apply_series(series, :trim_leading, [nil])
 
-  def trim_leading(%Series{dtype: dtype}), do: dtype_error("trim_leading/1", dtype, [:string])
+  @doc """
+  Returns a string where all leading examples of the provided binary have been removed.
```
DeemoONeill, line 4338: not really sure on the wording of this

native/explorer/src/series.rs
```diff
@@ -1345,23 +1345,38 @@ pub fn s_trim(s1: ExSeries, pattern: Option<&str>) -> Result<ExSeries, ExplorerE
...
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn s_trim_leading(s1: ExSeries) -> Result<ExSeries, ExplorerError> {
+pub fn s_trim_leading(s1: ExSeries, pattern: Option<&str>) -> Result<ExSeries, ExplorerError> {
     // There are no eager strip functions.
-    Ok(ExSeries::new(s1.utf8()?.replace(r#"^[ \s]+"#, "")?.into()))
+    let pattern = match pattern {
+        None => String::from(r#"^[ \s]+"#),
+        Some(string) => format!(r#"^[{}]+"#, &string),
```
DeemoONeill, line 1361: should this be a group like `^(str)+` rather than a list of chars `^[str]+`

lib/explorer/series.ex
philss: Maybe something like:

```suggestion
  Returns a series where all leading examples of the provided string have been removed.
```

lib/explorer/series.ex
```diff
@@ -4351,9 +4369,27 @@ defmodule Explorer.Series do
   @doc type: :string_wise
   @spec trim_trailing(Series.t()) :: Series.t()
   def trim_trailing(%Series{dtype: :string} = series),
-    do: apply_series(series, :trim_trailing)
+    do: apply_series(series, :trim_trailing, [nil])
+
+  @doc """
+  Returns a string where all trailing examples of the provided binary have been removed.
```
philss, line 4375: Same here:

```suggestion
  Returns a series where all trailing examples of the provided string have been removed.
```

philss (approved):
LGTM! :+1:

native/explorer/src/series.rs
DeemoONeill: @philss what do you think here? currently passing `trim_leading("ab")`

remove all leading `b` and/or all leading `a`

so string `aaaabbbbhello` would become `hello`

doing a non capturing group like `^(?:ab)+` would remove nothing

DeemoONeill: or we could add a guard only allowing strings of length 1 to be passed?

philss: @DeemoONeill oh, good point! I think people would expect it to remove "ab" as a whole. So I would change it to a group.
WDYT, @josevalim ?

josevalim: > should this be a group like ^(str)+ rather than a list of chars ^[str]+

Correct. That's how Elixir's trim_trailing/trim_leading behaves. We can also compare with how rstrip/lstrip behaves by adding tests with two letters. :)

DeemoONeill: okie doke! that's tomorrows job :)

DeemoONeill: well, that was unexpected. 

i wrote a two letter test with rstrip/lstrip and  `"abab"` will return `"hello"` on `"abhelloba"`.

I think matching the expr_function will probably cause the least confusion here

DeemoONeill: `"nx_"` as the pattern will strip `"_nx"` too

josevalim: So it seems it behaves different from Elixir and the given string is a set of characters to remove. So I think we should move away from the `trim` name in Explorer and move it all to `strip`, `rstrip`, and `lstrip`. What do you think?

DeemoONeill: That's probably a safer way to handle it, i've added in the docstring that it removes the set of characters provided. I'll do a rename

PR #669, opened 2023-07-27
implement slice_string/3

DeemoONeill:
add implementation for slicing strings. 

implements a slice_string/2 and a slice_string/3, the former just takes an offset and the latter takes an offset and length

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -116,7 +116,7 @@ defmodule Explorer.Backend.LazySeries do
     trim: 2,
     upcase: 1,
     downcase: 1,
-    slice_string: 3,
+    substring: 3,
     # Float round
     round: 2,
     floor: 1,
@@ -892,8 +892,8 @@ defmodule Explorer.Backend.LazySeries do
   end
 
   @impl true
-  def slice_string(series, offset, length) do
-    data = new(:slice_string, [lazy_series!(series), offset, length])
+  def substring(series, offset, length) do
+    data = new(:substring, [lazy_series!(series), offset, length])
 
     Backend.Series.new(data, :string)
   end
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -116,6 +116,7 @@ defmodule Explorer.Backend.LazySeries do
     trim: 2,
     upcase: 1,
     downcase: 1,
+    slice_string: 3,
     # Float round
     round: 2,
     floor: 1,
@@ -886,6 +887,13 @@ defmodule Explorer.Backend.LazySeries do
   @impl true
   def trim_trailing(series) do
     data = new(:trim_trailing, [lazy_series!(series)])
+
+    Backend.Series.new(data, :string)
+  end
+
+  @impl true
+  def slice_string(series, offset, length) do
+    data = new(:slice_string, [lazy_series!(series), offset, length])
 
     Backend.Series.new(data, :string)
   end
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -4355,6 +4355,61 @@ defmodule Explorer.Series do
 
   def trim_trailing(%Series{dtype: dtype}), do: dtype_error("trim_trailing/1", dtype, [:string])
 
+  @doc """
+  Returns a string sliced from the offset to the end of the string, supporting
+  negative indexing
+
+  ## Examples
+
+      iex> s = Explorer.Series.from_list(["earth", "mars", "neptune"])
+      iex> Explorer.Series.slice_string(s, -3)
+      #Explorer.Series<
+        Polars[3]
+        string ["rth", "ars", "une"]
+      >
+
+      iex> s = Explorer.Series.from_list(["earth", "mars", "neptune"])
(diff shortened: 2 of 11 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -116,6 +116,7 @@ defmodule Explorer.Backend.LazySeries do
     trim: 2,
     upcase: 1,
     downcase: 1,
+    slice_string: 3,
```
DeemoONeill, line 119: i'm unsure what to call this as slice for arrays

DeemoONeill:
this also took me longer than i would like to admit to realise i had to add the new functions to lib.rs. I must have run the tests 50 times getting `nif not found` before i realised 🤦‍♂️

lib/explorer/series.ex
```diff
@@ -4355,6 +4355,61 @@ defmodule Explorer.Series do
 
   def trim_trailing(%Series{dtype: dtype}), do: dtype_error("trim_trailing/1", dtype, [:string])
 
+  @doc """
+  Returns a string sliced from the offset to the end of the string, supporting
+  negative indexing
```
josevalim, line 4360: I assume this works on Unicode, so can we add tests for something like: こんにちは世界?

lib/explorer/backend/lazy_series.ex
josevalim: Yeah, I am not sure what to call it either. And I assume that, if we want to slice lists later, we don't need a list_slice, we would simply overload this function, so we shouldn't have a `_string` suffix. It could be `slice_value` or `slice_contents` or something, but I am not sure.

DeemoONeill: Maybe I can overload the existing slice to work when dtype is string. I'll check tomorrow

josevalim: That won't work because you may still want to slice a string series both ways. :( by list, I meant when we have a series of lists in the future!

cigrainger: I'd go with `substring`, e.g. https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/substr and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring. We don't necessary _have_ to overload it for lists, right?

DeemoONeill: If we call it substring we have the same issue if/when similar functionality is implemented for lists though right?

In that instance would it be called `sublist`?

It's a slight shame its not namespaced. Then `Series.slice` would slice the series. `Series.String.slice` would slice strings etc.

I'm not sure... I'm happy to go any way on this

josevalim: I actually think the namespacing here would not help because using the same function for doing very different things could be confusing. :( I think we have two options here:

1. `substring` as proposed by @cigrainger 

2. `string_part` (to mirror Elixir's `binary_part`) - however, `binary_part` is strict, `binary_part("elixir", 0, 13)` raises if "length" falls outside and it does not support a negative index, so I don't think it is suitable.

Well, that said, then :+1: for `substring` and we might add `sublist` and `subbinary` in the future if necessary. Can you please amend? Also please add tests for things like both start and length falling out of range. :)

lib/explorer/series.ex
```diff
@@ -4355,6 +4355,61 @@ defmodule Explorer.Series do
...
+      #Explorer.Series<
+        Polars[3]
+        string ["arth", "ars", "eptun"]
+      >
+  """
+  @doc type: :string_wise
+  @spec slice_string(Series.t(), integer(), integer()) :: Series.t()
+  def slice_string(%Series{dtype: :string} = series, offset, length)
+      when is_integer(offset)
+      when is_integer(length),
```
josevalim, line 4407:
```suggestion
      when is_integer(length) and length >= 0,
```

lib/explorer/series.ex
```diff
@@ -4355,6 +4355,61 @@ defmodule Explorer.Series do
...
+
+      iex> s = Explorer.Series.from_list(["earth", "mars", "neptune"])
+      iex> Explorer.Series.slice_string(s, 1, 5)
+      #Explorer.Series<
+        Polars[3]
+        string ["arth", "ars", "eptun"]
+      >
+  """
+  @doc type: :string_wise
+  @spec slice_string(Series.t(), integer(), integer()) :: Series.t()
```
josevalim, line 4404:
```suggestion
  @spec slice_string(Series.t(), integer(), non_neg_integer()) :: Series.t()
```

lib/explorer/backend/lazy_series.ex
DeemoONeill: I'll get on it after work. Thanks for the input both :)

lib/explorer/series.ex
DeemoONeill: added

josevalim:
This is awesome @DeemoONeill, thank you for taking the time to improve Explorer! If you need further APIs in the future, don't hesitate to start a discussion! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

DeemoONeill:
Thanks José, appreciate the feedback and help!

PR #672, opened 2023-07-28
Never expect `nil` for CSV IO `dtypes`

cnpryer:
As mentioned in #667, `dtypes` for CSV IO (read) operations will never be `nil`, so this PR adds the changes to reflect that.


josevalim:

PR #674, opened 2023-07-28
Rename trim functions

DeemoONeill:
due to difference in behavior between the polars string functions and elixir's string functions they have been renamed here to align more to the polars names to avoid confusion


DeemoONeill:
I'm not sure how you want to do this, so i've done it renamed them on this branch.

josevalim:

PR #675, opened 2023-07-29
Change :datetime dtype to be {:datetime, time_unit}

sasikumar87:
Based on the comment https://github.com/elixir-explorer/explorer/pull/642#issuecomment-1623283607

This PR handles the first part

1. Change :datetime dtype to be {:datetime, time_unit} ([as listed here](https://docs.rs/polars/latest/polars/chunked_array/object/enum.DataType.html#variants)). Change all conversion functions to handle datetimes with different time units.


lib/explorer/backend/data_frame.ex
```diff
@@ -248,10 +248,18 @@ defmodule Explorer.Backend.DataFrame do
 
         data = container_doc(open, values, close, inspect_opts, &Explorer.Shared.to_string/2)
 
+        type =
+          case Series.dtype(series) do
+            {:datetime, :milli_seconds} -> "datetime[ms]"
+            {:datetime, :micro_seconds} -> "datetime[μs]"
+            {:datetime, :nano_seconds} -> "datetime[ns]"
```
josevalim, line 255: Elixir standard library uses  :second | :millisecond | :microsecond | :nanosecond. So maybe we should align?

```suggestion
            {:datetime, :milli_seconds} -> "datetime[ms]"
            {:datetime, :micro_seconds} -> "datetime[μs]"
            {:datetime, :nano_seconds} -> "datetime[ns]"
```

lib/explorer/backend/series.ex
```diff
@@ -277,7 +277,16 @@ defmodule Explorer.Backend.Series do
...
     close = A.color("]", :list, inspect_opts)
-    dtype = A.color("#{Series.dtype(series)} ", :atom, inspect_opts)
+
+    type =
+      case Series.dtype(series) do
+        {:datetime, :milli_seconds} -> "datetime[ms]"
+        {:datetime, :micro_seconds} -> "datetime[μs]"
+        {:datetime, :nano_seconds} -> "datetime[ns]"
+        others -> others
+      end
```
josevalim, line 287: We should probably encapsulate this function in the `Shared` module. Maybe `dtype_to_string`?

lib/explorer/data_frame.ex
```diff
@@ -2799,7 +2799,7 @@ defmodule Explorer.DataFrame do
           datetime = %NaiveDateTime{} ->
             lazy_s = LazySeries.new(:to_lazy, [datetime])
 
-            Explorer.Backend.Series.new(lazy_s, :datetime)
+            Explorer.Backend.Series.new(lazy_s, {:datetime, :micro_seconds})
```
josevalim, line 2802: Maybe we should keep the default of `microsecond`, which is the precision Elixir uses for its own datetime. WDYT?

lib/explorer/polars_backend/shared.ex
```diff
@@ -108,19 +109,36 @@ defmodule Explorer.PolarsBackend.Shared do
       :category -> Native.s_from_list_categories(name, list)
       :date -> Native.s_from_list_date(name, list)
       :time -> Native.s_from_list_time(name, list)
-      :datetime -> Native.s_from_list_datetime(name, list)
+      {:datetime, _} -> Native.s_from_list_datetime(name, list)
```
josevalim, line 112: Shouldn't we pass the `precision` as argument to `s_from_list_datetime`?

lib/explorer/shared.ex
```diff
@@ -6,7 +6,25 @@ defmodule Explorer.Shared do
...
+      :float,
+      :integer,
+      :string
+    ]
+
+  @doc """
+  Supported datetime dtypes.
+  """
+  def datetime_types,
+    do: [{:datetime, :nano_seconds}, {:datetime, :micro_seconds}, {:datetime, :milli_seconds}]
```
josevalim, line 27: Do we need to make this function public?

lib/explorer/series.ex
```diff
@@ -61,18 +63,27 @@ defmodule Explorer.Series do
...
           | :time
-          | :datetime
+          | {:datetime, :nano_seconds}
+          | {:datetime, :micro_seconds}
+          | {:datetime, :milli_seconds}
           | :float
           | :integer
           | :string
 
+  @type datetime ::
```
josevalim, line 82:
```suggestion
          | datetime_dtype
          | :float
          | :integer
          | :string

  @type datetime_dtype ::
```

josevalim:
Hi @sasikumar87, this looks great! I have dropped some comments. Also, in encoding.rs, we still pattern match on `DataType::Datetime(TimeUnit::Microseconds, None)` but I think it can be `DataType::Datetime(_, None)`.

lib/explorer/series.ex
```diff
@@ -61,18 +63,27 @@ defmodule Explorer.Series do
   alias Explorer.Shared
 
   @valid_dtypes Explorer.Shared.dtypes()
+  @datetime_dtypes Explorer.Shared.datetime_types()
+  @date_or_datetime_dtypes K.++(Explorer.Shared.datetime_types(), [:date])
```
josevalim, line 67:
```suggestion
  @date_or_datetime_dtypes [:date | Explorer.Shared.datetime_types()]
```

josevalim:

PR #677, opened 2023-08-02
Make download of CSV files from S3 work for "DF.from_csv/2"

philss:
This is going to save a file in a cache dir and then try to read a data frame from there.

The implementation only works for the eager backend. We should not support lazy for this (and we should improve the error messages).


lib/explorer/fss/s3.ex
```diff
@@ -0,0 +1,63 @@
...
+    uri =
+      if endpoint = entry.config.endpoint do
+        URI.parse(endpoint)
+      else
+        URI.parse("https://s3.#{entry.config.region}.amazonaws.com")
+      end
+
+    uri
+    |> append_path("/" <> entry.bucket)
+    |> append_path("/" <> entry.key)
```
josevalim, line 28:
```suggestion
    |> append_path("/" <> entry.bucket <> "/" <> entry.key)
```

lib/explorer/fss/utils/http.ex
```diff
@@ -0,0 +1,150 @@
+defmodule Explorer.FSS.Utils.HTTP do
+  # Copied from Livebook's code: https://github.com/livebook-dev/livebook/blob/481d39d7edbb7498c8372d0e35f143935e0aeb92/lib/livebook/utils/http.ex
```
josevalim, line 2: You can move the functions in this module to Explorer.FSS.Utils :)

lib/explorer/polars_backend/shared.ex
```diff
@@ -177,4 +177,26 @@ defmodule Explorer.PolarsBackend.Shared do
...
+  it's not created yet.
+  """
+  def build_path_for_entry(%FSS.S3.Entry{} = entry) do
+    hash =
+      :crypto.hash(:sha256, entry.bucket <> "/" <> entry.key) |> Base.url_encode64(padding: false)
+
+    id = "s3-file-#{hash}"
+
+    cache_opts = if System.get_env("MIX_XDG"), do: %{os: :linux}, else: %{}
+    base_dir = :filename.basedir(:user_cache, "elixir-explorer/datasets", cache_opts)
```
josevalim, line 197: I am wondering if we should use System.tmp_dir instead.  The XDG cache is actually not guaranteed to be cleaned by the OS.

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -35,20 +35,39 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
+        nil_values,
+        skip_rows,
+        header?,
+        encoding,
+        max_rows,
+        columns,
+        infer_schema_length,
+        parse_dates,
+        eol_delimiter
+      )
```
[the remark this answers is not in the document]
philss: doing that now: [6bc71a8](https://github.com/elixir-explorer/explorer/pull/677/commits/6bc71a8982b747902c6b3ec9a8f0db03009d0413)

lib/explorer/polars_backend/shared.ex
philss: Makes sense! Changed in [5215adb](https://github.com/elixir-explorer/explorer/pull/677/commits/5215adb344ae453241e9b527e94707ec70dff907)

PR #682, opened 2023-08-19
Fix FSS.S3 support when bucket is not provided

philss:
This makes Explorer treat the "endpoint" as already containing the bucket name, and force the virtual host style in order to make Polars ignore the bucket name.

Since FSS is always building the correct endpoint in case the user does not provide, we changed the endpoint to be required in the Rust struct.

This work is related to https://github.com/elixir-explorer/fss/pull/2


lib/explorer/polars_backend/shared.ex
```diff
@@ -185,8 +185,10 @@ defmodule Explorer.PolarsBackend.Shared do
   the `System.tmp_dir()`.
   """
   def build_path_for_entry(%FSS.S3.Entry{} = entry) do
+    bucket = entry.config.bucket || "default-explorer-bucket"
+
     hash =
-      :crypto.hash(:sha256, entry.bucket <> "/" <> entry.key) |> Base.url_encode64(padding: false)
+      :crypto.hash(:sha256, bucket <> "/" <> entry.key) |> Base.url_encode64(padding: false)
```
josevalim, line 191: I would encode the endpoint in the hash too.

lib/explorer/fss/s3.ex
```diff
@@ -16,15 +16,19 @@ defimpl Explorer.FSS, for: FSS.S3.Entry do
...
 
   defp url(%FSS.S3.Entry{} = entry) do
+    config = entry.config
+
+    uri = URI.parse(config.endpoint)
+
     uri =
-      if endpoint = entry.config.endpoint do
-        URI.parse(endpoint)
+      if not is_nil(config.bucket) and not String.contains?(config.endpoint, config.bucket) do
```
josevalim, line 24: Invert the if and else branches to simplify the condition?

josevalim:
Hi @philss, unfortunately I am worried of false positives. Imagine you have a bucket named "co" or "zon". Those will both be part of the URL. A better check would be to get the endpoint, split on "." and "/", and then check if it is present, but even that will have a false positive on "com".

I would prefer to be explicit. Could we do this instead:

1. If a bucket is given, assume it is a path-style route, and not in the endpoint
2. If a bucket is not given, assume it is virtual host, and it must be in the endpoint

Then, we don't have to guess. Kino will always return virtual style, with no bucket. When parsing "s3://" style URLs, we can do either "s3://bucket/foobar" (the bucket is not in the endpoint) and "s3:///foobar" (virtual host style, the bucket will be in the endpoint).

This should support all cases with no ambiguity. Would this work?

josevalim (approved):

josevalim (approved):

PR #683, opened 2023-08-20
Add duration datatypes

billylanchantin:
## Description

**_This is a WIP_.** I'm hoping for some feedback before I get too into the weeds!

_EDIT: Rewrote this section for clarity._

Adds the following dtypes:

* `{:duration, :nanosecond}`
* `{:duration, :microsecond}`
* `{:duration, :millisecond}`

Adds support for some basic operations with those dtypes:

* `Explorer.Series.from_list/2`
* `Explorer.Series.{add/2, subtract/2, multiply/2, divide/2}`
  * These now work in combination with the `{:datetime, *}` dtypes

In principle we could also support most 1-arity functions which accept `:integer` since durations are `:integer` under the hood. But I wanted to stop here to see if this is roughly what the team had in mind.

## Examples

Adapted from `test/duration_test.exs`:

### `Series.add/2`

```elixir
iex(1)> eleven = Series.from_list([~N[2023-08-20 11:00:00.0000000]])
iex(2)> one_hour = Series.from_list([3600 * 1_000], dtype: {:duration, :millisecond})
```

```
#Explorer.Series<
  Polars[1]
  duration[ms] [3600000]
>
```

```elixir
iex(3)> sum = Series.add(eleven, one_hour)
```

```
#Explorer.Series<
  Polars[1]
  datetime[μs] [2023-08-20 12:00:00.000000]
>
```

<details>

<summary>Click to show/hide more examples</summary>

### `Series.subtract/2`

```elixir
iex(1)> eleven = Series.from_list([~N[2023-08-20 11:00:00.0000000]])
iex(2)> twelve = Series.from_list([~N[2023-08-20 12:00:00.0000000]])
iex(3)> one_hour = Series.subtract(twelve, eleven)
```

```
#Explorer.Series<
  Polars[1]
  duration[μs] [3600000000]
>
```

### `DataFrame.mutate/2`

```elixir
iex(1)> df = DF.new(eleven: [~N[2023-08-20 12:00:00.0000000]], twelve: [~N[2023-08-20 11:00:00.0000000]])
iex(2)> df_with_diff = DF.mutate(df, diff: twelve - eleven)
```

```
#Explorer.DataFrame<
  Polars[1 x 3]
  eleven datetime[μs] [2023-08-20 11:00:00.000000]
  twelve datetime[μs] [2023-08-20 12:00:00.000000]
  diff duration[μs] [3600000000]
>
```

</details>

## Related Work

- https://github.com/elixir-explorer/explorer/issues/568
- https://github.com/elixir-explorer/explorer/pull/642
  - Comment from @josevalim: https://github.com/elixir-explorer/explorer/pull/642#issuecomment-1623283607

## Overview of approach

_EDIT: Rewrote this section for clarity._

The main idea is to rely on polars native support for durations. For example:

https://www.rustexplorer.com/b/yzh47r

```rust
// Excerpt from example
// ...
let dt1 = Series::new("datetime1", dt1_values);
let dt2 = Series::new("datetime2", dt2_values);
let diff = Series::new("diff", &dt2 - &dt1);
println!("{:?}", diff);
```

```
shape: (2,)
Series: 'diff' [duration[ms]]
[
	2d
	3d
]
```

To enable this, I've created explicit duration dtypes on the Elixir side and mappings to/from the Rust side which mirror the equivalent ones for `datetime`:

  * `s_from_list_datetime`
  * `s_from_list_duration` <-- new
  * `datetime_series_to_list`
  * `duration_series_to_list` <-- new

Then on the Elixir side, I'm doing a bunch of case work to make sure the `Explorer.Series.{add,subtract,multiply,divide}` functions recognize situations where a duration is involved and let that operation through. Before they'd only allow something like (pseudocode):

  * `:integer - :integer`

They now allow things like:

  * `{:datetime, :microsecond} - {:duration, :nanosecond}`

This required that I figure out what situations are allowed ahead of time. I took a stab at what I thought the allowed operations should be:

<details>

<summary>Click to show/hide</summary>

---

Operation | Left | Right | Output
:-- | :-- | :-- | :--
`:add` | `%NaiveDateTime{}` | `{:duration, r}` | `{:datetime, r}`
`:add` | `{:duration, l}` | `%NaiveDateTime{}` | `{:datetime, l}`
`:add` | `{:datetime, l}` | `{:duration, r}` | `{:datetime, higher(l, r)}`
`:add` | `{:duration, l}` | `{:datetime, r}` | `{:datetime, higher(l, r)}`
`:add` | `{:duration, l}` | `{:duration, r}` | `{:duration, higher(l, r)}`
`:subtract` | `%NaiveDateTime{}` | `{:datetime, r}` | `{:duration, r}`
`:subtract` | `{:datetime, l}` | `%NaiveDateTime{}` | `{:duration, l}`
`:subtract` | `%NaiveDateTime{}` | `{:duration, r}` | `{:datetime, r}`
`:subtract` | `{:datetime, l}` | `{:datetime, r}` | `{:duration, higher(l, r)}`
`:subtract` | `{:datetime, l}` | `{:duration, r}` | `{:datetime, higher(l, r)}`
`:subtract` | `{:duration, l}` | `{:duration, r}` | `{:duration, higher(l, r)}`
`:multiply` | `{:duration, l}` | `:integer` | `{:duration, l}`
`:multiply` | `:integer` | `{:duration, r}` | `{:duration, r}`
`:divide` | `{:duration, l}` | `:integer` | `{:duration, l}`

where `higher/2` selects the higher precision assuming `:nanosecond > :microsecond > :millisecond`.

---

</details>

We can edit this as needed.

Notably, durations are currently getting mapped to bare integers on the Elixir side. E.g.

```elixir
eleven = Series.from_list([~N[2023-08-20 11:00:00.0000000]])
twelve = Series.from_list([~N[2023-08-20 12:00:00.0000000]])
diff = Series.subtract(twelve, eleven)

diff |> Series.to_list() |> IO.inspect() #=> [3_600_000_000]
```

Since there's no native duration in Elixir, I think that's the best we can do with this approach unless we create our own `%Explorer.Duration{}`.

## Open questions

- [x] Are we okay with the durations being bare integers on the elixir side?
  - No, we're going with `%Explorer.Duration{}`
- [x] Should we support durations for `Time` and `Date`? I only implemented comparing `NaiveDateTime`s. Doing so will be a bit tricky.
  - No, we'll support `date - date` in a later PR and won't support `time - time` at all.
- [x] How do we feel about picking the highest precision? (See the operation examples)
  - 👍

## To-do

- [x] `Series.to_iovec` support
- [x] Use `%Explorer.Duration{}` instead of bare integers
- [ ] Don't post-cast
  - [x] add / subtract
  - [ ] multiply / divide (blocked)
- [x] More tests for series
- [ ] More tests for dataframes
- [ ] Revisit property tests
- [ ] Revisit 1-arity functions like `Series.skew`


philss (commented):
Hey @billylanchantin, this is looking awesome!

I dropped some comments, but I think this is a good start!

lib/explorer/backend/lazy_series.ex
```diff
@@ -635,6 +661,51 @@ defmodule Explorer.Backend.LazySeries do
...
+      {:add, {:duration, ltu}, {:duration, rtu}} -> {:duration, highest_precision(ltu, rtu)}
+      {:subtract, {:datetime, ltu}, {:datetime, rtu}} -> {:duration, highest_precision(ltu, rtu)}
+      {:subtract, {:datetime, ltu}, {:duration, rtu}} -> {:datetime, highest_precision(ltu, rtu)}
+      {:subtract, {:duration, ltu}, {:duration, rtu}} -> {:duration, highest_precision(ltu, rtu)}
+      {:multiply, :integer, {:duration, tu}} -> {:duration, tu}
+      {:multiply, {:duration, tu}, :integer} -> {:duration, tu}
+      {:divide, {:duration, tu}, :integer} -> {:duration, tu}
+      {:divide, _, _} -> :float
+      _ -> resolve_numeric_dtype([left, right])
+    end
```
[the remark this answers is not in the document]
josevalim: :+1: for more tests. I am just not sure if property based testing helps because it will be mostly hardcoding the rules above. :(

lib/explorer/backend/lazy_series.ex
```diff
@@ -635,6 +661,51 @@ defmodule Explorer.Backend.LazySeries do
...
+    case {op, ldt, rdt} do
+      {:add, {:datetime, ltu}, {:duration, rtu}} -> {:datetime, highest_precision(ltu, rtu)}
+      {:add, {:duration, ltu}, {:datetime, rtu}} -> {:datetime, highest_precision(ltu, rtu)}
+      {:add, {:duration, ltu}, {:duration, rtu}} -> {:duration, highest_precision(ltu, rtu)}
+      {:subtract, {:datetime, ltu}, {:datetime, rtu}} -> {:duration, highest_precision(ltu, rtu)}
+      {:subtract, {:datetime, ltu}, {:duration, rtu}} -> {:datetime, highest_precision(ltu, rtu)}
+      {:subtract, {:duration, ltu}, {:duration, rtu}} -> {:duration, highest_precision(ltu, rtu)}
+      {:multiply, :integer, {:duration, tu}} -> {:duration, tu}
+      {:multiply, {:duration, tu}, :integer} -> {:duration, tu}
+      {:divide, {:duration, tu}, :integer} -> {:duration, tu}
```
josevalim, line 674: Should we raise for `{:divide, :integer, {:duration, _}}`?

josevalim: Actually, I can see in series.ex that we do support `integer / duration` and we cast it into a duration? But I wonder if we should raise instead.

lib/explorer/series.ex
```diff
@@ -2201,14 +2215,13 @@ defmodule Explorer.Series do
   @spec skew(series :: Series.t(), opts :: Keyword.t()) :: float() | non_finite() | nil
   def skew(series, opts \\ [])
 
-  def skew(%Series{dtype: dtype} = series, opts)
-      when is_numeric_or_date_dtype(dtype) do
+  # BILLY-NOTE: Bug? They're checking for numeric-or-temporal but only allowing numeric
```
josevalim, line 2218: If it is simpler, let's only allow numeric here for now.

lib/explorer/series.ex
```diff
@@ -2546,7 +2578,25 @@ defmodule Explorer.Series do
       >
   """
   @doc type: :element_wise
-  @spec add(left :: Series.t() | number(), right :: Series.t() | number()) :: Series.t()
+  @spec add(
+          left :: Series.t() | number() | NaiveDateTime.t(),
+          right :: Series.t() | number() | NaiveDateTime.t()
+        ) :: Series.t()
+  def add(%NaiveDateTime{} = left, %Series{dtype: {:duration, timeunit}} = right),
+    do: apply_series_list(:add, [left, right]) |> cast({:datetime, timeunit})
```
josevalim, line 2586: Just a nitpick: I am not sure if we should cast here. If we cast here, we are assuming that the backend does not return something with the correct dtype. I think that's a backend logic: we say what is the expected dtype and the backend is responsible for dealing with it. If we remove the cast, what will happen? Do we get an integer or duration or datetime back?

If we get a different behaviour than polars, could this mean that we are actually not casting NaiveDateTime properly when sending it to Polars? Or are we choosing a different behaviour than Polars? If so, why?

The same rationale applies to `cast` in subtract, multiply, and divide.

josevalim (commented):
Amazing work @billylanchantin! 💯 

The direction is great, we added only some minor comments so we can go ahead with this.

Also note we need to add Series.to_iovec support. This should be similar to encoding datetime in encoding.rs, so I am hoping it is not a major addition. :)

lib/explorer/backend/lazy_series.ex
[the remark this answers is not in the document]
billylanchantin: > So many possibilities :O

Yeah it's not my most elegant code 😅. Especially because it's duplicated in a few places. _Definitely_ open to any thoughts on simplifying, though I suspect all versions will be similarly gnarly.

> This may be too much, but if we can add some property tests, it can be easier to guarantee some part of the behaviour.

I think I agree with @josevalim: we'd probably just be duplicating the same logic, meaning the properties don't add anything. However, I can envision properties like the following:

* `add`/`subtract` and `multiply`/`divide` are inverse operations (modulo infinity), so things like `x |> add(y) |> subtract(y)` should either `== x` or type error

Proposal: I'll focus on exhaustively testing the known cases up front then come back and see if there are any properties which make sense. Thoughts?

native/explorer/src/encoding.rs
```diff
@@ -467,6 +479,7 @@ pub fn term_from_value<'b>(v: AnyValue, env: Env<'b>) -> Result<Term<'b>, Explor
         AnyValue::Date(v) => encode_date(v, env),
         AnyValue::Time(v) => encode_time(v, env),
         AnyValue::Datetime(v, time_unit, None) => encode_datetime(v, time_unit, env),
+        AnyValue::Duration(v, _time_unit) => Ok(Some(v).encode(env)),
```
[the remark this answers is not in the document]
billylanchantin: I actually started down that road before I went with integers for simplicity. I still think that may be the right choice, but I wanted to hear y'all's thoughts first.

In particular, bare integers can't capture the units and so don't survive (de)serialization. E.g. this doesn't work:

```elixir
import Explorer.Series

dt1 = from_list([~N[2023-01-01 00:00:00]])
dt2 = from_list([~N[2023-01-02 00:00:00]])

dt2 |> subtract(dt1) |> to_list |> from_list |> add(dt1)
```

You get

```
** (ArgumentError) cannot invoke Explorer.Series.add/2 with mismatched dtypes: :integer and {:datetime, :millisecond}
```

If we had a custom `%Explorer.Duration{}` or similar, the above would work just fine. The downside is that `Explorer` would need to internally represent durations. That would increase the maintenance burden and (possibly?) make it harder to implement custom backends.

I still think it may be worth it but that's the tradeoff IMO.

lib/explorer/backend/lazy_series.ex
billylanchantin: I think we should raise because `integer / duration` would have the wrong units. `duration / integer` can still be measured in seconds, but `integer / duration` would be measured in Hertz.

lib/explorer/backend/lazy_series.ex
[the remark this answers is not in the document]
josevalim: Perfect.

lib/explorer/backend/lazy_series.ex
josevalim: :+1: on raising then. But we would raise it on Explorer.Series, so not a concern here. :)

native/explorer/src/encoding.rs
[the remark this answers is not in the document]
josevalim: I think it is fine to keep it as integer, but we need to choose a unique representation, which would most likely be nanoseconds?

lib/explorer/series.ex
josevalim: Right, I would like to avoid the postcast in general. I also think we should consider converting the NaiveDateTime to a series in `series.ex`, so the backend does not have to handle it either.

There was a reason why we didn't do it, but i don't fully recall. I will investigate.

lib/explorer/series.ex
```diff
@@ -2497,14 +2511,32 @@ defmodule Explorer.Series do
...
   # Arithmetic
 
+  defp enforce_highest_precision(
+         %Series{dtype: {left_base, left_timeunit}} = left,
+         %Series{dtype: {right_base, right_timeunit}} = right
+       )
+       when K.and(
+              K.in(left_base, [:datetime, :duration]),
+              K.in(right_base, [:datetime, :duration])
+            ) do
```
josevalim, line 2529:
```suggestion
       ) do
```

native/explorer/src/encoding.rs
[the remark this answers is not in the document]
billylanchantin: Clarification: I was saying a `%Explorer.Duration{}` struct might be the way to go despite the overhead. (Sorry about my wording.)

Representing durations as integers in Elixir forces the programmer to always pass along type hints whenever a duration is initially created. That may be cumbersome if you're doing multiple transformations involving a duration column. This is unlike datetime where, once you've created the initial series, you can do:

```elixir
datetime = from_list([1649883642 * 1_000 * 1_000], dtype: {:datetime, :millisecond})
#=> #Explorer.Series<
#=>   Polars[1]
#=>   datetime[μs] [54252-09-24 11:40:00.000000]
#=> >
```

then

```elixir
datetime |> to_list |> from_list
#=> #Explorer.Series<
#=>   Polars[1]
#=>   datetime[μs] [54252-09-24 11:40:00.000000]
```

which preserves the precision without having to re-pass in the `:dtype` arg.

However, it looks like we're already in this situation WRT binary:

```elixir
binary = from_list([<<228, 146, 51>>, "Elixir"], dtype: :binary)
#=> #Explorer.Series<
#=>   Polars[2]
#=>   binary [<<228, 146, 51>>, "Elixir"]
#=> >
```

Here you have that same problem (actually it's even worse because we raise):

```elixir
binary |> to_list |> from_list
#=> ** (ArgumentError) argument error
```

So I think my objection isn't as strong.

If we do stay with integers, then I think normalizing on nanosecond precision is the way to go. I'm gonna assume that's the plan for now.

lib/explorer/series.ex
billylanchantin: Will do.

lib/explorer/backend/lazy_series.ex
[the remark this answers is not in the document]
billylanchantin: Partially addressed by: [7e7b961](https://github.com/elixir-explorer/explorer/pull/683/commits/7e7b961f834ea41c40fec1b433b70312b8c0339c)

I've got a note to revisit the property tests in my to-dos.

lib/explorer/backend/lazy_series.ex
billylanchantin: Addressed by:
* [3357f90](https://github.com/elixir-explorer/explorer/pull/683/commits/3357f90078de9187bafc18499df89a3682e8d020)
* [743b306](https://github.com/elixir-explorer/explorer/pull/683/commits/743b3063512e1e8813c6ec01f48cb77ccbc29714)

lib/explorer/series.ex
billylanchantin: Did this for add and subtract here: [2f77e76](https://github.com/elixir-explorer/explorer/pull/683/commits/2f77e76d304bdc2bb76a4510fcdd542c35323247)

Multiply and divide seem determined to get returned as integers. Will investigate what's going on tomorrow.

lib/explorer/series.ex
```diff
@@ -2676,6 +2749,12 @@ defmodule Explorer.Series do
   """
   @doc type: :element_wise
   @spec divide(left :: Series.t() | number(), right :: Series.t() | number()) :: Series.t()
+  def divide(%Series{dtype: {:duration, _} = dtype} = left, %Series{dtype: :integer} = right),
+    do: apply_series_list(:divide, [left, right]) |> cast(dtype)
+
+  def divide(_, %Series{dtype: {:duration, _}}),
+    do: raise("cannot divide by duration")
```
josevalim, line 2756:
```suggestion
    do: raise(ArgumentError, "cannot divide by duration")
```

native/explorer/src/encoding.rs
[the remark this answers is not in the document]
josevalim: There is an issue in normalizing to nanoseconds: a very large value will overflow 64 bits, which is fine for Elixir, but it will lead issues when converting back to Explorer. So I think you are right, `Explorer.Duration` is best way to go indeed.

How would we implement it? Do we keep two fields? Value and precision? Elixir's Calendar types represent the precision as a number (from 0 to 6) we could do the same here and represent it as a number from 0 to 9, but given it will be different anyway, the precision as one of :millisecond | :microsecond | :nanosecond is fine.

Thoughts?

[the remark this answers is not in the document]
josevalim: Something to add: in Elixir, the precision is about the _representation_. If we say the datetime is a single digit, then it is a single digit precision. In here, the precision is about low-level storage (i.e. the biggest/smallest number we can represent). Given they are different, let's go with the precision being an atom (the time unit).

native/explorer/src/encoding.rs
```diff
@@ -493,6 +506,7 @@ pub fn list_from_series(s: ExSeries, env: Env) -> Result<Term, ExplorerError> {
         DataType::Date => date_series_to_list(&s, env),
         DataType::Time => time_series_to_list(&s, env),
         DataType::Datetime(time_unit, None) => datetime_series_to_list(&s, *time_unit, env),
+        DataType::Duration(time_unit) => duration_series_to_list(&s, *time_unit, env),
```
josevalim, line 509: Just a note that whatever we decide above we need to mirror here. :)

lib/explorer/backend/lazy_series.ex
```diff
@@ -635,6 +661,53 @@ defmodule Explorer.Backend.LazySeries do
...
+      {:divide, _, {:duration, _}} -> raise("cannot divide by duration")
+      {:divide, _, _} -> :float
+      _ -> resolve_numeric_dtype([left, right])
+    end
+  end
+
+  defp resolve_numeric_temporal_dtype(op, left, %Series{dtype: dtype} = right) do
```
josevalim, line 681: I think we can remove this clause, right? NaiveDateTimes will be normalized to series in series.ex, so no longer a concern here. :)

josevalim (commented):
I did another pass. I have added some minor notes, but it seems the major decisions have been taken, so we are on a solid path!

native/explorer/src/encoding.rs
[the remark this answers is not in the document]
billylanchantin: Yeah I was thinking exactly that:

```elixir
%Explorer.Duration{
  value: integer,
  precision: :millisecond | :microsecond | :nanosecond
}
```

We may find we more fields down the road of course. But right now this struct only needs to hold on to enough information to make the mapping into polars unambiguous.

billylanchantin:
@josevalim Awesome, thanks! I feel like we're in a pretty good spot too :)

One other major decision to be made: do we want to support `date - date`? I checked and polars does support it (and does not support `time - time`):

https://www.rustexplorer.com/b/cpbrw8

josevalim:
Let’s add date-date in another PR so we pack this one sooner. :)

test/explorer/series/duration_test.exs
```diff
@@ -0,0 +1,368 @@
...
+               duration[ms] [1ms]
+             >\
+             """
+
+      assert inspect(us) == """
+             #Explorer.Series<
+               Polars[1]
+               duration[μs] [1ms]
+             >\
+             """
```
billylanchantin, line 125: I took a pass at implementing the string-related protocols while I was adding the duration struct. Not having them broke one of my tests (plus I was tired of counting 0's). Feel free to 👍 / 👎 on the specifics. I mostly tried to copy what polars was doing. But for `inspect` I went with:

```elixir
inspect(%Explorer.Duration{value: 1, precision: :millisecond})
#=> "Duration[1ms]"
```

I don't feel strongly about that choice, I just went with something.

native/explorer/src/encoding.rs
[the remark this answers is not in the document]
billylanchantin: Addressed by: [98a8ff7](https://github.com/elixir-explorer/explorer/pull/683/commits/98a8ff75d1114250d52d319f233c72cbd187d92d)

native/explorer/src/encoding.rs
billylanchantin: Addressed by: [98a8ff7](https://github.com/elixir-explorer/explorer/pull/683/commits/98a8ff75d1114250d52d319f233c72cbd187d92d)

lib/explorer/series.ex
billylanchantin: Ok I think I know what's going on. Polars seems to be treating scalar and element-wise multiplication differently.

https://www.rustexplorer.com/b/4cdukd

```rust

println!("{:?}", diff);                                  // [1d]
println!("{:?}", Series::new("prod1", &diff * 10));      // [10d]
println!("{:?}", Series::new("prod2", (10).mul(&diff))); // [10d]
println!("{:?}", Series::new("prod3", &ten * &diff));    // [864000000]
println!("{:?}", Series::new("prod4", &diff * &ten));    // [864000000]
```

Unfortunately it's the element-wise multiplication that's returning integers. Also:

```rust
println!("{:?}", Series::new("prod4", &diff * &diff));
//=> mul operation not supported for dtypes `duration[ms]` and `duration[ms]`
```

So I think we either have to drop support or post-cast.

I wonder if this is a bug in polars. Seems like an odd choice if it's intentional. Pandas seems to support what we're going for:

```python
import pandas as pd

duration = pd.Series([pd.Timedelta(1, "d"), pd.Timedelta(2, "d")])
integer = pd.Series([10, 10])
print(duration * integer)
#=> 0   10 days
#=> 0   20 days
#=> dtype: timedelta64[ns]
```

lib/explorer/backend/lazy_series.ex
billylanchantin: Yes I think so. Done: [98c2885](https://github.com/elixir-explorer/explorer/pull/683/commits/98c28851b3c72cd3aa68298a17e80bdaee3e16fb)

billylanchantin: Seems like we can make a similar simplification to `resolve_numeric_dtype/1` in that same module.

josevalim: I have another PR for this, which I want to merge after this one to not mess up your code. :)

lib/explorer/series.ex
```diff
@@ -932,6 +942,7 @@ defmodule Explorer.Series do
     * `:date` - Date type that unwraps to `Elixir.Date`
     * `:time` - Time type that unwraps to `Elixir.Time`
     * `:datetime` - DateTime type that unwraps to `Elixir.NaiveDateTime`
+    * `:duration` - Duration type that unwraps to ... BILLY-TODO
```
josevalim, line 945:
```suggestion
    * `:duration` - Duration type that unwraps to `Explorer.Duration`
```

lib/explorer/series.ex
```diff
@@ -2201,14 +2215,13 @@ defmodule Explorer.Series do
   @spec skew(series :: Series.t(), opts :: Keyword.t()) :: float() | non_finite() | nil
   def skew(series, opts \\ [])
 
-  def skew(%Series{dtype: dtype} = series, opts)
-      when is_numeric_or_date_dtype(dtype) do
+  # BILLY-NOTE: Bug? They're checking for numeric-or-temporal but only allowing numeric
+  def skew(%Series{dtype: dtype} = series, opts) when is_numeric_or_temporal_dtype(dtype) do
```
josevalim, line 2219:
```suggestion
  def skew(%Series{dtype: dtype} = series, opts) when is_numeric_dtype(dtype) do
```

josevalim:
This is awesome @billylanchantin, thank you so much! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

This is complete as is and will unblock us on upcoming work. Feel free to send PRs for the time-time and the other missing pieces. :)

josevalim:
Just one additional note. I removed the post-cast  from multiply/divide because we are not allowing integers on add/subtract, so I thought we should wait and tackle all of this at once. :)

josevalim:
Another note: to support operations between duration and integers, I think you can change basic_numeric_operation directly. I have just pushed some changes to make that a bit easier too.

native/explorer/cargo-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "nightly-2023-06-23"
```
billylanchantin, line 2: I forgot to call this out earlier. The most recent rust nightly didn't work for me and I had to use the specific version from the README. I've never really used rust before and IDK if what I did is a good practice. 🤷‍♂️

Just wanted to make sure this wasn't lost in the shuffle.

philss: No problem! I think it's a good idea to keep this file. Polars is doing the same.
I think we can simplify our CI files to read from this, but I don't know how yet.

Thanks for pointing this out!

josevalim:
I have just pushed a refactoring to `basic_numeric_operation` and the boolean mask now called `cast_to_ordered_series` which should make it easier to add operations between integers and durations.

PR #685, opened 2023-08-22
Convert numeric values to series early

josevalim:
To be merged after #683.

We can also do similar work for comparison operators, effectively removing the need for `to_series`/`to_polars_series` in the Explorer.PolarsBackend.Series.


josevalim:

PR #691, opened 2023-08-24
Adds tests showing problem detecting :nan with `in`

kellyfelkins:
With a float series created with `divide()`, :nan (NaN) values cannot be detected with `in()`.

With a float series created with `from_list()` containing :nan, values can be detected with `in()`

I added these tests but, so far, the solution is beyond me. The problem may be in Polars.

I'm providing this as a PR even though the tests do not pass.

### DIFF
--- a/test/explorer/series_test.exs
+++ b/test/explorer/series_test.exs
@@ -1141,6 +1141,71 @@ defmodule Explorer.SeriesTest do
       assert s1 |> Series.in(s2) |> Series.to_list() == [true, false, true, true, true]
     end
 
+    test "with results from divide can find nan and infinity" do
+      s1 = Series.from_list([1, 0, 0])
+      s2 = Series.from_list([4, 1, 0])
+      s3 = Series.divide(s2, s1)
+
+      assert inspect(s3) ==
+               """
+               #Explorer.Series<
+                 Polars[3]
+                 float [4.0, Inf, NaN]
+               >
+               """
+               |> String.trim()
+
+      assert Series.to_list(s3) == [4.0, :infinity, :nan]
+
+      s4 = Series.from_list([:nan, :infinity])
+
+      assert inspect(s4) ==
+               """
+               #Explorer.Series<
+                 Polars[2]
+                 float [NaN, Inf]
+               >
+               """
+               |> String.trim()
+
+      boolean_series = Series.in(s3, s4)
+
+      assert Series.to_list(boolean_series) == [false, true, true]
+    end
+
+    test "with float series can find nan and infinity" do
+      #      s1 = Series.from_list([1, 0, 0])
+      #      s2 = Series.from_list([4, 1, 0])
+      #      s3 = Series.divide(s2, s1)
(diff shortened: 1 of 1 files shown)


philss:
Hey @kellyfelkins :wave: 

Thank you for opening the issue / PR. We are investigating it. The problem seems to be on how the `in` operation is implemented on the version of Polars we are using.

I'm going to check if this is fixed on their main branch. If it is not, I should open an issue there. I should reply here tomorrow :)

philss:
After thinking for a while about this, and after [this comment](https://github.com/pola-rs/polars/issues/2643#issuecomment-1039415001), I think the behavior is correct. It's just not what we expect.

What it's happening is that the "NaN" is not guarantee to be equal to another "NaN", like the [Rust docs explains](https://doc.rust-lang.org/std/primitive.f64.html#associatedconstant.NAN):

> Note that IEEE 754 doesn’t define just a single NaN value; a plethora of bit patterns are considered to be NaN. Furthermore, the standard makes a difference between a “signaling” and a “quiet” NaN, and allows inspecting its “payload” (the unspecified bits in the bit pattern). This constant isn’t guaranteed to equal to any specific NaN bitpattern, and the stability of its representation over Rust versions and target platforms isn’t guaranteed.

The confusion happens because the value used to create a `:nan` using `Series.from_list/2` is from a constant, and that constant is most likely to be equal if the series you are testing is also created using `Series.from_list/2`.

When checking for NaNs in series, it's better to always use [`Series.is_nan/1`](https://hexdocs.pm/explorer/Explorer.Series.html#is_nan/1), since that can guarantee to return true for any NaN value. WDYT?

philss:
I'm going to close this PR, but we can open an issue if you think this deserves more attention (feel free to reply here before that :). Thanks!

josevalim:
It may be worth filling this out as a Polars bug. Their implementation uses hashing (which will depend on the constant values, as the Rust docs describe) and does not properly check for nans. They may decide to not address it in favor of performance though.

kellyfelkins:
Using Series.is_nan/1 is reasonable.

I was trying to use Series.in/2 to mutate results from a divide (percentage change), converting :nan's and :infinity to nil, so that sorted results would make sense. This was solved by not performing the divide when the potential divisor is zero.

However, I wish I could prevent others from going down this path.

Can we add an error message for Series.in/2 if the series on the right contains :nan?

Does the same issue occur with Series.equal/2?

kellyfelkins:
Thank you @philss and @josevalim for addressing these issues and following up.

josevalim:
> Can we add an error message for Series.in/2 if the series on the right contains :nan?

Unfortunately checking the query contains a nan would be expensive and not possible in all cases such as queries. :( We will report upstream to see if we can at least make it consistent.

For Series.equal/2, comparing NaNs is always false, which is the correct behaviour.

philss:
@josevalim @kellyfelkins report created: https://github.com/pola-rs/polars/issues/10808 :)

PR #701, opened 2023-09-03
Support duration-related Polars fixes in exprs

billylanchantin:
## Description

This PR attempts to fix the expr-related inconsistencies introduced by: https://github.com/elixir-explorer/explorer/pull/696.

Reminder: Polars was returning unexpected dtypes for duration arithmetic. We fixed it for the eager series functions but got stuck on doing the same thing for the lazy ones. The issue was that we didn't have the dtype info we needed for an analogous fix.

* Where we left it:
    * https://github.com/elixir-explorer/explorer/pull/696#issuecomment-1697566043
* Discussion on possible fixes:
    * https://github.com/elixir-explorer/explorer/pull/696#issuecomment-1697957315

## Approach

I noticed that, so far as I could tell, every time we made a `LazySeries` struct we also happened to have the `dtype` on hand. So tried just adding a `dtype` field to the `LazySeries` struct and it seemed to work. The only questionable thing I found was a test that I'll call out below.

With a `:dtype` field on the `LazySeries` struct, it was straightforward to add a patch to `expression.ex`.

## Main Question

**Is this actually something that's safe to do?**

I haven't audited enough of the code to know if we'll actually always have the dtype info. But it sure seemed like it.

### DIFF
--- a/lib/explorer/polars_backend/expression.ex
+++ b/lib/explorer/polars_backend/expression.ex
@@ -13,7 +13,6 @@ defmodule Explorer.PolarsBackend.Expression do
   @type t :: %__MODULE__{resource: reference()}
 
   @all_expressions [
-    add: 2,
     all_equal: 2,
     argmax: 1,
     argmin: 1,
@@ -29,7 +28,6 @@ defmodule Explorer.PolarsBackend.Expression do
     minute: 1,
     second: 1,
     distinct: 1,
-    divide: 2,
     equal: 2,
     exp: 1,
     abs: 1,
@@ -50,7 +48,6 @@ defmodule Explorer.PolarsBackend.Expression do
     mean: 1,
     median: 1,
     min: 1,
-    multiply: 2,
     n_distinct: 1,
     nil_count: 1,
     not_equal: 2,
@@ -130,6 +127,9 @@ defmodule Explorer.PolarsBackend.Expression do
   ]
 
   @custom_expressions [
+    add: 2,
+    divide: 2,
+    multiply: 2,
     cast: 2,
     fill_missing_with_strategy: 2,
     from_list: 2,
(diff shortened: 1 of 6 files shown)


test/explorer/backend/lazy_series_test.exs
```diff
@@ -5,7 +5,7 @@ defmodule Explorer.Backend.LazySeriesTest do
   alias Explorer.Backend.LazySeries
 
   test "inspect/2 gives a basic hint of lazy series" do
-    data = LazySeries.new(:column, ["col_a"])
+    data = LazySeries.new(:column, ["col_a"], :unknown)
```
billylanchantin, line 8: In a few tests like this, I added an `:unknown` dtype. This was the only file I found where we were making a `LazySeries` and didn't also know the `dtype` when we did so.

lib/explorer/polars_backend/expression.ex
```diff
@@ -241,6 +241,47 @@ defmodule Explorer.PolarsBackend.Expression do
...
+
+  def to_expr(%LazySeries{op: :divide, args: [left, right]}) do
+    expr = Native.expr_divide(to_expr(left), to_expr(right))
+
+    cond do
+      match?({:duration, _}, dtype(left)) and dtype(right) in [:integer, :float] ->
+        wrap_in_cast(expr, dtype(left))
+
+      match?({:duration, _}, dtype(right)) ->
+        raise(ArgumentError, "cannot divide by duration")
```
josevalim, line 278: Btw, since this raise is never allowed, we should move it to series.ex :)

josevalim: Oh, it was already there, so I just removed it. :)

josevalim:
Awesome work @billylanchantin! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

One quick question: do you think it is worth changing the series callbacks so we pass the return type to the series backends?

For example, instead of calling `apply_list(series, :add, [left, right])`, we would call `apply_list(series, :add, [:float, left, right])` (if the result is a float). This has a couple benefits:

1. We align all backends to return the same return type

2. We no longer have to duplicate logic of computing the return type in add/multiply/etc in both series.ex and the backends

/cc @philss

billylanchantin:
Thanks :)

> do you think it is worth changing the series callbacks so we pass the return type to the series backends?

That's a great thought! I like it. It makes those callbacks more declarative and dedups logic 👍

josevalim:
Is this something you would like to work? Or should we open a separate issue for it?

billylanchantin:
@josevalim I'm happy to take it :)

Issue #702, opened 2023-09-03
Pass dtype to Series callbacks


josevalim:
/cc @billylanchantin

PR #703, opened 2023-09-05
Series callbacks accept `out_dtype` as input

billylanchantin:
## Description

This PR changes several `Explorer.Backend.Series` callbacks to require the expected output dtype as one of the arguments. This change was made for:

  * `Series.add/2`
  * `Series.subtract/2`
  * `Series.multiply/2`
  * `Series.divide/2`

## Related Work

Closes: https://github.com/elixir-explorer/explorer/issues/702

## Notes

Reviewing `Explorer.Backend.Series`, I noticed a few categories WRT the output dtype:

`out_dtype`                  | Example
:--                          | :--
N/A (result is not a series) | `Series.dtype/1`
Known at compile time        | `Series.is_nil/1`
Always matches the input     | `Series.head/1`
More complicated             | `Series.add/2`

I could see requiring `out_dtype` as a generally useful pattern when defining callbacks. But I wasn't sure how often requiring `out_dtype` would actually help given the above categories, so I stuck to the 4 basic arithmetic operations with this PR.

How often do we think this pattern would be helpful?

### DIFF
--- a/lib/explorer/backend/series.ex
+++ b/lib/explorer/backend/series.ex
@@ -11,7 +11,16 @@ defmodule Explorer.Backend.Series do
   @type lazy_s :: Explorer.Series.lazy_t()
   @type df :: Explorer.DataFrame.t()
   @type dtype :: Explorer.Series.dtype()
-  @type valid_types :: number() | boolean() | String.t() | Date.t() | Time.t() | NaiveDateTime.t()
+
+  @type valid_types ::
+          number()
+          | boolean()
+          | String.t()
+          | Date.t()
+          | Time.t()
+          | NaiveDateTime.t()
+          | Explorer.Duration.t()
+
   @type non_finite :: Explorer.Series.non_finite()
   @type option(type) :: type | nil
   # x - x is defined for type x.
--- a/lib/explorer/backend/series.ex
+++ b/lib/explorer/backend/series.ex
@@ -23,8 +23,6 @@ defmodule Explorer.Backend.Series do
 
   @type non_finite :: Explorer.Series.non_finite()
   @type option(type) :: type | nil
-  # x - x is defined for type x.
-  @type diffable :: number() | Date.t() | NaiveDateTime.t() | Explorer.Duration.t()
 
   # Conversion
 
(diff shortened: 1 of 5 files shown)


lib/explorer/backend/series.ex
```diff
@@ -11,9 +11,20 @@ defmodule Explorer.Backend.Series do
...
+          | String.t()
+          | Date.t()
+          | Time.t()
+          | NaiveDateTime.t()
+          | Explorer.Duration.t()
+
   @type non_finite :: Explorer.Series.non_finite()
   @type option(type) :: type | nil
+  # x - x is defined for type x.
+  @type diffable :: number() | Date.t() | NaiveDateTime.t() | Explorer.Duration.t()
```
billylanchantin, line 27: I almost called this `ordered`, but I wasn't sure if that was the correct terminology for this codebase.

philss (approved):
LGTM!

lib/explorer/backend/series.ex
josevalim: We always cast in the "front-end" now, so we can remove the `number()` stuff! I will remove it from other typespecs soon.

Also, super minor nitpick, can you please make the dtype the first argument? Because we want to add it to all operations later, having it first will make it consistent for everyone. :)

Thank you!

billylanchantin: > We always cast in the "front-end" now, so we can remove the number() stuff! I will remove it from other typespecs soon.

Oh right! 😅 I'll drop this and just leave the series input.

> Also, super minor nitpick, can you please make the dtype the first argument?

Sure thing!

> Because we want to add it to all operations later, having it first will make it consistent for everyone. :)

I think that makes sense. Though when you say "all operations", are you including the ones where `out_dtype` is known at compile time like `Series.is_nil/1`?

josevalim: I didn’t think about is_nil :D maybe we can skip for Boolean ones? But it doesn’t hurt to be consistent everywhere. Thoughts?

josevalim:

lib/explorer/backend/series.ex
billylanchantin: These were the 4 situations I thought we needed to consider:

`out_dtype`                  | Example
:--                          | :--
N/A (result is not a series) | `Series.dtype/1`
Known at compile time        | `Series.is_nil/1`
Always matches the input     | `Series.head/1`
More complicated             | `Series.add/2`

Obviously for the N/A operations, there is no `out_dtype` so it doesn't make sense to pass it in.

For the other three, I think we have two choices:

1. Always pass `out_dtype` for consistency
2. Only pass `out_dtype` when it's not "obvious"

I generally lean towards consistency. Then there's no need to question "is this one _really_ obvious?" for individual operations. Plus, these details are completely elided from the end user. It's only backend implementers who need to care, so the extra argument isn't a big cost.

That said, many of the operations are like `Series.head/1` where `out_dtype` isn't known at compile time but is still "obvious". For those cases, we may want to consider having an additional `out_dtype` called `:matches` or something. Then we don't need to do any runtime checking, but the backend will still know what's expected of it.

josevalim: Awesome analysis @billylanchantin, thank you!

What if we start the more complicated ones and move on from there? I would perhaps do it for the known at compile time as well but it can be the next step.

billylanchantin: > What if we start the more complicated ones and move on from there?

Sounds good! These are mostly the other arithmetic operations.

BTW, I think this might be the way to address https://github.com/elixir-explorer/explorer/issues/374. Want me to do that while I'm at it?

josevalim: That would be awesome!

Issue #698, opened 2023-08-30
Missing behaviour on filtering DataFrames on categories.


brendon9x:
This is a followup to https://github.com/elixir-explorer/explorer/issues/690

Following https://github.com/elixir-explorer/explorer/commit/d7b204f10702d6294b6bff18d1bbe24bbab7f305, it is now possible to do this:

```elixir
Explorer.Datasets.iris() 
|> Explorer.DataFrame.mutate(%{species: cast(species, :category)})
|> Explorer.DataFrame.filter(species == "Iris-versicolor")
```

but it is not possible to do this:

```elixir
Explorer.Datasets.iris() 
|> Explorer.DataFrame.mutate(%{species: cast(species, :category)})
|> Explorer.DataFrame.filter(species in ["Iris-versicolor"])
```

The error is:
```
** (RuntimeError) Polars Error: string caches don't match: cannot compare categoricals coming from different sources, consider setting a global StringCache.

Help: if you're using Python, this may look something like:

    with pl.StringCache():
        # Initialize Categoricals.
        df1 = pl.DataFrame({'a': ['1', '2']}, schema={'a': pl.Categorical})
        df2 = pl.DataFrame({'a': ['1', '3']}, schema={'a': pl.Categorical})
        # Your operations go here.
        pl.concat([df1, df2])

Alternatively, if the performance cost is acceptable, you could just set:

    import polars as pl
    pl.enable_string_cache(True)

on startup.
    (explorer 0.8.0-dev) lib/explorer/polars_backend/shared.ex:70: Explorer.PolarsBackend.Shared.apply_dataframe/4
```

philss:
I'm investigating this, but I don' know how to solve it yet. The proposed solution for Python could have something equivalent for our backend.

josevalim:
FWIW, I wrote these tests:

```
    test "compare categories with strings" do
      s = Series.from_list(["a", "b", "c", nil, "a"], dtype: :category)

      assert Series.in(s, Series.from_list(["a", "b"])) |> Series.to_list() ==
               [true, false, false, nil, true]

      assert Series.in(s, ["a"]) |> Series.to_list() == [true, false, false, nil, true]
      assert Series.in(s, ["z"]) |> Series.to_list() == [false, false, false, false, false]
      assert Series.in(s, ["a", "z"]) |> Series.to_list() == [false, false, false, false, false]
    end
```

josevalim:
Closing in favor of #704.

PR #704, opened 2023-09-06
Fix `Series.in/2` to work with series of the `:category` dtype

philss:
This partially solves the problem reported in https://github.com/elixir-explorer/explorer/issues/698

This PR also contains a new feature to the `Series.categorise/2` function, to support string series
on the left-hand side.

### DIFF
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -1149,10 +1149,20 @@ pub fn s_categories(s: ExSeries) -> Result<ExSeries, ExplorerError> {
 
 #[rustler::nif(schedule = "DirtyCpu")]
 pub fn s_categorise(s: ExSeries, cat: ExSeries) -> Result<ExSeries, ExplorerError> {
-    let chunks = s.cast(&DataType::UInt32)?.u32()?.clone();
-
     match cat.dtype() {
         DataType::Categorical(Some(mapping)) => {
+            let chunks = if s.dtype() == &DataType::Utf8 {
+                let ids: ChunkedArray<UInt32Type> = s
+                    .utf8()?
+                    .into_iter()
+                    .map(|opt_str| opt_str.and_then(|slice| mapping.find(slice)))
+                    .collect();
+
+                ids
+            } else {
+                s.cast(&DataType::UInt32)?.u32()?.clone()
+            };
+
             let categorical_chunks = unsafe {
                 CategoricalChunked::from_cats_and_rev_map_unchecked(chunks, mapping.clone())
             };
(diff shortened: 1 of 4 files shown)


native/explorer/src/series.rs
```diff
@@ -548,7 +548,52 @@ pub fn s_in(s: ExSeries, rhs: ExSeries) -> Result<ExSeries, ExplorerError> {
         DataType::Date => s.date()?.is_in(&rhs)?,
         DataType::Time => s.time()?.is_in(&rhs)?,
         DataType::Datetime(_, _) => s.datetime()?.is_in(&rhs)?,
-        dt => panic!("is_in/2 not implemented for {dt:?}"),
+        DataType::Categorical(Some(mapping)) => {
+            let l_logical = s.categorical()?.logical();
+
+            match rhs.dtype() {
+                DataType::Utf8 => {
+                    let mut r_ids: Vec<Option<u32>> = vec![];
```
josevalim, line 556: Are we sure we should implement this? If Polars don't implement it, then maybe we should not? Or we should ask it to be implemented there.

native/explorer/src/series.rs
```diff
@@ -548,7 +548,52 @@ pub fn s_in(s: ExSeries, rhs: ExSeries) -> Result<ExSeries, ExplorerError> {
...
+                    for opt in rhs.categorical()?.iter_str() {
+                        match opt {
+                            Some(slice) => {
+                                if let Some(id) = mapping.find(slice) {
+                                    r_ids.push(Some(id));
+                                }
+                            }
+                            None => r_ids.push(None),
+                        }
+                    }
```
josevalim, line 586: So the idea is that we don't need this branch either, both categories should point to the same mapping in memory by default (that's why Polars raise and that's why we changed categorize to support strings, so we can build a category from an existing dtype=category sharing the same mapping).

philss: Yeah, I know we discussed this previously. The problem was only with the cases where `categorise` a string series would create a series with `nils` for elements that are not present. We would  have a new category series that would return `true` for `nils`. I could remove the nil elements, but I wouldn't know if the original series contained a `nil`.

## Examples

```elixir
cats = Explorer.Series.from_list(["a", "b", "c", nil], dtype: :category)
strings = Explorer.Series.from_list(["b", "a", "d", "c"])

new_cats = Explorer.Series.categorise(strings, cats)
#=> #Explorer.Series<
#=>  Polars[4]
#=>  category ["b", "a", nil, "c"]
#=> >
```

And if we use `Series.in/2`:

```elixir
Explorer.Series.in(cats, new_cats)
#=> #Explorer.Series<
#=>   Polars[4]
#=>   boolean [true, true, true, true]
#=> >
```
Since the `nil` was introduced in the `new_cats` because `d` was not an element, it is considering `nil` as an element.
But I think this is not what the user is expecting.

I guess we could require that the comparison - in `Series.in/2` - would work only with the same `:category` dtype, and we could assume that the person did the categorisation and the filter of nils correctly. WDYT?

josevalim: I think that’s true for all series though, no? If I have a list of integers that are nils, and I check against a series of integers with nil, then they will be all true too, right?

philss: Yes, but that is expected. I think if we use `categorise/2` underneath - in the implementation of `Series.in/2`, and the categorised series contains a `nil` that was a result of a missing element, this is going to be a bug. Makes sense?

josevalim (approved):

Issue #515, opened 2023-02-21
Improve boolean support in query


josevalim:
The goal is to support this:

    DF.mutate(foo: select(a == b, c, 0))

As well as multi-clause:

    DF.mutate(foo: select do
      a == b -> c
      d < e -> f
      true -> 0
    end)

We chose `select do` instead of `cond` because, opposite to `cond` which only evaluates the clauses that matches, queries always evaluates all clauses in order to build the expression.

In order to support this, we will need:

  1. [x] To allow any value as the second and third arguments of `select` (the first argument will always be a list). In this case, it should be enough to call `from_list([...])` on arguments that are not lists
  2. [x] Provide `and/2`, `or/2`, and `not/1`
  3. [x] Implement and document `select/1` macro defined above

The reason why we decided to handle this at the query level is because `Series.and(series, false)` does not really have a purpose and we should keep `and/2` and in `Explorer.Series` remain strict about expecting series. We may want to lift this restriction in the future but, at a first glance, it sounds reasonable.

sasikumar87:
@josevalim yes, third item in the list above.

josevalim:
@sasikumar87 do you want to send a PR for the second item?

The idea is to introduce functions for `and`, `or`, and `not`. `and/2` would look like this:

```elixir
def left and right when Kernel.and(is_boolean(left), is_boolean(right), do: Kernel.and(left, right)
def left and right, do: Explorer.Series.and(boolean!(left), boolean!(right))

defp boolean!(%Series{dtype: :boolean} = series) do
  series
end

defp boolean!(other) do
  raise ArgumentError, "boolean operators require either a boolean (true/false) or a boolean series, got: #{inspect(other)}"
end
```

It is untested but that's the general idea. :)

sasikumar87:
@josevalim will do

sasikumar87:
@josevalim to confirm for `and/2`, we should also support one of them being a boolean and other being series, right?

josevalim:
Ah yes, I forgot a clause in boolean! That accepts a Boolean and converts it a series!

PR #706, opened 2023-09-11
Introduce `cond/1` support in queries

sasikumar87:
Referring to https://github.com/elixir-explorer/explorer/issues/515, this PR introduces multi-clause select as a macro, which in-turn expand to nested `Series.select/3` queries. This makes nested selects easy to write in a flat structure and improves readability.

Below are the list of assumptions made in this PR
- Each clause will be prioritised by top-down by their order of definition
- If none of the clauses match, it will be calculated to `nil`

Creating the PR for initial feedback, will add the docs if the approach is good.

### DIFF
--- a/lib/explorer/query.ex
+++ b/lib/explorer/query.ex
@@ -353,6 +353,49 @@ defmodule Explorer.Query do
     {{:"::", meta, [left, right]}, vars}
   end
 
+  defp traverse({:cond, _meta, [[do: clauses]]}, vars, state) do
+    conditions =
+      clauses
+      |> Enum.map(fn {:->, _, [[on_condition], on_true]} ->
+        {condition, _} = traverse(on_condition, vars, state)
+        {truthy, _} = traverse(on_true, vars, state)
+
+        {condition, truthy}
+      end)
+      |> Enum.reverse()
+
+    block =
+      quote do
+        import Explorer.Query
+
+        unquote(conditions)
+        |> Enum.reduce(nil, fn
+          {true, truthy}, _acc ->
+            Explorer.Shared.lazy_series!(truthy)
+
+          {false, _truthy}, acc ->
+            Explorer.Shared.lazy_series!(acc)
+
+          {nil, _truthy}, acc ->
+            Explorer.Shared.lazy_series!(acc)
+
+          {condition, truthy}, acc ->
+            predicate = Explorer.Shared.lazy_series!(condition)
+            on_true = Explorer.Shared.lazy_series!(truthy)
+
+            on_false =
+              case acc do
+                nil -> Explorer.Backend.LazySeries.from_list([nil], on_true.dtype)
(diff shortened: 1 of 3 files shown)


philss (commented):
For me it looks amazing! :heart_eyes: 

But let's wait to hear from José :)

lib/explorer/query.ex
```diff
@@ -599,5 +599,30 @@ defmodule Explorer.Query do
     |> Enum.map(&%{Explorer.Shared.apply_impl(df, :pull, [&1]) | name: &1})
   end
 
+  defmacro select(do: clauses) do
+    conditions =
+      clauses
+      |> Enum.map(fn {:->, _, [[condition], on_true]} -> [condition, on_true] end)
```
josevalim, line 605: You can do the reverse here, so we don't do it at runtime, and you can use tuples (semantically better):

```suggestion
      |> Enum.map(fn {:->, _, [[condition], on_true]} -> {condition, on_true} end)
      |> Enum.reverse()
```

lib/explorer/query.ex
```diff
@@ -599,5 +599,30 @@ defmodule Explorer.Query do
...
+
+    quote do
+      import Explorer.Query
+
+      unquote(conditions)
+      |> Enum.reverse()
+      |> Enum.reduce(nil, fn [condition, truthy], acc ->
```
josevalim, line 612: Something you could optimize is to check if `condition == true`. If that's the case, you know it will always be true and you can skip the whole select. Something like this:

```
  {true, truthy}, _acc -> Explorer.Shared.lazy_series!(truthy)
  {condition, truthy}, acc -> ...
```

josevalim (commented):
This is great @sasikumar87!

The only question I have is if we want to call it `cond do` instead of `select`. I would prefer to keep it closer to Elixir, so `cond` is my vote.

In that case, instead of defining a `defmacro select(...) do`, you should match on `defp traverse({:cond, _meta, clauses}) do`. Everything else should remain the same. The documentation would go in the module docs.

---

PS: Now that I think about it, we could also support `case ... do` but an initial version would require all patterns to be literals (i.e. integers or strings). But that's for another PR. :)

sasikumar87:
>The only question I have is if we want to call it cond do instead of select. I would prefer to keep it closer to Elixir, so cond is my vote.

@josevalim I read the following in https://github.com/elixir-explorer/explorer/issues/515,
> We chose `select do` instead of `cond` because, opposite to `cond` which only evaluates the clauses that matches, queries always evaluates all clauses in order to build the expression.

Do you think this argument still has merit to use `select`? Personally I would prefer `cond` too.

josevalim:
I see @josevalim from the past disagrees with @josevalim from the future. Let me think about it, haha.

josevalim:
@sasikumar87 let's go with `cond`. Thank you!

josevalim:

Issue #718, opened 2023-10-13
missing replace/3 expression


DeemoONeill:
When trying to do a mutate in a dataframe

(ignoring the use of float for what is clearly money)

doing this:
```elixir
DataFrame.mutate(
      df,
      money_in: lstrip(money_in, "£") |> cast(:float),
      money_out: lstrip(money_out, "£") |> cast(:float),
      balance: lstrip(balance, "£")  |> replace(",", "") |> cast(:float)
    )
```
results in a `Erlang error: :nif_not_loaded`  error

looking at explorer, there is a `replace/3` in the `expressions.ex` file, but no underlying implementation in `native`

I was thinking something like this for the implementation:

```rust
#[rustler::nif]
pub fn expr_replace(expr: ExExpr, pat: ExExpr, value: ExExpr) -> ExExpr {
    let expr = expr.clone_inner();
    let pat = pat.clone_inner();
    let value = value.clone_inner();

    ExExpr::new(expr.str().replace(pat, value, true))
}
```

is this desirable?

PR #719, opened 2023-10-13
Add replace/3 expressions

DeemoONeill:
This is a naive initial implementation this might not work with more complex expressions.

The api for Polars replace is a funny one, it takes expressions not strings, this is so they can do things like [passing columns as arguments](https://stackoverflow.com/questions/73427091/polars-replace-part-of-string-in-column-with-value-of-other-column)

however if i make the argument type an expression it fails due to a missing decodor in rustler::nif

If i make it an ExExpr then i get an argument error.

Any ideas on this?

links to https://github.com/elixir-explorer/explorer/issues/718

### DIFF
--- a/native/explorer/src/expressions.rs
+++ b/native/explorer/src/expressions.rs
@@ -795,6 +795,15 @@ pub fn expr_substring(expr: ExExpr, offset: i64, length: Option<u64>) -> ExExpr 
     let expr = expr.clone_inner();
     ExExpr::new(expr.str().str_slice(offset, length))
 }
+#[rustler::nif]
+pub fn expr_replace(expr: ExExpr, pat: String, value: String) -> ExExpr {
+    let expr = expr.clone_inner();
+    ExExpr::new(expr.str().replace_all(
+        Expr::Literal(LiteralValue::Utf8(pat)),
+        Expr::Literal(LiteralValue::Utf8(value)),
+        true,
+    ))
+}
 
 #[rustler::nif]
 pub fn expr_round(expr: ExExpr, decimals: u32) -> ExExpr {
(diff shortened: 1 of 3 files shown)


josevalim:
Hi @DeemoONeill! You don't have to worry about the other arguments right now because Explorer.Series requires both arguments to be strings. If we want to support series as arguments, then we would change lazy_series.ex to cast all of `replace/3` arguments to expressions, and then you could match on `ExExpr` as it would be guaranteed you get all expressions.

You are welcome to try that in another PR if you want, but this one looks great to me. :heart:

native/explorer/src/expressions.rs
```diff
@@ -795,6 +795,15 @@ pub fn expr_substring(expr: ExExpr, offset: i64, length: Option<u64>) -> ExExpr
     let expr = expr.clone_inner();
     ExExpr::new(expr.str().str_slice(offset, length))
 }
+#[rustler::nif]
```
josevalim, line 798:
```suggestion

#[rustler::nif]
```

josevalim:

DeemoONeill:
@josevalim do the expressions expect the same arguments as their series counterparts? for example if i changed the series Replace types, would that affect the expressions too? 

I've not quite worked out how we know what types the expressions recieve

josevalim:
Series are represented like this:

    %Series{dtype: ..., data: %backend{}}

By default, the backend is a Polars series, but inside `mutate`, the backend is Explorer.Backend.LazySeries.

In other words, when you call `Explorer.Series.replace/3` inside mutate, you always know if the arguments are a string or a series and, if a series, their dtype. We then proxy the call to the backend.

So it goes like this:

1. You call `Explorer.Series.replace(series, "foo",  "bar")`

2. Which calls `series.data.__struct__.replace(series, "foo", "bar")`

3. Since `data.__struct__` is `Explorer.Backend.LazySeries`, you call `Explorer.Backend.LazySeries.replace(series, "foo", "bar")`, which then makes sure all of the arguments are converted to expressions/literals before you call Rust

Now, more than one of the `replace` arguments can be a series, but that introduces some complexities. For example, should their dtypes match? What happens if they come from different backends? That's the most annoying part, you can look at how other functions, such as arithmetic operators and `select` deal with this. :)

PR #724, opened 2023-10-28
Update Polars to v0.34.2

philss:
What changed: https://github.com/pola-rs/polars/releases/tag/rs-0.34.0

This version, as always, contain a bunch of improvements and bug fixes. We had to update the examples from tests using "sample".

There is also a implicit casting when "clip" is using floats as min or max.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -3997,16 +3997,16 @@ defmodule Explorer.DataFrame do
       iex> Explorer.DataFrame.sample(df, 3, seed: 100)
       #Explorer.DataFrame<
         Polars[3 x 10]
-        year integer [2011, 2012, 2011]
-        country string ["SERBIA", "FALKLAND ISLANDS (MALVINAS)", "SWAZILAND"]
-        total integer [13422, 15, 286]
-        solid_fuel integer [9355, 3, 102]
-        liquid_fuel integer [2537, 12, 184]
-        gas_fuel integer [1188, 0, 0]
-        cement integer [342, 0, 0]
-        gas_flaring integer [0, 0, 0]
-        per_capita float [1.49, 5.21, 0.24]
-        bunker_fuels integer [39, 0, 1]
+        year integer [2012, 2013, 2014]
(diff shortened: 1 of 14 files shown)


lib/explorer/polars_backend/series.ex
```diff
@@ -363,7 +363,7 @@ defmodule Explorer.PolarsBackend.Series do
       do: Shared.apply_series(s, :s_clip_integer, [min, max])
 
   def clip(%Series{} = s, min, max),
-    do: Shared.apply_series(s, :s_clip_float, [min * 1.0, max * 1.0])
+    do: Shared.apply_series(cast(s, :float), :s_clip_float, [min * 1.0, max * 1.0])
```
cigrainger, line 366: Nitpick
```suggestion
    do: s |> cast(:float) |> Shared.apply_series(:s_clip_float, [min * 1.0, max * 1.0])
```

lib/explorer/data_frame.ex
```diff
@@ -3997,16 +3997,16 @@ defmodule Explorer.DataFrame do
...
+        year integer [2012, 2013, 2014]
+        country string ["SYRIAN ARAB REPUBLIC", "EGYPT", "AFGHANISTAN"]
+        total integer [12198, 58198, 2675]
+        solid_fuel integer [1, 224, 1194]
+        liquid_fuel integer [7909, 26501, 1393]
+        gas_fuel integer [3265, 24672, 74]
+        cement integer [816, 6800, 14]
+        gas_flaring integer [208, 0, 0]
+        per_capita float [0.61, 0.66, 0.08]
+        bunker_fuels integer [437, 694, 9]
```
cigrainger, line 4009: So their sampling changed even if we use the same seed? 😬

cigrainger (approved):
❤️

josevalim (approved):

lib/explorer/data_frame.ex
philss: Yeah, unfortunately :/

And I tried to "find" the same seed, but I couldn't. The strange part is that if you are using the ":replace" option, it won´t change the results.

Issue #471, opened 2023-01-12
Normalize numeric columns after perform a LazyFrame IO read


philss:
We want to keep numeric columns only using either `Int64` of `Float64`, so we should cast integer and float columns that are not using those two types.

Extracted from https://github.com/elixir-nx/explorer/pull/470

cigrainger:
@philss do we do this? Happy to pick it up if not.

josevalim:
I don't think we do it, so a patch is necessary. But at some point we should also discuss adding more precision to types (which I know you don't want to :D).

cigrainger:
I have seen the light and am happy with adding more precision to types, especially since we get zero copy with Nx. I think it's possible we should have a default precision so we can still have just `:integer` and `:float` (or some other solution to keep things simple for easy data exploration) but I could be convinced otherwise. I do think folks will footgun if we're not careful.

josevalim:
I think keeping :integer and :float, with their current meanings, is completely fine, as they are the closest to their current precision on the Erlang VM.

cigrainger:
So we would just add more types w/ additional precision? E.g. keep `:integer` but add `:i32` for example?

josevalim:
I would follow Nx here. So I would add: {:u, 32}, {:s, 32}, {:f, 32} and so on. Probably start with floats as they are simpler.

philss:
I'm closing this issue because it does not makes sense anymore. I should open a new issue for the remaining work on supporting the new integer dtypes.

PR #739, opened 2023-11-23
Change `:float` to `{:f, 64}` and add `{:f, 32}` dtype

philss:
The idea is to support more dtypes in the way they are represented in Arrow/Polars.

In the near future, we won't need to "normalise" numeric columns anymore, and by consequence, Explorer is going to use less memory if needed.

This is related to https://github.com/elixir-explorer/explorer/issues/471


philss:
I feel that we may want to have a "shortcut" version for declaring the dtypes, like in [Nx](https://hexdocs.pm/nx/Nx.html).
We could accept either `:float` for `{:f, 64}` or `:f32` and `:f64`.

Also, the inspection of the series and dataframes could be different, and follow Nx as well - just `f32` or `f64` -, or slightly larger, like `float64`.

WDYT @josevalim  @cigrainger  @billylanchantin ?

josevalim:
I think we should definitely still support `:float` for backwards compatibility. We can add f32/f64 later, it should be a quick addition. Also, please add tests that `to_iovec` with f32 works. Other than that, this is great, ship it!

josevalim:
> I think we should definitely still support :float for backwards compatibility. 

Suggestion: rename `valid_dtype?` to `normalize_dtype` (which returns nil if no dtype) and rename `validate_dtype` to `normalize_dtype`, which returns the normalize dtype. Now change all call sites of `normalize_dtype` and `normalize_dtype!` to consider its return value (which is the normalized dtype). Then it should be easy to add `:float`, but I think we can do that in the next PR. So let's just add the test for `to_iovec` with f32 to this one and the normalize stuff can come in a separate one!

cigrainger:
Yep I agree -- let's keep the 'simple' names like `:float` and give them defaults as they are today. Otherwise this is fantastic. Re: inspection, I prefer to follow Nx as closely as possible.

philss:
@josevalim @cigrainger I made changes applying your suggestions. Please take a look.

The only thing I didn't change was to add the aliases `:f32` and `:f64` for the float dtypes. I'm planning to do that in another PR tomorrow, with docs as well.

josevalim (approved):

README.md
```diff
@@ -13,7 +13,7 @@ data exploration to Elixir.
 
 Explorer high-level features are:
 
-- Simply typed series: `:binary`, `:boolean`, `:category`, `:date`, `:datetime`, `:duration`, `:float`, `:integer`, `:string`, and `:time`.
+- Simply typed series: `:binary`, `:boolean`, `:category`, `:date`, `:datetime`, `:duration`, `{:f, 32}`, `{:f, 64}`, `:integer`, `:string`, and `:time`.
```
cigrainger, line 16: I wonder if we shouldn't do something like:
```suggestion
- Simply typed series: `:binary`, `:boolean`, `:category`, `:date`, `:datetime`, `:duration`, `:float` (`{:f, 32}` or `{:f, 64}`), `:integer`, `:string`, and `:time`.
```

WDYT?

cigrainger (approved):
Dropped one Q but otherwise this looks fantastic. I'm excited to support more explicit datatypes, esp w/ zero-copy conversion to tensors.

README.md
philss: Agreed! I added more mentions of it in other places as well: https://github.com/elixir-explorer/explorer/pull/739/commits/3605cfcc7ca6877ea7568626d91b933e97148068

PR #725, opened 2023-10-31
Add initial support to the list dtype

philss:
This is the initial work to support "list" datatypes from Arrow/Polars in Explorer.

It may close https://github.com/elixir-explorer/explorer/issues/296.

## Tasks

- [x] `{:list, :integer}`
- [x] `{:list, :boolean}`
- [x] `{:list, :string}`
- [x] `{:list, :float}` (missing special values, like `:nan`) 
- [x] `{:list, :binary}`
- [x] `{:list, :date}`
- [x] `{:list, :time}`
- [x] `{:list, :category}`
- [x] `{:list, {:datetime, _precision}}`
- [x] `{:list, {:duration, _precision}}`
- [x] support creation of homogeneous recursive series - that is: multiple level series, with the same list dtype
- [x] support creation of heterogeneous recursive series - that is: multiple level series, with **different** list dtype
- [x] support inspection of list series
- [x] support casting of list series
- [x] support reads/writes from and to Parquet / IPC / NDJSON files (CSV is not supported). 

---
PS: this is the second attempt to add the list dtype. The first one was: https://github.com/elixir-explorer/explorer/pull/401

PR #741, opened 2023-11-26
Add support for grouping into series

costaraphael:
Hey folks!

My first intention with this PR was to implement [implode](https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.Expr.implode.html) to allow gathering all the values of a given group into a Series. Halfway through it, I noticed I was getting some double wrapping (i.e. `[[[1, 2]], [[3, 4]]]` instead of `[[1, 2], [3, 4]]`).

Then I looked a bit further and noticed there isn't really a need for `implode` to have this feature, we just need to allow plain lazy series in summarize and it _just works_, which is the approach this PR follows. I also updated the docs to reflect that.

Please let me know if there's something I'm overlooking with this PR, or if this is not a wanted feature. I'm more than happy to start this over with a different approach or just scrap it altogether 🙂

### DIFF
--- a/test/explorer/data_frame_test.exs
+++ b/test/explorer/data_frame_test.exs
@@ -3564,6 +3564,21 @@ defmodule Explorer.DataFrameTest do
              }
     end
 
+    test "allows returning the group as a list" do
+      df =
+        DF.new(
+          letters: ~w(a b c d e f g h i j),
+          is_vowel: [true, false, false, false, true, false, false, false, false, false]
+        )
+        |> DF.group_by(:is_vowel)
+        |> DF.summarise(letters: letters)
+
+      assert DF.to_columns(df, atom_keys: true) == %{
+               is_vowel: [true, false],
+               letters: [["a", "e"], ["b", "c", "d", "f", "g", "h", "i", "j"]]
+             }
+    end
+
     test "mode/1" do
       df =
         Datasets.iris()
--- a/test/explorer/data_frame_test.exs
+++ b/test/explorer/data_frame_test.exs
@@ -3568,14 +3568,14 @@ defmodule Explorer.DataFrameTest do
       df =
         DF.new(
           letters: ~w(a b c d e f g h i j),
-          is_vowel: [true, false, false, false, true, false, false, false, false, false]
+          is_vowel: [true, false, false, false, true, false, false, false, true, false]
         )
         |> DF.group_by(:is_vowel)
         |> DF.summarise(letters: letters)
 
       assert DF.to_columns(df, atom_keys: true) == %{
                is_vowel: [true, false],
-               letters: [["a", "e"], ["b", "c", "d", "f", "g", "h", "i", "j"]]
+               letters: [["a", "e", "i"], ["b", "c", "d", "f", "g", "h", "j"]]
              }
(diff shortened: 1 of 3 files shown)


billylanchantin:
@costaraphael Thanks for the PR!

I know that after https://github.com/elixir-explorer/explorer/pull/725 we have our eye on functions like `implode`. Though this does seem like an elegant way to get a similar (the same?) thing!

When you say you were getting double lists, what caused it?

test/explorer/data_frame_test.exs
```diff
@@ -3564,6 +3564,21 @@ defmodule Explorer.DataFrameTest do
...
+          letters: ~w(a b c d e f g h i j),
+          is_vowel: [true, false, false, false, true, false, false, false, false, false]
+        )
+        |> DF.group_by(:is_vowel)
+        |> DF.summarise(letters: letters)
+
+      assert DF.to_columns(df, atom_keys: true) == %{
+               is_vowel: [true, false],
+               letters: [["a", "e"], ["b", "c", "d", "f", "g", "h", "i", "j"]]
+             }
```
billylanchantin, line 3579:
```suggestion
          is_vowel: [true, false, false, false, true, false, false, false, true, false]
        )
        |> DF.group_by(:is_vowel)
        |> DF.summarise(letters: letters)

      assert DF.to_columns(df, atom_keys: true) == %{
               is_vowel: [true, false],
               letters: [["a", "e", "i"], ["b", "c", "d", "f", "g", "h", "j"]]
             }
```

Don't forget about `i` :)

costaraphael:
@billylanchantin I first tried to get something like this working:

```elixir
DF.new(
  letters: ~w(a b c d e f g h i j),
  is_vowel: [true, false, false, false, true, false, false, false, false, false]
)
|> DF.group_by(:is_vowel)
|> DF.summarise(letters: implode(letters))
```

To get this working, I went through the normal steps (adding the function to `Explorer.Series`, and `Explorer.PolarsBackend.Series`, etc), all very straight forward, since `implode` is already implemented as an eager function in `Series` and in `Expr`. To keep just the relevant bits, here's the Rust implementation I added for lazy series:

```rust
#[rustler::nif]
pub fn expr_last(expr: ExExpr) -> ExExpr {

    ExExpr::new(expr.implode())
}
```

When I had this, the `summarise` call above would yield the following:

```
#Explorer.DataFrame<
  Polars[2 x 2]
  is_vowel boolean [true, false]
  letters list[string] [[["a", "e"]], [["b", "c", "d"]]]
>
```

I spent some time trying to understand what was happening, and at one point I tried removing the `implode` call from the Rust code as a sanity check:

```rust
#[rustler::nif]
pub fn expr_last(expr: ExExpr) -> ExExpr {

    ExExpr::new(expr)
}
```

Lo-and-behold, it just worked:

```
#Explorer.DataFrame<
  Polars[2 x 2]
  is_vowel boolean [true, false]
  letters list[string] [["a", "e"], ["b", "c", "d"]]
>
```

Which is when I realized that maybe we don't need the `implode` call at all 😅

josevalim (approved):

josevalim:

Issue #705, opened 2023-09-09
Use sink cloud versions of ipc


josevalim:
We need a new Explorer version:

* https://github.com/pola-rs/polars/pull/10060
* https://github.com/pola-rs/polars/pull/11008

Qqwy:
@josevalim FYI I briefly spoke with Ritchie on the Polars Discord today and he merged the `sink_ipc_cloud` PR just now, so work on the Explorer side can now continue!

philss:
First of all: awesome work there on Polars, @Qqwy !! :tada: 

Just to let you know that I'm working on this to bring the functionality to Explorer :)

philss:
Hey @Qqwy, I had some issues trying to make the Sink IPC work, so I opened an issue upstream: https://github.com/pola-rs/polars/issues/13614

I tried to find a fix for it, but I couldn't understand where the "block_on" is happening (twice). Would you mind to take a look?

Qqwy:
Hi there! Thanks for tagging me, I'll dig into this problem. `block_on` is probably called from within the `#[tokio::main]` decorators. The crash happens because for some reason an attempt is made to start an async Tokio runtime from within a pre-existing Tokio runtime. _Why_ that happens is of course the thing to figure out as it shouldn't.

Just to double-check: Is this only a problem with sink_ipc_cloud and not with sink_parquet_cloud?

josevalim:
As far as _I_ know, only ipc. @philss will be able to confirm though!

philss:
@Qqwy yeah, it's only with the `sink_ipc_cloud`. The `sink_parquet_cloud` is working good, and it's already in main.

PS: sorry, I didn't see the ping before :grimacing:

Qqwy:
@philss I have created a PR with a fix on the Polars repo: https://github.com/pola-rs/polars/pull/14262

PR #742, opened 2023-11-27
Update Polars to v0.35

philss:
Also update object_store and other Rust dependencies.


philss:
The integration tests are failing :grimacing: 
I'm going to check in the morning!

josevalim (approved):

philss:
I had to use the version from GitHub that contains a bug fix for the problem we had with the integration tests.
It's only a few commits ahead of the published version, so I think this is no big deal.

Bonus is that we gain the changes needed for https://github.com/elixir-explorer/explorer/issues/705

PR #746, opened 2023-11-27
Add `Series.length/1` and `Series.member?/2`

costaraphael:
It's me again! 💜 

This PR is adding `Series.length/1` (to calculate the length of list elements) and `Series.member?/2` (to check membership in list elements). There are a couple of things to consider:

- Function names. I'm not sure that's the pattern you want to follow (it seems like it is with `join/2`), but I also thought of something like `list_lenghts/1` and `list_member?/2`. Let me know if you have any thoughts.
- `join/2` was not being implemented for `LazySeries`, so I added a test for it in `DataFrame.mutate/2` and implemented it as well.
- I moved the conversion of struct to expression literals closer to the structs (by directly implementing the `Literal` trait. This felt cleaner/easier to re-use, but I'm more than happy to undo it if you prefer the original way.
- For the eager implementation of `member?/2`, I followed the [Python](https://github.com/pola-rs/polars/blob/38d016b61971c1ce3789fced9a7aa519983d5de9/py-polars/polars/series/list.py#L26-L27) [approach](https://github.com/pola-rs/polars/blob/38d016b61971c1ce3789fced9a7aa519983d5de9/py-polars/polars/series/utils.py#L28) (wrap the series into a frame and re-use the lazy implementation). Let me know if you have any thoughts on this.
- To make `member?/2` work with all our allowed Elixir types, I introduced an `ExValidValue` enum in the Rust code, which essentially dynamically decodes the Erlang terms into a single type, allowing us to have a single function definition in the Rust code. I think this could be re-used for functions like `fill_missing/2` that currently use a double-dispatch approach.
- `member?/2` is not working properly for `:datetime`. It doesn't error out, but it returns false when it shouldn't (I've left a skipped test with a comment to demonstrate this). My gut feeling is that it is a precision issue, but I ran out of time to continue investigating tonight. Please let me know if you have any ideas 😅 

At the same time that it feels like this PR is doing too much, the changes do seem connected, so I've decided to keep it as is instead of splitting it preemptively. Please let me know if you want me to split it up.

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -137,7 +137,7 @@ defmodule Explorer.Backend.LazySeries do
     # List functions
     join: 2,
     length: 1,
-    member: 2
+    member: 3
   ]
 
   @comparison_operations [:equal, :not_equal, :greater, :greater_equal, :less, :less_equal]
@@ -1000,8 +1000,8 @@ defmodule Explorer.Backend.LazySeries do
   end
 
   @impl true
-  def member?(series, value) do
-    data = new(:member, [lazy_series!(series), value], :boolean)
+  def member?(%Series{dtype: {:list, inner_dtype}} = series, value) do
+    data = new(:member, [lazy_series!(series), value, inner_dtype], :boolean)
 
     Backend.Series.new(data, :boolean)
   end
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -135,7 +135,9 @@ defmodule Explorer.Backend.LazySeries do
     minute: 1,
     second: 1,
     # List functions
-    join: 2
+    join: 2,
+    length: 1,
+    member: 2
   ]
 
   @comparison_operations [:equal, :not_equal, :greater, :greater_equal, :less, :less_equal]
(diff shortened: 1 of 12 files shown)


josevalim:
It all looks good to me, awesome job. We just need to tackle the issue with `member?` and datetimes.

Also, we should probably make it so `left in right` supports `literal in series[list]`, but it can be made after this is merged. :)

costaraphael:
@josevalim I figured what was the problem with the datetimes not working! It was indeed a precision issue.

By default, datetimes are mapped with precision `:microsecond`, while the literal conversion provided out of the box for `chrono::naive::NaiveDateTime` may use nanoseconds or microseconds, depending on the year of the datetime. So in that specific test case, it was comparing microseconds with nanoseconds.

I figured out a way to solve this by introducing a `lit_with_matching_precision` function to the `ExValidValue` struct, which checks the target data type and casts precisions accordingly if necessary.

This worked wonders, I can even mix-and-match duration precisions, and as long as the absolute values are the same it just works (I've even updated the duration test to showcase that).

The only caveat is that for this to work nicely for both eager and lazy cases, I started passing the data type information from Elixir to Rust (not really necessary in the eager case, but I thought it would be worth it to keep things consistent).

cigrainger:
This is awesome! Yeah... re: naming I do think there's a lot of potential to get confused with `Series.length/1` and `Series.member?/2`. I think `list_{fun}` makes sense here but I'm open to being convinced otherwise. I would definitely get confused between `Series.length/1` and `Series.size/1`.

cigrainger:
Oh I like `lengths`. Sold. I'm fine with `member?` as well.

costaraphael:
Aye, pushing a commit with the rename now! 😄

josevalim:

Issue #743, opened 2023-11-27
Expose ddof option for `Series.covariance/1`


philss:
This is similar to #740 and it's possible after #742 get merged.

billylanchantin:
Oh nice! I think this may also unlock #729. I tried it a few weeks ago but got stuck on the IPC tests. They treated integer strings a bit loosely, which it looks like this change addresses.

philss:
This one was completed. Thanks, @cigrainger!

PR #747, opened 2023-11-29
Expose ddof on covariance

cigrainger:
Completes #743


josevalim (approved):

philss:
I think one of the failures it's because we are transforming every argument in an expression. Maybe we need to do just like we did with `correlation` here: https://github.com/elixir-explorer/explorer/blob/78d95911263a013b01004a0a5bbeadf4b67bb862/lib/explorer/polars_backend/expression.ex#L206, and have a custom implementation.

The other test is just the arity, for what I could understand.

cigrainger:
> I think one of the failures it's because we are transforming every argument in an expression. Maybe we need to do just like we did with `correlation` here:
> 
> https://github.com/elixir-explorer/explorer/blob/78d95911263a013b01004a0a5bbeadf4b67bb862/lib/explorer/polars_backend/expression.ex#L206
> 
> , and have a custom implementation.
> The other test is just the arity, for what I could understand.

Yes! Just realised that _then_ saw your comment. Would have saved me a little bit 😂. Anyway should pass now. Thanks!

PR #754, opened 2023-12-02
Add `Series.all?/1` and `Series.any?/1`

costaraphael:
This PR introduces two new aggregations: `Series.all?/1` and `Series.any?/1`.

The main thing to keep in mind with these functions is how to deal with `nil`. I could see three approaches:

- Ignore `nil` values (`all?([nil, nil, nil]) == all?([])`)
- Use [Kleene Logic](https://en.wikipedia.org/wiki/Three-valued_logic), treating `nil` as _unknown_, and return `nil` whenever it makes the result also unknown (`any?([nil, true]) == true` and `all?([nil, false]) == false`, but `any?([nil, false]) == nil` and `all?([nil, true]) == nil`)
  - this is how the existing `Series.and/2` and `Series.or/2` functions behave
- Treat `nil` as `false` just like in the `Enum` functions (`all?([nil, true]) == false`)

I went with the first option here as it seems to be the default behavior in Polars, but I can see the appeal for either of them. The one thing to consider about the third option above is that it doesn't have native support in Polars, so it would need to be implemented on the Explorer side (should be straightforward with a call to `fill_missing(false)` before aggregating).

Another route I can see is supporting all behaviors, and letting the caller decide through an option:

```elixir
all?(s, nils: :as_false | :ignore | :kleene)
```

Let me know what your thoughts are!

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -2555,7 +2555,7 @@ defmodule Explorer.Series do
       iex> Series.all?(s)
       true
 
-  Nil values are ignored:
+  Opposite to Elixir but similar to databases, `nil` values are ignored:
 
       iex> s = Series.from_list([nil, true, true])
       iex> Series.all?(s)
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -2597,7 +2597,7 @@ defmodule Explorer.Series do
       iex> Series.any?(s)
       false
 
-  Nil values are ignored:
+  Opposite to Elixir but similar to databases, `nil` values are ignored:
 
       iex> s = Series.from_list([nil, true, true])
       iex> Series.any?(s)
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -2528,6 +2528,90 @@ defmodule Explorer.Series do
     basic_numeric_operation(:covariance, left, right, [ddof])
   end
 
+  @doc """
+  Returns if all the values in a boolean series are true.
+
+  ## Supported dtypes
+
+    * `:boolean`
+
+  ## Examples
+
+      iex> s = Series.from_list([true, true, true])
+      iex> Series.all?(s)
+      true
+
+      iex> s = Series.from_list([true, false, true])
+      iex> Series.all?(s)
+      false
+
+      iex> s = Series.from_list([1, 2, 3])
+      iex> Series.all?(s)
+      ** (ArgumentError) Explorer.Series.all?/1 not implemented for dtype :integer. Valid dtypes are [:boolean]
+
+  An empty series will always return true:
+
+      iex> s = Series.from_list([], dtype: :boolean)
+      iex> Series.all?(s)
+      true
+
+  Nil values are ignored:
+
+      iex> s = Series.from_list([nil, true, true])
+      iex> Series.all?(s)
+      true
+
+      iex> s = Series.from_list([nil, nil, nil], dtype: :boolean)
+      iex> Series.all?(s)
+      true
+  """
+  @doc type: :aggregation
+  @spec all?(series :: Series.t()) :: boolean()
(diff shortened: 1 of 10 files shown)


billylanchantin:
@costaraphael Thanks this is looking good!

RE the approaches: I think consistency with `or` and `and` is the right call. Having them behave differently would be very surprising, especially if you ever tried to combine their outputs.

Also, I think we want to avoid as much as possible doing work on the Elixir side. It's significantly more costly.

costaraphael:
> Also, I think we want to avoid as much as possible doing work on the Elixir side. It's significantly more costly.

100%! When I said "on the Explorer side", I meant in the Rust code. Something like:

```rust
pub fn s_all(s: ExSeries) -> Result<bool, ExplorerError> {
    let s = s.clone_inner();

    Ok(s.bool()?.fill_null_with_values(false)?.all())
}
```

costaraphael:
> I think consistency with or and and is the right call. Having them behave differently would be very surprising, especially if you ever tried to combine their outputs.

Yeah, I can definitely see that. This is also how PostgreSQL deals with `NULL`s.

Do you think it makes sense to support all the modes through an option, and default to `:kleene`?

billylanchantin:
> Do you think it makes sense to support all the modes through an option, and default to :kleene?

Ah I'm just looking at the Polars implementations now:

* https://docs.rs/polars/latest/polars/prelude/enum.Expr.html#method.any
* https://docs.rs/polars/latest/polars/prelude/enum.Expr.html#method.all

They both provide an `ignore_nulls` option where `true` is your `:ignore` option and `false` is your `:kleene` option. I'm now thinking we just mirror their approach. If you want to do it with `nils: :ignore | :kleene` I'm not against it. (IMHO named options > booleans.) But you'll definitely need at least a sentence to explain what `:kleene` is. For reference Polars says this:

> If `ignore_nulls` is `False`, [Kleene logic](https://en.wikipedia.org/wiki/Three-valued_logic) is used to deal with nulls: if the column contains any null values and no `true` values, the output is null.

That also implies not supporting your `as_false` option. The usual approach is to handle `nil` with an explicit imputation. So I think that if someone wants to treat `nil` as `false`, they should do the `fill_missing` call themselves.

lib/explorer/series.ex
```diff
@@ -2528,6 +2528,90 @@ defmodule Explorer.Series do
...
+      iex> Series.all?(s)
+      ** (ArgumentError) Explorer.Series.all?/1 not implemented for dtype :integer. Valid dtypes are [:boolean]
+
+  An empty series will always return true:
+
+      iex> s = Series.from_list([], dtype: :boolean)
+      iex> Series.all?(s)
+      true
+
+  Nil values are ignored:
```
josevalim, line 2558:
```suggestion
  Opposite to Elixir but similar to databases, `nil` values are ignored:
```

lib/explorer/series.ex
```diff
@@ -2528,6 +2528,90 @@ defmodule Explorer.Series do
...
+      iex> Series.any?(s)
+      ** (ArgumentError) Explorer.Series.any?/1 not implemented for dtype :integer. Valid dtypes are [:boolean]
+
+  An empty series will always return `false`:
+
+      iex> s = Series.from_list([], dtype: :boolean)
+      iex> Series.any?(s)
+      false
+
+  Nil values are ignored:
```
josevalim, line 2600:
```suggestion
  Opposite to Elixir but similar to databases, `nil` values are ignored:
```

lib/explorer/series.ex
costaraphael: I thought PostgreSQL always followed Kleene logic for boolean operations, but apparently, that's just true for `AND`, `OR`, etc. The `bool_and` and `bool_or` aggregations just ignore `NULL`s.

TIL!

josevalim:

PR #756, opened 2023-12-03
Add initial support for struct dtype

costaraphael:
This PR is heavily inspired by the work done [here](https://github.com/elixir-explorer/explorer/pull/725).

What works at this point:

- defining basic K/V structures
- defining nested K/V structures
- casting
- inspection
- importing/exporting from parquet/IPC/NDJSON (CSV is not supported)

The goal is to add initial support to the type now and add extra functionality (like `Series.field/2` and `DataFrame.unnest/2`) in the future.

I'm well aware that I'm opening a PR with some big changes without discussing it first, so if this is not a feature the Explorer team wants to add, or if you want to take a completely different approach than what I did, please feel free to scrape this. 🙂

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -16,7 +16,8 @@ defmodule Explorer.Series do
     * `:time` - Time type that unwraps to `Elixir.Time`
     * `{:list, dtype}` - A recursive dtype that can store lists. Examples: `{:list, :integer}` or
       a nested list dtype like `{:list, {:list, :integer}}`.
-    * `{:struct, %{key => dtype}}` - A recursive dtype that can store structs. Examples:
+    * `{:struct, %{key => dtype}}` - A recursive dtype that can store Arrow/Polars structs (not to be
+      confused with Elixir's struct). This type unwraps to Elixir maps with string keys. Examples:
       `{:struct, %{"a" => :integer}}` or a nested struct dtype like `{:struct, %{"a" => {:struct, %{"b" => :integer}}}}`.
 
   The following data type aliases are also supported:
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -16,6 +16,8 @@ defmodule Explorer.Series do
     * `:time` - Time type that unwraps to `Elixir.Time`
     * `{:list, dtype}` - A recursive dtype that can store lists. Examples: `{:list, :integer}` or
       a nested list dtype like `{:list, {:list, :integer}}`.
+    * `{:struct, %{key => dtype}}` - A recursive dtype that can store structs. Examples:
+      `{:struct, %{"a" => :integer}}` or a nested struct dtype like `{:struct, %{"a" => {:struct, %{"b" => :integer}}}}`.
 
   The following data type aliases are also supported:
 
@@ -77,7 +79,7 @@ defmodule Explorer.Series do
   @numeric_dtypes [:integer | @float_dtypes]
   @numeric_or_temporal_dtypes @numeric_dtypes ++ @temporal_dtypes
 
-  @io_dtypes Shared.dtypes() -- [:binary, :string, {:list, :any}]
+  @io_dtypes Shared.dtypes() -- [:binary, :string, {:list, :any}, {:struct, :any}]
 
   @type dtype ::
           :binary
@@ -92,11 +94,13 @@ defmodule Explorer.Series do
           | :integer
           | :string
           | list_dtype
+          | struct_dtype
 
   @type time_unit :: :nanosecond | :microsecond | :millisecond
   @type datetime_dtype :: {:datetime, time_unit}
   @type duration_dtype :: {:duration, time_unit}
   @type list_dtype :: {:list, dtype()}
+  @type struct_dtype :: {:struct, %{String.t() => dtype()}}
 
   @type t :: %Series{data: Explorer.Backend.Series.t(), dtype: dtype()}
   @type lazy_t :: %Series{data: Explorer.Backend.LazySeries.t(), dtype: dtype()}
@@ -2202,7 +2206,7 @@ defmodule Explorer.Series do
 
   ## Supported dtypes
 
-  All except `:list`.
+  All except `:list` and `:struct`.
 
   ## Examples
 
@@ -2230,7 +2234,7 @@ defmodule Explorer.Series do
   @doc type: :aggregation
   @spec mode(series :: Series.t()) :: Series.t() | nil
   def mode(%Series{dtype: {:list, _} = dtype}),
-    do: dtype_error("mode/1", dtype, Shared.dtypes() -- [{:list, :any}])
+    do: dtype_error("mode/1", dtype, Shared.dtypes() -- [{:list, :any}, {:struct, :any}])
 
   def mode(%Series{} = series),
     do: Shared.apply_impl(series, :mode)
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -2,7 +2,7 @@ defmodule Explorer.Shared do
   # A collection of **private** helpers shared in Explorer.
   @moduledoc false
 
-  @non_list_types [
+  @scalar_types [
     :binary,
     :boolean,
     :category,
@@ -27,7 +27,7 @@ defmodule Explorer.Shared do
   within lists inside.
   """
   def dtypes do
-    @non_list_types ++ [{:list, :any}]
+    @scalar_types ++ [{:list, :any}, {:struct, :any}]
   end
 
   @doc """
@@ -37,7 +37,20 @@ defmodule Explorer.Shared do
     if maybe_dtype = normalise_dtype(inner), do: {:list, maybe_dtype}
   end
 
-  def normalise_dtype(dtype) when dtype in @non_list_types, do: dtype
+  def normalise_dtype({:struct, inner_types}) do
+    normalized_dtypes =
+      for {key, dtype} <- inner_types, into: %{} do
+        {key, normalise_dtype(dtype)}
+      end
+
+    if normalized_dtypes |> Map.values() |> Enum.any?(&is_nil/1) do
+      nil
+    else
+      {:struct, normalized_dtypes}
+    end
+  end
+
+  def normalise_dtype(dtype) when dtype in @scalar_types, do: dtype
   def normalise_dtype(dtype) when dtype in [:float, :f64], do: {:f, 64}
   def normalise_dtype(:f32), do: {:f, 32}
   def normalise_dtype(_dtype), do: nil
@@ -220,16 +233,11 @@ defmodule Explorer.Shared do
       Enum.reduce(list, initial_type, fn el, type ->
         new_type = type(el, type) || type
 
-        cond do
-          leaf_dtype(new_type) == :numeric and leaf_dtype(type) in [:integer, {:f, 32}, {:f, 64}] ->
-            new_type
-
-          new_type != type and type != nil ->
-            raise ArgumentError,
-                  "the value #{inspect(el)} does not match the inferred series dtype #{inspect(type)}"
-
-          true ->
-            new_type
+        if new_type_matches?(type, new_type) do
+          new_type
+        else
+          raise ArgumentError,
+                "the value #{inspect(el)} does not match the inferred series dtype #{inspect(type)}"
         end
       end)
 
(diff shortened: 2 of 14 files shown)


native/explorer/src/encoding.rs
```diff
@@ -573,6 +574,12 @@ pub fn term_from_value<'b>(v: AnyValue, env: Env<'b>) -> Result<Term<'b>, Explor
         AnyValue::Duration(v, time_unit) => encode_duration(v, time_unit, env),
         AnyValue::Categorical(idx, mapping, _) => Ok(mapping.get(idx).encode(env)),
         AnyValue::List(series) => list_from_series(ExSeries::new(series), env),
+        AnyValue::Struct(_, _, fields) => v
+            ._iter_struct_av()
```
costaraphael, line 578: This function looks very undocumented, but it is part of the public API from Polars: https://docs.rs/polars/latest/polars/datatypes/enum.AnyValue.html#method._iter_struct_av

The Python implementation deals with struct values through it as well.

test/explorer/series/struct_test.exs
```diff
@@ -0,0 +1,139 @@
...
+
+  describe "inspect/1" do
+    test "struct with integer values" do
+      s = Series.from_list([%{a: 1}, %{a: 2}])
+
+      assert inspect(s) ==
+               """
+               #Explorer.Series<
+                 Polars[2]
+                 struct[1] [%{"a" => 1}, %{"a" => 2}]
```
costaraphael, line 122: I followed the Polars approach here and only displayed the number of keys in the struct series. This made sense to me, as it would get out of hand pretty quick with large structs and nested values. 

Let me know if you disagree and believe this should be more descriptive.

lib/explorer/shared.ex
```diff
@@ -2,7 +2,7 @@ defmodule Explorer.Shared do
   # A collection of **private** helpers shared in Explorer.
   @moduledoc false
 
-  @non_list_types [
+  @scalar_types [
```
costaraphael, line 5: With struct type being added the name didn't seem right anymore. Hope I got the intention right here. 😅

lib/explorer/shared.ex
```diff
@@ -27,7 +27,7 @@ defmodule Explorer.Shared do
   within lists inside.
   """
   def dtypes do
-    @non_list_types ++ [{:list, :any}]
+    @scalar_types ++ [{:list, :any}, {:struct, :any}]
```
costaraphael, line 30: I followed a similar approach from the list type implementation and used `:any` to denote "a struct of any shape". I'm not 100% happy with it though, please let me know if you have any suggestions.

lib/explorer/shared.ex
```diff
@@ -286,38 +338,43 @@ defmodule Explorer.Shared do
...
+  def cast_numerics(list, dtype) do
     {cast_numerics_deep(list, dtype), cast_numeric_dtype_to_float(dtype)}
   end
 
-  def cast_numerics(list, type), do: {list, type}
+  defp cast_numerics_deep(list, {:struct, dtypes}) when is_list(list) do
+    Enum.map(list, fn item ->
+      Map.new(item, fn {field, inner_value} ->
+        inner_dtype = Map.fetch!(dtypes, to_string(field))
+        [casted_value] = cast_numerics_deep([inner_value], inner_dtype)
```
costaraphael, line 349: This feels kinda inefficient (it's going to wrap/run through `Enum.map`/unwrap each value of the list of maps), but I was not too worried about it since I don't see `Series.from_list` as being the main way people will interact with this dtype. I feel like building a struct series inside a dataframe, or reading a struct series from an external file would be far more common.

Let me know if you disagree though.

lib/explorer/shared.ex
```diff
@@ -277,6 +304,31 @@ defmodule Explorer.Shared do
...
 
+  defp new_type_matches?(type, new_type)
+
+  defp new_type_matches?(nil, _new_trpe), do: true
+
+  defp new_type_matches?({:struct, types}, {:struct, new_types}) do
+    types
+    |> Map.keys()
+    |> Enum.all?(fn key ->
+      type = Map.get(types, key)
```
josevalim, line 315:
```suggestion
    types
    |> Enum.all?(fn {key, type} ->
```

lib/explorer/shared.ex
```diff
@@ -277,6 +304,31 @@ defmodule Explorer.Shared do
...
+      end
+    end)
+  end
+
+  defp new_type_matches?(type, new_type) do
+    if leaf_dtype(new_type) == :numeric and leaf_dtype(type) in [:integer, {:f, 32}, {:f, 64}] do
+      true
+    else
+      type == new_type
+    end
```
josevalim, line 329: Define a `  defp new_type_matches?(type, type), do: true` at the top and this code can become:

```suggestion
    leaf_dtype(new_type) == :numeric and leaf_dtype(type) in [:integer, {:f, 32}, {:f, 64}]
```

lib/explorer/shared.ex
```diff
@@ -37,7 +37,20 @@ defmodule Explorer.Shared do
     if maybe_dtype = normalise_dtype(inner), do: {:list, maybe_dtype}
   end
 
-  def normalise_dtype(dtype) when dtype in @non_list_types, do: dtype
+  def normalise_dtype({:struct, inner_types}) do
+    normalized_dtypes =
+      for {key, dtype} <- inner_types, into: %{} do
+        {key, normalise_dtype(dtype)}
+      end
```
josevalim, line 44: You can use `Enum.reduce_while` and half if normalise_dtype returns nil, doing this traversal only once.

josevalim (approved):
Looks great to me, only some minor nits!

lib/explorer/shared.ex
costaraphael: Thanks, looks a lot cleaner now!

Added in [53e3e87](https://github.com/elixir-explorer/explorer/pull/756/commits/53e3e8727369475c78c9242267dcc1867c0192f9)

lib/explorer/shared.ex
costaraphael: Done here: [53e3e87](https://github.com/elixir-explorer/explorer/pull/756/commits/53e3e8727369475c78c9242267dcc1867c0192f9)

lib/explorer/series.ex
```diff
@@ -92,11 +94,13 @@ defmodule Explorer.Series do
           | :integer
           | :string
           | list_dtype
+          | struct_dtype
 
   @type time_unit :: :nanosecond | :microsecond | :millisecond
   @type datetime_dtype :: {:datetime, time_unit}
   @type duration_dtype :: {:duration, time_unit}
   @type list_dtype :: {:list, dtype()}
+  @type struct_dtype :: {:struct, %{String.t() => dtype()}}
```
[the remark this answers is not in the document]
billylanchantin: Hi! I really like this PR and think this missing functionality would be very helpful.

That said, I'm hesitant about the term "struct" since as you say it's overloaded. What do you think about `DF.from_rows`? Then it would pair with [`DF.to_rows`](https://hexdocs.pm/explorer/Explorer.DataFrame.html#to_rows/2) just like `Series.from_list`/`Series.to_list`.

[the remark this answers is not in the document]
josevalim: So this is not importing data from Elixir structs, but literally having a struct column inside Explorer. You could even have lists of structs. :)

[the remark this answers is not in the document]
billylanchantin: 1. Whoops! Thank you, I was skimming 😅 
2. I still worry that "struct" is overloaded ("map" may be better?), but now less so. It's sorta like how we have `{:f, 32}` even though there's no real equivalent in Elixir.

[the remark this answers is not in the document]
costaraphael: Yeah, I was a bit iffy about naming it `struct` as well because of the pre-existing meaning within Elixir, but also naming it the more Elixir-appropriate `map` would mean people used to Polars would have to keep this "translation" in mind when using Explorer.

Feels like a "pick your poison" type of choice 😅

lib/explorer/series.ex
```diff
@@ -16,6 +16,8 @@ defmodule Explorer.Series do
     * `:time` - Time type that unwraps to `Elixir.Time`
     * `{:list, dtype}` - A recursive dtype that can store lists. Examples: `{:list, :integer}` or
       a nested list dtype like `{:list, {:list, :integer}}`.
+    * `{:struct, %{key => dtype}}` - A recursive dtype that can store structs. Examples:
+      `{:struct, %{"a" => :integer}}` or a nested struct dtype like `{:struct, %{"a" => {:struct, %{"b" => :integer}}}}`.
```
philss, line 20: I think we could mention that `:struct` is not related to Elixir structs, but to [Arrow](https://arrow.apache.org/docs/format/Columnar.html#struct-layout)/Polars structs.

lib/explorer/series.ex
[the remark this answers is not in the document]
philss: I think we could solve the naming issue by documenting it (like I mentioned in my other comment). Since we are following Arrow data structs, we could just document that. WDYT?

lib/explorer/shared.ex
```diff
@@ -277,6 +305,25 @@ defmodule Explorer.Shared do
     dtype_from_list!(items, leaf_dtype(type))
   end
 
+  defp new_type_matches?(type, new_type)
+
+  defp new_type_matches?(type, type), do: true
+
+  defp new_type_matches?(nil, _new_trpe), do: true
```
philss, line 312:
```suggestion
  defp new_type_matches?(nil, _new_type), do: true
```

native/explorer/src/encoding.rs
philss: I see. Maybe it's a unstable API, since normally it would not have the `_` at the beginning. We can ask them later if this should be used, or if there is an alternative.

test/explorer/series/struct_test.exs
```diff
@@ -0,0 +1,147 @@
...
+    s = Series.from_list([%{a: %{b: 1}, c: [2]}, %{a: %{b: 2}, c: [4]}])
+
+    assert inspect(s) ==
+             """
+             #Explorer.Series<
+               Polars[2]
+               struct[2] [%{"a" => %{"b" => 1}, "c" => [2]}, %{"a" => %{"b" => 2}, "c" => [4]}]
+             >\
+             """
+  end
```
philss, line 146: I was thinking about a crazy scenario where people try something like `{:list, {:struct, any()}` :sweat_smile: 
Would this work? If so, would you mind to add a test case for that?

lib/explorer/series.ex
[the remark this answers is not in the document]
billylanchantin: @philss Yeah if struct aligns with polars and arrow then it's the right call 👍 And documenting the name in case of potential confusion is def the way to go.

lib/explorer/series.ex
costaraphael: That's nice! I'll make an edit adding this later.

test/explorer/series/struct_test.exs
costaraphael: When you say `{:list, {:struct, any()}}` do you mean mixing datatypes? Something like:

```elixir
Series.from_list([[%{a: 1}, %{a: "b"}]])
# or
Series.from_list([[%{a: 1}], [%{a: "b"}]])
```

If so, this should give a type error when building the series. I can add a test to make sure that's what happens, just in case 😅

philss: The idea would be to have a nested dtype mixing struct and list. It should be valid:

```elixir
Series.from_list([[%{a: 1}, %{a: 2}]])
# or
Series.from_list([[%{a: 1}], [%{a: 2}]])
```

This should give a `{:list, {:struct, %{"a" => :integer}}}` dtype, right?

costaraphael: Oh yeah, this is valid and should work. I can add a test for it just in case (I think I only have the "list inside of a struct" case covered).

costaraphael: Done here: [a5eb947](https://github.com/elixir-explorer/explorer/pull/756/commits/a5eb947fcaac4a9d09e6607bec76829a652275a2)

lib/explorer/series.ex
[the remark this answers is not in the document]
costaraphael: I've added the extra bits to the docs here: e565494

WDYT?

josevalim:
Thank you! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

I assume the next step is an API for adding/updating/removing struct fields? :D

costaraphael:
> I assume the next step is an API for adding/updating/removing struct fields? :D

Something like that!

My next target would be to allow breaking a struct series into its component series (`DataFrame.unnest/2`) and building a struct from existing series (`Series.build_struct/1`). These two operations would unblock any workflow regarding parsing/exporting data with nested values.

Operations within struct series are kinda limited in Polars after that, but the interesting features I can see being added:

- Plucking a single field as a series (`Series.field/2`)
- Renaming struct fields (`Series.rename_fields/2`)
- JSON encoding/decoding (`Series.json_(encode|decode)/1`)

Should I open an issue to keep track of this?

josevalim:
No need for an issue, i think for these it is fine to contribute as the need arises :)

PR #760, opened 2023-12-06
Hopefully improves the description of DataFrame.arrange_with()

kellyfelkins:
The prior description of DataFrame.arrange_with appears to be missing the word "not".

With less confidence, I thought the phrasing could be improved. This is less important than adding "not".

I'm offering this because one of the things that contributes to my experience with Elixir is the wonderful, quality documentation.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -3150,9 +3150,8 @@ defmodule Explorer.DataFrame do
   @doc """
   Arranges/sorts rows by columns using a callback function.
 
-  The callback receives a lazy dataframe. A lazy dataframe does
-  hold any values, instead it stores all operations in order to
-  execute all sorting performantly.
+  The callback receives a lazy dataframe. A lazy dataframe does not
+  hold values; instead it stores operations for efficient sorting.
 
   This is a callback version of `arrange/2`.
 


kellyfelkins:
The previous version joined 2 sentences with ", instead". This didn't sound right to me, so I did some searching and found these:

https://www-s3-live.kent.edu/s3fs-root/s3fs-public/file/commas-to-combine-sentences.pdf?VersionId=64YStJA56RH9TwmLQE8r1OaZq2gXcYPB#:~:text=Commas%20alone%20cannot%20join%20two,of%20equal%20weight%20and%20importance.&text=Comma%20splice%20refers%20to%20two,without%20the%20following%20coordinating%20conjunction.

https://www.reddit.com/r/grammar/comments/8jjoza/butinstead/#:~:text=In%20that%20sense%2C%20but%20is,semicolon%20or%20a%20coordinating%20conjunction).

lib/explorer/data_frame.ex
```diff
@@ -3150,9 +3150,8 @@ defmodule Explorer.DataFrame do
   @doc """
   Arranges/sorts rows by columns using a callback function.
 
-  The callback receives a lazy dataframe. A lazy dataframe does
-  hold any values, instead it stores all operations in order to
-  execute all sorting performantly.
+  The callback receives a lazy dataframe. A lazy dataframe does not
+  hold values; instead it stores operations for efficient sorting.
```
josevalim, line 3154: TIL!

I think using `; instead` increases the barrier for contribution though... as we are going into advanced english. Could we use something else? "rather"? remove "instead" altogether? Thoughts?

billylanchantin: If both ";" and "," trip folks up for one reason or another, here's an alternative:

> The callback receives a lazy dataframe which stores operations instead of values for efficient sorting.

And this is off topic, but we (I) might want to consider writing a guide on lazy frames. It's a somewhat complex topic. If you come from Pandas there's no equivalent AFAIK.

kellyfelkins: > TIL!
> 
> I think using `; instead` increases the barrier for contribution though... as we are going into advanced english. Could we use something else? "rather"? remove "instead" altogether? Thoughts?

I think the semicolon is correct, but I normally use a dash or emdash....they seem to make more sense to me.

> A lazy dataframe does not hold values–instead it stores operations for efficient sorting.

But I like Billy's suggestion.

josevalim, line 3154:
```suggestion
  The callback receives a lazy dataframe which stores
  operations instead of values for efficient sorting.
```

Issue #730, opened 2023-11-08
Additional `_` macros in `Series`


billylanchantin:
### Overview

With https://github.com/elixir-explorer/explorer/pull/728 merged, we now have a strategy for adding macros to Series. The aim is closer parity with DataFrames.

This issue follows up from https://github.com/elixir-explorer/explorer/issues/726. We need to decide:

* which DataFrame macros we want Series to also support and
* how to name them.

### Current Macros

DataFrames support four macros/callbacks:

* `arrange`/`arrange_with`
* `filter`/`filter_with`
* `mutate`/`mutate_with`
* `summarise`/`summarise_with`

We've already done `filter`. We probably don't want to support `summarise`. That leaves `arrange` and `mutate`.

### Name options

DF | Parity option | Rename option
:-- | :-- | :--
`DF.arrange{_with}` | `S.arrange{_with}` | `S.sort{_with}`
`DF.mutate{_with}` | `S.mutate{_with}` | `S.map{_with}`

### Considerations

1. There already exists an `S.sort` function, and it allows an additional `nils: :first | :last` option.
2. There exists an `S.transform` function which is similar to `DF.mutate`, but it's for Elixir functions instead of Series functions.

cigrainger:
I'm not a huge fan of the idea of `mutate` for series. Maybe it's just the R in me. `mutate` carries the create/delete aspects of mutating a dataframe to me and it's been around forever in R. I think I could get past it though for continuity. 

We should just bring the nils opt to `DF.sort` for parity, no? Or is it not available in the expression?

billylanchantin:
I'm leaning toward `map` as well.

Like you say, it doesn't carry any baggage over from R. And the parallels to `Enum` make it approachable for Elixir devs. Heck, we could probably document `DF.mutate` in terms of `S.map` and newcomers might even pick it up quicker.

> We should just bring the nils opt to `DF.sort` for parity, no? Or is it not available in the expression?

(`DF.arrange`?) Yes I believe it's available. Though I'm still getting used to Rust TBH.

[...]

PR #762, opened 2023-12-07
Add `Series.sort_by` and `Series.sort_with`

billylanchantin:
Adds:

* `Series.arrange` (macro)
* `Series.arrange_with` (function)

Closes:

* https://github.com/elixir-explorer/explorer/issues/730

I'm opening this up early. There are two decisions we need to make before I write docs:

* **Names:**
  * `Series.arrange{_with}`
  * `Series.sort{_with}` (already taken)
* **API:**
  * `Series.arrange(s, _, direction: :asc)` (matches `Series.sort`)
  * `Series.arrange(s, asc: _)` (matches `DataFrame.arrange`)

### DIFF
--- a/test/explorer/series_test.exs
+++ b/test/explorer/series_test.exs
@@ -2853,7 +2853,7 @@ defmodule Explorer.SeriesTest do
       require Explorer.Series
 
       s1 = Series.from_list([1, 2, 3])
-      result = Series.arrange(s1, remainder(_, 3))
+      result = Series.sort_by(s1, remainder(_, 3))
       assert Series.to_list(result) == [3, 1, 2]
     end
 
@@ -2861,7 +2861,7 @@ defmodule Explorer.SeriesTest do
       require Explorer.Series
 
       s1 = Series.from_list([1, 2, 3])
-      result = Series.arrange(s1, remainder(_, 3), direction: :desc)
+      result = Series.sort_by(s1, remainder(_, 3), direction: :desc)
       assert Series.to_list(result) == [2, 1, 3]
     end
 
@@ -2869,7 +2869,7 @@ defmodule Explorer.SeriesTest do
       require Explorer.Series
 
       s1 = Series.from_list([1, nil, 2, 3])
-      result = Series.arrange(s1, remainder(_, 3))
+      result = Series.sort_by(s1, remainder(_, 3))
       assert Series.to_list(result) == [3, 1, 2, nil]
     end
 
@@ -2877,7 +2877,7 @@ defmodule Explorer.SeriesTest do
       require Explorer.Series
 
       s1 = Series.from_list([1, nil, 2, 3])
-      result = Series.arrange(s1, remainder(_, 3), nils: :first)
+      result = Series.sort_by(s1, remainder(_, 3), nils: :first)
       assert Series.to_list(result) == [nil, 3, 1, 2]
     end
   end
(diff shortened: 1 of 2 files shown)


philss (approved):
LGTM!

For the name, I don't have a strong opinion about it. I think it would be nice to keep the `sort` naming, since it seems more clear in the context of the series (and matches `Enum`, as you said in the issue). But again, I'm fine with the `arrange` naming :)

For the API, I think the version you did here seems better. I would change the `arrange` from the DF to follow the same API.

billylanchantin:
@philss Thanks for looking!

> I think it would be nice to keep the sort naming

Agreed. But can we without a breaking change? Today way support:

```elixir
s = Explorer.Series.from_list([2, 1, 3])
Explorer.Series.sort(s, direction: :asc)
```

If we want to also support:

```elixir
Explorer.Series.sort(s, remainder(_, 3))
```

There would be a conflict for the `sort/2` namespace.

billylanchantin:
We could also do `sort_by` and `sort_with`.

(Or the cursed `sort_by` and `sort_by_with` 😅)

philss:
> There would be a conflict for the sort/2 namespace.

Yeah, I see :/

> We could also do `sort_by` and `sort_with`.

It's actually a good idea IMHO :D

josevalim:
I like sort_by a lot and had the same idea. Which means we should pass direction as option for symmetry with `sort`.

I am not ecstatic about sort_with. We could just allow a function in `sort`, but that would differ from map/filter, so I don’t know.

cigrainger:
I agree -- I like sort_by. Generally I'm keen to use sort consistently with DFs as well and rename arrange -> sort.

josevalim:
But would we make them all consistent? Arrange? Mutate?

To be honest, I dislike mutate, but it would be a huge departure for the library. I think rooting in R or Pandas make sense, and the general agreement is that R wins API-elegance wise.

cigrainger:
I still think mutate doesn't really apply to series so there's nothing to make consistent there. What else would you call it?

100% R wins elegance-wise but it's not necessarily in the names -- in fact, R has a lot of 'creative' naming to avoid namespace conflicts with base. The elegance comes from matching 'verbs' to (tidy data) principles and maintaining parsimony. Which I think we're doing well! I'm pretty sure arrange is one of the cases of avoiding namespace conflict in R.

josevalim:
For me a mutate could be a map, transform, or anything else that does not imply an in-memory change, but it is not a hill I would die on.

if we are renaming arrange, would we call it sort or sort_by?

cigrainger:
Yeah no I totally agree. But mutate in R has always been a bit weird. It's specifically about changing a data frame in some way. Adding, removing, or changing columns.

I'd just call it sort. I'm totally cool with unifying on arrange for both too. I mainly think they should be consistent. I just don't want to skip out on the more obvious `sort` just to maintain a naming convention from R that was an artifact of necessity rather than choice.

josevalim:
Although in one we do asc:/desc: and the other does direction: :asc/:desc, so they wouldn’t quite mirror.

billylanchantin:
Update:

* Names:
    * `sort_by`, `sort_with`
* API
    * `sort_by(s, -1 * _, direction: :asc)`

I've also found some mild inconsistencies in the sorting API (like `stable` is available on `sort_by` but not `sort`). But I think it'd be better to raise an issue and address that as part of a holistic pass that normalizes the entire sorting API.

josevalim:
Yeah, I think there are two issues next: making sure the options are consistent and then renaming DF.arrange.

test/explorer/series_test.exs
```diff
@@ -2848,6 +2848,66 @@ defmodule Explorer.SeriesTest do
     end
   end
 
+  describe "arrange/2" do
```
josevalim, line 2851:
```suggestion
  describe "sort_by/2" do
```

perhaps you need another describe for sort_with.

test/explorer/series_test.exs
```diff
@@ -2848,6 +2848,66 @@ defmodule Explorer.SeriesTest do
...
+    test "nils first" do
+      require Explorer.Series
+
+      s1 = Series.from_list([1, nil, 2, 3])
+      result = Series.sort_by(s1, remainder(_, 3), nils: :first)
+      assert Series.to_list(result) == [nil, 3, 1, 2]
+    end
+  end
+
+  describe "arrange_with/2" do
```
josevalim, line 2885:
```suggestion
  describe "sort_with/2" do
```

PR #767, opened 2023-12-12
Enables alternative correlation methods

cristineguadelupe:
I'm passing the correlation method as string to follow the `Series.rank` convention but I'd suggest to use atoms instead. WDYT?

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -5678,9 +5678,9 @@ defmodule Explorer.DataFrame do
   * `:column_name` - the name of the column with column names. Defaults to "names".
   * `:ddof` - the 'delta degrees of freedom' - the divisor used in the correlation
     calculation. Defaults to 1.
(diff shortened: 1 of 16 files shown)


lib/explorer/data_frame.ex
```diff
@@ -5678,6 +5678,9 @@ defmodule Explorer.DataFrame do
   * `:column_name` - the name of the column with column names. Defaults to "names".
   * `:ddof` - the 'delta degrees of freedom' - the divisor used in the correlation
     calculation. Defaults to 1.
+  * `method` refers to the correlation method. The following methods are available:
```
josevalim, line 5681:
```suggestion
  * `:method` refers to the correlation method. The following methods are available:
```

josevalim:
Looks great but indeed atoms will be better. I think you need to declare an enum in Rust. Ask @philss, he will know for sure!

philss (approved):
:shipit:

josevalim:

PR #769, opened 2023-12-12
Add more integer dtypes

philss:
This PR adds support for signed and unsigned integer datatypes. There are now 8 different dtypes for integers, and they have 8, 16, 32 or 64 bits in size.

The idea is to follow the same naming from [Nx](https://github.com/elixir-nx/nx). `{:s, size}` is for signed integers, and `{:u, size}` is for the unsigned.

In this PR I'm adding only the new dtypes. The `:integer` dtype remains, and I should "rename" it - to `{:s, 64}` - in another PR.

---
PS: there are some work to be done in the area of projecting the dtypes of "out dfs" that I plan to do in another PR as well.


josevalim (approved):

josevalim:
Also don't forget to add to your todo list the refactoring of iotype. :)

Issue #765, opened 2023-12-11
Ensure all options for sorting functions are available


billylanchantin:
## Description

We should expose all the options for the sorting functions where possible.

## Background 

After https://github.com/elixir-explorer/explorer/pull/762, we now have a number of sorting related functions:

* `DataFrame.arrange/3`
* `DataFrame.arrange_with/3`
* `Series.argsort/2`
* `Series.sort/2`
* `Series.sort_by/3`
* `Series.sort_with/3`

These sorting functions (generally) have the options which are available on the [`SortOptions`](https://docs.rs/polars/latest/polars/chunked_array/ops/struct.SortOptions.html) struct:

```rust
pub struct SortOptions {
    pub descending: bool,
    pub nulls_last: bool,
    pub multithreaded: bool,
    pub maintain_order: bool,
}
```

We've made these options available in the following way:

Explorer | Choices | Polars Boolean | Description
:-- | :-- | :-- | :--
`direction` | `:asc \| :desc` | `descending` | Sort ascending or descending
`multithreaded`<sup>*</sup> | `false \| true` | `multithreaded` | Compute in parallel
`nils` | `:first \| :last` | `nulls_last` | Put `nil`s first or last in result
`stable` | `false \| true` | `maintain_order` | Maintain the order of ties

<sup>*</sup>Note: we've not actually exposed `multithreaded` anywhere yet, but there's a TODO.

However, we've not done so in a consistent manner. Sometimes only some of the options are available.

billylanchantin:
Closed by: https://github.com/elixir-explorer/explorer/pull/771

PR #771, opened 2023-12-14
Normalize sorting options

billylanchantin:
Closes:

* https://github.com/elixir-explorer/explorer/issues/765

Each of these options are now available on all the sorting functions:

* `direction: :asc | :desc`
  * This takes the form of keys in the first arg on the DF functions
* `nils: :first | :last`
* `parallel: true | false`
  * This isn't available on the lazy functions
* `stable: true | false`

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -3057,29 +3057,15 @@ defmodule Explorer.DataFrame do
 
     * `:parallel` - boolean.
       Whether to parallelize the sorting.
-      By default it is `false`.
+      By default it is `true`.
 
-      > #### Notice {: .notice}
-      >
-      > The `:parallel` option is applicable under one condition: all lazy series must reduce to
-      > `op: :column`. E.g.
-      >
-      > ```elixir
-      > df = Explorer.DataFrame.new(a: [1, 2, 3])
-      > Explorer.DataFrame.arrange(df, desc: a)
-      > ```
-      >
-      > If any additional computation takes place, e.g.
-      >
-      > ```elixir
-      > Explorer.DataFrame.arrange(df, desc: 1 / a)
-      > ```
-      > parallelism is not available and this option is ignored.
+      Parallel sort isn't available on certain lazy operations.
+      In those situations this option is ignored.
 
     * `:stable` - boolean.
       Determines if the sorting is stable (ties are guaranteed to maintain their order) or not.
       Unstable sorting may be more performant.
-      By default it is `true`.
+      By default it is `false`.
 
   ## Examples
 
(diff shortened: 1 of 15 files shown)


lib/explorer/data_frame.ex
```diff
@@ -3052,14 +3052,34 @@ defmodule Explorer.DataFrame do
...
 
-    * `:stable` - `boolean()`.
+    * `:parallel` - boolean.
+      Whether to parallelize the sorting.
+      By default it is `false`.
+
+      > #### Notice {: .notice}
+      >
+      > The `:parallel` option is applicable under one condition: all lazy series must reduce to
+      > `op: :column`. E.g.
```
billylanchantin, line 3065: I had trouble explaining/wording this notice.

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -643,22 +643,44 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
+        |> Enum.unzip()
+
+      Shared.apply_dataframe(df, out_df, :df_arrange, [
+        column_names,
+        directions,
+        maintain_order?,
+        multithreaded?,
+        nulls_last?,
+        df.groups
+      ])
```
billylanchantin, line 669: We weren't ever actually calling `df_arrange` from the Elixir side. You can think of this clause as basically an optimization.

lib/explorer/shared.ex
```diff
@@ -475,4 +475,58 @@ defmodule Explorer.Shared do
     |> Enum.sort(&(elem(&1, 1) <= elem(&2, 1)))
     |> Enum.map(fn {_, key} -> ["      * ", inspect(key), ?\n] end)
   end
+
+  @doc """
+  Validate options to be used for the various sorting functions.
+  """
+  def validate_sort_options!(opts) do
```
billylanchantin, line 482: Someone might wanna double check my logic on this one 😅

lib/explorer/data_frame.ex
philss: I like this way with the example!

philss (approved):
LGTM!

josevalim (commented):
Beautiful work. The only suggestion I have is about the defaults.

Numpy and Pandas do not use stable sorting by default nor Polars. Polars is also parallel by default, which I believe we should try to be as well. So my suggestion is to make stable default to false and parallel default to true. This way, we can also simplify the explanation of parallel. We could say "Use parallel algorithms when possible. Set it to false to disable it.".

Thoughts?

billylanchantin:
@josevalim That sounds great. I was keeping the defaults we had, but I hadn't thought about what they _ought_ to be. I can definitely have `:stable` default to `false` and `:parallel` default to `true`.

> This way, we can also simplify the explanation of parallel. We could say "Use parallel algorithms when possible. Set it to false to disable it.".

That's a good tactic: describe the philosophy instead of belaboring the implementation (which may change). I think I'll still add a sentence about parallelism not always being available on lazy operations. That way the reader will have some idea of what we mean by "when possible".

lib/explorer/data_frame.ex
billylanchantin: Thanks! I ended up dropping it for simpler wording: [3c77fe5](https://github.com/elixir-explorer/explorer/pull/771/commits/3c77fe564f05a0f12a448e1e83ace922537259a6)

But if enough users say they're confused, we can resurrect that notice.

josevalim (approved):

Issue #766, opened 2023-12-11
Revist the names of `DataFrame.arrange/2` and `DataFrame.arrange_with/2`


billylanchantin:
There's been some discussion of renaming the `arrange` functions to `sort`. E.g. here:

* https://github.com/elixir-explorer/explorer/pull/762#issuecomment-1846552261

This would align better with the `Series` function names. Plus, the `arrange` names come from R where they were chosen to avoid namespace conflicts. We don't have to worry about that here.

However, this would be a breaking change. This issue is to hash out that discussion.

josevalim:
I am happy to roll with `sort_by` and `sort_with`. We should keep `arrange` and `arrange_with` around with `@deprecated "foobar"` annotations. They can simply delegate to the new `sort_by` and `sort_with`.

cigrainger:
Yep agreed!

PR #777, opened 2023-12-16
Rename `DataFrame.arrange` as `DataFrame.sort`

billylanchantin:
Renames:

* `DataFrame.arrange` -> `DataFrame.sort_by`
* `DataFrame.arrange_with` -> `DataFrame.sort_with`

Closes:

* https://github.com/elixir-explorer/explorer/issues/766

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -162,7 +162,7 @@ defmodule Explorer.Backend.DataFrame do
   @callback mask(df, mask :: series) :: df
   @callback filter_with(df, out_df :: df(), lazy_series()) :: df
   @callback mutate_with(df, out_df :: df(), mutations :: [{column_name(), lazy_series()}]) :: df
-  @callback arrange_with(
+  @callback sort_with(
               df,
               out_df :: df(),
               directions :: [{:asc | :desc, lazy_series()}],
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -61,7 +61,7 @@ defmodule Explorer.DataFrame do
   - `select/2` for picking columns and `discard/2` to discard them
   - `filter/2` for picking rows based on predicates
   - `mutate/2` for adding or replacing columns that are functions of existing columns
-  - `arrange/2` for changing the ordering of rows
+  - `sort_by/2` for changing the ordering of rows
   - `distinct/2` for picking unique rows
   - `summarise/2` for reducing multiple rows down to a single summary
   - `pivot_longer/3` and `pivot_wider/4` for massaging dataframes into longer or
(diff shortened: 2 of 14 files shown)


billylanchantin:
### Naming

There's a naming issue I hadn't considered. So far I've named the DataFrame functions to match their Series counterparts:

Type       | Series             | DataFrame
:--        | :--                | :--
Function   | `Series.sort`      | N/A
Expression | `Series.sort_by`   | `DataFrame.sort_by`
Callback   | `Series.sort_with` | `DataFrame.sort_with`

But this is how I think we should name them instead:

Type       | Series               | DataFrame
:--        | :--                  | :--
Function   | `Series.sort_values` | N/A
Expression | `Series.sort`        | `DataFrame.sort`
Callback   | `Series.sort_with`   | `DataFrame.sort_with`

That change would be breaking because `Series.sort` would change meaning. (We discussed this [here](https://github.com/elixir-explorer/explorer/pull/762).)

However, I hadn't considered "five main verbs" angle. As it stands, this is how that section of "Exploring Explorer" reads:

> In `Explorer`, like in `dplyr`, we have five main verbs to work with dataframes:
> * select
> * filter
> * mutate
> * **sort_by**
> * summarise

Now we could just change that bullet to "sort" and be done with it. But each of the other four follows the:

* `Explorer.DataFrame.verb` (expression)
* `Explorer.DataFrame.verb_with` (callback)

pattern. Since we're pre-1.0, I think it's worth breaking `Series.sort` to remain consistent.

As a bonus, there'd be an opportunity for a `DataFrame` analog of `Series.sort_values`. Maybe something like:

```elixir
DataFrame.sort_columns(asc: "column_name")
```

WDYT? /cc @josevalim @philss @cigrainger

josevalim (approved):

philss:
@billylanchantin I'm more inclined to the version we have in this PR - the `sort` / `sort_by` / `sort_with`. But I'm not against the proposed new names! And I think it's fine to break the API at this point.

philss (approved):

jcmkk3:
I'm just a dataframe geek that likes to keep an eye on this project, but I don't find `sort_by` to be a problem. There is also `group_by`, which is another verb that provides context to subsequent operations.

billylanchantin:
Thanks @philss! I'm gonna wait for @cigrainger to weigh in too just in case. He was who originally brought up changing the `arrange` choice.

cigrainger:
Hey! Sorry for weighing in late. I don't think `sort_by` is a problem. It's also 100% fine to break stuff at this point in time, so don't let that bound you in. We should focus on getting it right. Thanks @jcmkk3 I agree totally. And I think the `_by` suffix makes sense in a dataframe.

cigrainger (approved):
<3

billylanchantin:
Thanks all! If no one objects to `_by`, then I don't either.

Also sorry for the snub, @jcmkk3! I never saw your input. Turns out we commented within 60s of each other 😅 I think `group_by` is a great parallel to draw.

PR #785, opened 2023-12-22
Rename `:integer` dtype to `{:s, 64}`

philss:
This PR is a followup of https://github.com/elixir-explorer/explorer/pull/769.

The main goal of this PR is to "rename" the main integer dtype to `{:s, 64}` and introduce the other signed and unsigned dtypes.

There is still a big work to do regarding making the operations compatible when using different types of series.
My idea is to add support for it in smaller PRs.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2810,7 +2810,7 @@ defmodule Explorer.DataFrame do
                     "and then `Explorer.DataFrame.put/3` to add the series to your dataframe."
 
           number when is_number(number) ->
-            dtype = if is_integer(number), do: :integer, else: {:f, 64}
+            dtype = if is_integer(number), do: {:s, 64}, else: {:f, 64}
             lazy_s = LazySeries.new(:lazy, [number], dtype)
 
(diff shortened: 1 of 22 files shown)


lib/explorer/data_frame.ex
```diff
@@ -3019,7 +3019,13 @@ defmodule Explorer.DataFrame do
...
     opts = Keyword.validate!(opts, [:dtype])
     name = to_column_name(column_name)
-    dtype = opts[:dtype]
+
+    dtype =
+      if opts[:dtype] in [nil, :infer] do
+        opts[:dtype]
+      else
+        Shared.normalise_dtype!(opts[:dtype])
+      end
```
josevalim, line 3028: We are checking for dtype being nil or infer in 3 diofferent places now. Maybe we can move this normalization to line 3039 (a couple lines below?).

josevalim (approved):
Just one comment and ship it!

lib/explorer/data_frame.ex
philss: Something like this? 94daa48

PR #790, opened 2023-12-29
Support more integer dtypes in `Series.from_binary/2`

philss:
This is a continuation of #769 and #785.


lib/explorer/polars_backend/shared.ex
```diff
@@ -186,9 +186,30 @@ defmodule Explorer.PolarsBackend.Shared do
...
+        Native.s_from_binary_i8(name, binary)
+
+      {:s, 16} ->
+        Native.s_from_binary_i16(name, binary)
+
+      {:s, 32} ->
+        Native.s_from_binary_i32(name, binary)
+
       {:s, 64} ->
         Native.s_from_binary_i64(name, binary)
```
josevalim, line 199: I wonder if it is worth having a single function where we pass the integer or the size as argument? Not sure what is easier on the Rust side.

lib/explorer/shared.ex
```diff
@@ -64,11 +64,11 @@ defmodule Explorer.Shared do
 
   def normalise_dtype(dtype) when dtype in @scalar_types, do: dtype
   def normalise_dtype(dtype) when dtype in [:float, :f64], do: {:f, 64}
-  def normalise_dtype(dtype) when dtype in [:integer, :i64], do: {:s, 64}
+  def normalise_dtype(dtype) when dtype in [:integer, :i64, :s64], do: {:s, 64}
```
josevalim, line 67: We don't have i8/16/32/64 in Nx, so we shouldn't expose them here. It should be s8/s16/s32/s64 (it seems I missed this in a previous PR, apologies).

lib/explorer/polars_backend/shared.ex
josevalim, line 199: Let's rename these to s64/s32 and friends. We don't use i32 in Nx.

lib/explorer/shared.ex
philss: No problem! I changed it to follow the `s*` naming.

josevalim (approved):

Issue #791, opened 2023-12-29
Question/Possible improvement about the encoding option in the DataFrame.from_csv/2 function


Jhonatannunessilva:
- The documentation for the function [DataFrame.from_csv/2](https://hexdocs.pm/explorer/Explorer.DataFrame.html#from_csv/2-options) is missing an explanation of the "encoding" option.
- And even if this encoding option is passed, the code will only accept the "utf8-lossy" option, as any other string will be considered as "utf8": [io.rs](https://github.com/elixir-explorer/explorer/blob/5c51c26fc70f9a4ac3a6eeefd3c8c1241e5ceeec/native/explorer/src/dataframe/io.rs#L44-L47).
- WDYT? Is it worth including this information in the DataFrame.from_csv/2 docs?

philss:
@Jhonatannunessilva unfortunately we can only accept the [options provided by Polars](https://github.com/pola-rs/polars/blob/b44adb9a27607c19689bffcf237e49327c9087cb/crates/polars-io/src/csv/read.rs#L10). But you are right, we are missing docs around that! Please go ahead if you plan to open a PR for this! :D

PR #770, opened 2023-12-13
Rank method as atom

cristineguadelupe:

PR #793, opened 2024-01-01
Improve documentation about encoding option

Jhonatannunessilva:
This PR resolves https://github.com/elixir-explorer/explorer/issues/791

After this PR:

DataFrame.from_csv/2:

image

DataFrame.load_csv/2:

image

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -543,6 +543,9 @@ defmodule Explorer.DataFrame do
 
     * `:lazy` - force the results into the lazy version of the current backend.
 
+    * `:encoding` - Encoding to use when reading the file. For now, the only possible values are `utf8` and `utf8-lossy`.
+      The utf8-lossy option means that invalid utf8 values are replaced with � characters. (default: `"utf8"`)
+
   """
   @doc type: :io
   @spec from_csv(filename :: String.t() | fs_entry(), opts :: Keyword.t()) ::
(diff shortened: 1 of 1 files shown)


lib/explorer/data_frame.ex
```diff
@@ -627,6 +630,7 @@ defmodule Explorer.DataFrame do
     * `:eol_delimiter` - A single character used to represent new lines. (default: `"\n"`)
     * `:backend` - The Explorer backend to use. Defaults to the value returned by `Explorer.Backend.get/0`.
     * `:lazy` - force the results into the lazy version of the current backend.
+    * `:encoding` - Encoding to use when reading the file. For now, the only possible values are `utf8` and `utf8-lossy`. The utf8-lossy option means that invalid utf8 values are replaced with � characters. (default: `"utf8"`)
```
josevalim, line 633: @philss Should we make them atoms instead of regular strings?

josevalim, line 633:
```suggestion
    * `:encoding` - Encoding to use when reading the file. For now, the only possible values are
      `utf8` and `utf8-lossy`. The utf8-lossy option means that invalid utf8 values are
      replaced with � characters. (default: `"utf8"`)
```

philss: @josevalim yeah, I think so! 

@Jhonatannunessilva there is an example of what we want to do in https://github.com/elixir-explorer/explorer/pull/770
But if you prefer, we can do in another PR as well :)

Jhonatannunessilva: No problem, I can do it in this PR

Issue #729, opened 2023-11-04
cast datetime


sehHeiden:
Want to cast:
```elixir
S.cast(df["date_string"], {:datetime, :millisecond})
```

Input is:
```
#Explorer.Series<
  Polars[4563]
  string ["2023-08-29T17:39:43", "2023-08-29T17:20:09", "2023-08-29T16:53:00",
   "2023-08-29T15:38:00", "2023-08-29T16:56:49", "2023-08-29T16:30:13", "2023-08-29T16:21:20",
   "2023-08-29T15:49:26", "2023-08-29T15:48:53", "2023-08-29T15:01:57", "2023-08-29T14:56:54",
   "2023-08-29T14:55:55", "2023-08-29T14:52:09", "2023-08-29T14:17:28", "2023-08-29T14:09:57",
   "2023-08-29T13:53:19", "2023-08-29T13:48:23", "2023-08-29T13:17:42", "2023-08-29T12:53:17",
   "2023-08-29T06:15:00", "2023-08-29T11:07:01", "2023-08-28T13:23:17", "2023-08-27T15:50:36",
   "2023-08-26T15:47:56", "2023-08-26T13:09:13", "2023-08-21T12:28:48", "2023-08-16T17:33:10",
   "2023-08-12T06:25:07", "2023-08-08T17:29:40", "2023-08-05T13:41:05", "2023-08-01T05:56:35",
   "2023-07-31T19:35:24", "2023-07-30T06:39:49", "2023-07-20T11:35:10", "2023-07-19T17:24:04",
   "2023-07-14T09:02:04", "2023-07-07T09:55:04", "2023-07-07T09:35:13", "2023-07-05T16:44:39",
   "2023-07-05T16:42:05", "2023-08-27T20:13:19", "2023-08-25T19:35:33", "2023-08-21T08:29:36",
   "2023-08-15T19:39:36", "2023-08-07T20:21:36", "2023-07-06T22:01:22", "2023-04-12T13:33:31",
   "2023-01-01T13:32:50", "2023-08-29T07:55:27", "2023-08-29T05:41:04", ...]
```

I get:
```
#Explorer.Series<
  Polars[4563]
  datetime[ms] [nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil,
   nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil,
[...]

PR #795, opened 2024-01-02
Cast string to datetime

billylanchantin:
Fixes a bug (regression?) where casting from string to datetime always resulted in `nil`.

Closes:

* https://github.com/elixir-explorer/explorer/issues/729


lib/explorer/polars_backend/series.ex
```diff
@@ -38,12 +38,15 @@ defmodule Explorer.PolarsBackend.Series do
   def to_iovec(series), do: Shared.apply_series(series, :s_to_iovec)
 
   @impl true
+  def cast(%Series{dtype: :string} = series, {:datetime, precision}),
+    do: Shared.apply_series(series, :s_strptime, [nil, Atom.to_string(precision)])
+
```
billylanchantin, line 43: I wasn't sure if we wanted this logic here or in rust inside `s_cast`. I could see it going either way.

philss: Looks good to me here!

native/explorer/src/series.rs
```diff
@@ -1659,12 +1659,21 @@ pub fn s_second(s: ExSeries) -> Result<ExSeries, ExplorerError> {
 }
 
 #[rustler::nif(schedule = "DirtyCpu")]
-pub fn s_strptime(s: ExSeries, format_string: &str) -> Result<ExSeries, ExplorerError> {
+pub fn s_strptime(
+    s: ExSeries,
+    format_string: Option<&str>,
+    precision: Option<&str>,
```
philss, line 1665: We could use `ExTimeUnit` enum (that is declared inside the `src/datatypes/ex_dtypes.rs`) here - instead of `&str`. I think we could also implement the `TryFrom` trait for this conversion. WDYT?

lib/explorer/polars_backend/series.ex
philss: If we follow that path of using the `ExTimeUnit`, we don't need to convert the atom to string here :)

philss (commented):
I added a suggestion. Overall it looks good!

native/explorer/src/series.rs
billylanchantin: I gave it a shot!

[2953b52](https://github.com/elixir-explorer/explorer/pull/795/commits/2953b52748f81aef926251e0513387f9b0de3303)

I cargo-culted it. LMK if it wasn't right.

philss: Nice, it's correct! Thanks!

philss (approved):
:ship:

josevalim (approved):

PR #801, opened 2024-01-05
Bump Polars  to v0.36 - Join related changes

lkarthee:
continuation of Bump Polars v0.36 work

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -4636,7 +4636,7 @@ defmodule Explorer.DataFrame do
 
   # Two table verbs
 
-  @valid_join_types [:inner, :left, :right, :outer, :cross]
+  @valid_join_types [:inner, :left, :right, :outer, :outer_coalesce, :cross]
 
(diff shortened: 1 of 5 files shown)


lib/explorer/data_frame.ex
```diff
@@ -4636,7 +4636,7 @@ defmodule Explorer.DataFrame do
 
   # Two table verbs
 
-  @valid_join_types [:inner, :left, :right, :outer, :cross]
+  @valid_join_types [:inner, :left, :right, :outer, :outer_coalesce, :cross]
```
josevalim, line 4639: I wouldn't support outer_coalesce. It is very Polars specific and I am not sure there are other use cases for them besides backwards compatibility.

lkarthee: Ok, should we remove it ?

josevalim: Yes, I would not include outer_coalesce at all. :) We can remove it.

PR #805, opened 2024-01-09
implement - nil datatype - polar null type column

lkarthee:
For addressing issues relating to null datatype.

Its not ready for merging - need code review and some help.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -919,16 +919,20 @@ defmodule Explorer.Series do
   """
   @doc type: :element_wise
   @spec cast(series :: Series.t(), dtype :: dtype()) :: Series.t()
-  def cast(%Series{dtype: original_dtype} = series, dtype) do
-    if normalised = Shared.normalise_dtype(dtype) do
-      if normalised == original_dtype do
-        series
-      else
-        apply_series(series, :cast, [normalised])
-      end
-    else
-      dtype_error("cast/2", dtype, Shared.dtypes())
-    end
+  def cast(series, dtype) do
+    dtype |> Shared.normalise_dtype() |> maybe_cast_series(series)
+  end
+
+  defp maybe_cast_series({:error, dtype}, _), do: dtype_error("cast/2", dtype, Shared.dtypes())
+
+  defp maybe_cast_series(dtype, %Series{dtype: dtype} = series), do: series
+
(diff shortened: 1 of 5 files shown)


lkarthee:
@josevalim There are a lot of test failures - i will fix them after review ..

lib/explorer/series.ex
```diff
@@ -919,16 +919,20 @@ defmodule Explorer.Series do
...
+  defp maybe_cast_series({:error, dtype}, _), do: dtype_error("cast/2", dtype, Shared.dtypes())
+
+  defp maybe_cast_series(dtype, %Series{dtype: dtype} = series), do: series
+
+  defp maybe_cast_series(normalised, series), do: apply_series(series, :cast, [normalised])
+
+  @doc type: :element_wise
+  @spec nil_(series :: Series.t()) :: Series.t()
+  def nil_(series) do
+    cast(series, nil)
```
josevalim, line 935: I think I would call it `:null`, so we don't accidentally handle it as `nil`. And i don't think we need a `nil_` function, `cast(series, :null)` would be enough IMO.

josevalim: Also calling it `:null` would require fewer changes to the code, exactly becaude of stuff like `normalise_dtype` :D

lkarthee: Thank you, fixed as :null - that resolved all failing test cases.

josevalim:
Thank you!

Let's add some tests that `to_list` works and returns a list of nils. I assume `to_binary` and `to_iovec` both fail? Do they fail with a decent error message at least?

lkarthee:
Sure I will add test cases for `to_list`. 

Error message for `to_iovec` and `to_binary`.

```
** (ArgumentError) cannot convert series of dtype :null into iovec
    (explorer 0.8.0-dev) lib/explorer/series.ex:769: Explorer.Series.to_iovec/1
    #cell:2wexfollu25fyu43:14: (file)
```

For [1, 2, 3] `to_iovec` fails with : 
```
** (RuntimeError) Generic Error: cannot invoke to_iovec on series with nils
```

How to fix iovec error any pointers ?

josevalim:
Those are all expected and good. So we only need tests for to_list and we are good to go. :)

josevalim:

Issue #806, opened 2024-01-09
Support `Series.from_list([1, 2, 3], dtype: :null)`


josevalim:
We should also discuss if the default for a list of nils should still be `:f64` or if we should swap it to `:null`:

    Series.from_list([nil, nil, nil])

josevalim:
cc @lkarthee in case you want to take a look at this one too :)

lkarthee:
Was just testing this exact thing out in livebook. 👍

philss:
FYI, we may need to use `Series::new_null` in the crate side, to create a null series:

```rust
use polars::prelude::*;

fn main() {
    let s = Series::new_null("a", 0);

    println!("dtype is: {}", &s.dtype())
}
```

lkarthee:
I fixed `from_list`, then realised `DF.concat_rows` and `Series.concat` are broken. 

null -> any conversion needs to be added. 

I have noticed that: 
 - type conversion between distinct numeric series converts to floats. 
 - does this logic assumes there are only 2 numeric types - integer and float.  
 - This logic needs to be revisited after introduction of multiple numeric types ?

philss:
> existing logic assumes that there are only 2 numeric types - integer and float ?

Yes, but I'm changing this in #794 

> This logic needs to be revisited after introduction of multiple numeric types ?

Probably it's easier to revisit after I finish that work, yeah.

lkarthee:
One question about type casting of nulls when concat rows. In Polars, concat works as:
 - DataFrame has different conditions and types of concat apart from type.
 - Type ++ Null can be appended.
 - Null ++ Type fails with Polars Error.


Should we support both cases without throwing error by handling casting if needed or should we only support Type ++ Null ? 

I am trying to fix Series.concat for Type ++ Null as it is failing.

josevalim:
There is a chance that Null + Types not working is a Polars bug, so I would take a look and perhaps do a report there. If that’s the case, I would write your code as if both styles are support than let Polars blow up (until it is fixed).

lkarthee:
My line of thought is Type can have value nil, while concatenation why can't it be treated a set of rows having nil values being added ? Null ++ Type failure does not make sense for me particularly if a cast can solve it. I might be missing something. Please share your thoughts on this.

billylanchantin:
> write your code as if both styles are support and then let Polars blow up (until it is fixed).

[...]

PR #807, opened 2024-01-10
Implement Series.from_list of  null type 

lkarthee:
Fix for https://github.com/elixir-explorer/explorer/issues/806

Support create `Series.from_list([nil, nil])`

Even though tests pass - `DataFrame.concat_rows` may fail and `Series.concat` will fail if one of dtypes is :null (irrespective of this fix) - need to address it later.

```elixir
s2 = Series.from_list([4.0, 5.0, 6.4])
Series.concat([s2, sn]) # will fail - it is valid in polars
```


lkarthee:
ignore this - https://github.com/elixir-explorer/explorer/pull/808

PR #808, opened 2024-01-10
Implement Series.from_list of null type

lkarthee:
Fix for https://github.com/elixir-explorer/explorer/issues/806

Support create `Series.from_list([nil, nil])`

Even though tests pass - `DataFrame.concat_rows` may fail and `Series.concat` will fail if one of dtypes is :null (irrespective of this fix) - need to address it later. Series concat nulls has no test cases

```elixir
s2 = Series.from_list([4.0, 5.0, 6.4])
Series.concat([s2, sn]) # will fail - it is valid in polars
```

### DIFF
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -260,18 +260,17 @@ defmodule Explorer.Shared do
   """
   def dtype_from_list!(list, preferable_type \\ nil) do
     initial_type =
-      if leaf_dtype(preferable_type) in ([
-                                           :numeric,
-                                           :binary,
-                                           {:f, 32},
-                                           {:f, 64},
-                                           :category
-                                         ] ++ @integer_types),
+      if leaf_dtype(preferable_type) in ([:null, :numeric, :binary, {:f, 32}, {:f, 64}, :category] ++
+                                           @integer_types),
          do: preferable_type
+
+    # require Logger
 
     type =
       Enum.reduce(list, initial_type, fn el, type ->
         new_type = type(el, type) || type
+        # require Logger
+        # Logger.info("new_type - #{inspect(new_type)}")
 
         if new_type_matches?(type, new_type) do
           new_type
@@ -281,6 +280,8 @@ defmodule Explorer.Shared do
         end
       end)
 
+    # Logger.info("type - #{inspect type}")
+    # || preferable_type || {:f, 64}
     type || preferable_type || {:f, 64}
   end
 
@@ -312,6 +313,8 @@ defmodule Explorer.Shared do
   defp type(item, :category) when is_binary(item), do: :category
   defp type(item, _type) when is_binary(item), do: :string
 
+  defp type(nil, nil), do: :null
+  defp type(nil, :null), do: :null
   defp type(item, _type) when is_nil(item), do: nil
   defp type([], _type), do: nil
   defp type([_item | _] = items, type), do: {:list, result_list_type(items, type)}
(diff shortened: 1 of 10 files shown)


lkarthee:
Switched {:f, 64} to :null when a list of nils is passed to Series.from_list/1

lib/explorer/shared.ex
```diff
@@ -281,6 +280,8 @@ defmodule Explorer.Shared do
         end
       end)
 
+    # Logger.info("type - #{inspect type}")
+    # || preferable_type || {:f, 64}
     type || preferable_type || {:f, 64}
```
josevalim, line 285: I think we can instead change this line to:

```suggestion
    type || preferable_type || :null
```

And then we can most likely skip other changes to this file?

test/explorer/series/struct_test.exs
```diff
@@ -4,6 +4,23 @@ defmodule Explorer.Series.StructTest do
   alias Explorer.Series
 
   describe "from_list/2" do
+    test "allows struct of all nil value" do
```
josevalim, line 7: What about a list of lists of nils?

    Series.from_list([[nil, nil], [nil, nil, nil]])

test/explorer/series_test.exs
```diff
@@ -30,6 +30,20 @@ defmodule Explorer.SeriesTest do
   end
 
   describe "from_list/1" do
+    test "with nils" do
+      s = Series.from_list([nil, nil, nil])
+
+      assert Series.to_list(s) === [nil, nil, nil]
+      assert Series.dtype(s) == :null
+    end
```
josevalim, line 38: What about a list of lists of nils?

    Series.from_list([[nil, nil], [nil, nil, nil]])

lib/explorer/shared.ex
josevalim: To be precise, skip the changes to `type/2` and `new_type_matches?`. :)

test/explorer/series_test.exs
lkarthee: test/explorer/series/list_test.exs has below test case: 

```
test "list of list of nulls" do
       series = Series.from_list([[nil, nil], [nil]])
       assert series.dtype == {:list, :null}
       assert series[0] == [nil, nil]
       assert Series.to_list(series) == [[nil, nil], [nil]]
     end
```

lib/explorer/shared.ex
lkarthee: 3 existing test cases fail :
```elixir
s = Series.from_list([])
p = Series.cumulative_product(s)
**  (ArgumentError) Explorer.Series.cumulative_product/2 not implemented for dtype :null.
p = Explorer.Series.product(s)
**  (ArgumentError) Explorer.Series.product/1 not implemented for dtype :null
p = Series.less(s, nil)
**  cannot invoke Explorer.Series.less/2 with mismatched dtypes: {:f, 64} and nil

```
2 new test cases fail:
```elixir
s = Series.from_list([1, 2, 3], dtype: :null)
** (ArgumentError) the value 1 does not match the inferred series dtype :null
s =
     Series.from_list([
       %{a: nil, b: nil},
       %{a: 3, b: nil},
       %{a: 5, b: nil}
     ])
** (CaseClauseError) no case clause matching: nil
```

There is assumption that empty array is `{:f, 64}`

josevalim:

I will investigate the failures then, thank you!

PR #810, opened 2024-01-10
Add :null to scalar types

lkarthee:
To fix issue of `Shared.dtypes/0` not returning `:null` type


lkarthee:
This has been addressed by Jose in another merge.

Issue #794, opened 2024-01-02
Support operations using new signed and unsigned integers dtypes


philss:
The idea is to make sure we can support the most number of operations with these new dtypes,
including mixing them when it makes sense.

### Aggregation
- [x] eager `Explore.Series.min/1` and `Explore.Series.max/1` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explore.Series.mean/1` and `Explore.Series.median/1` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explore.Series.product/1` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explore.Series.skew/2` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explore.Series.standard_deviation/2` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explore.Series.sum/1` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explore.Series.variance/1` - https://github.com/elixir-explorer/explorer/pull/824

### Element-wise
- [x] eager `Explorer.Series.add/2` mixing signed and unsigned series - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explorer.Series.categorise/2` with indexes of `{:u, size}` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explorer.Series.divide/2` with both sides of `{:u, size}`, or mixing `{:u, size}` and `{:s, size}` - https://github.com/elixir-explorer/explorer/pull/824
- [x] lazy `Explorer.Series.remainder/2` with left or right of `{:u, size}` (raised inside `Series.filter_with/2`) - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explorer.Series.in/2` - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explorer.Series.subtract/2` (using the example of `map_with`) - https://github.com/elixir-explorer/explorer/pull/824
- [x]  lazy `Explorer.Series.subtract/2` (using the example of `map_with`) - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explorer.Series.multiply/2` with `{:u, size}` and mixing int types - https://github.com/elixir-explorer/explorer/pull/824
- [x] eager `Explorer.Series.peaks/2` - https://github.com/elixir-explorer/explorer/pull/824
[...]

PR #812, opened 2024-01-11
Series.concat handle multiple integer, float and null types

lkarthee:
This is an attempt handle multiple {:s, _}, {:u, _}, {:f, _} and new null type. Two issues are being addressed: 

 -  type casting / promotion
 -  concatenation involving a  :null series

### Concatenation of null series
- Type ++ Null  - allowed 
- Null ++ Type  - allowed
- Null ++ Null  - allowed

### Type casting
Type casting is allowed only when there is no loss of information.

- When same numeric types are involved, it is casted to largest type among the series. For example, {:s, 8}, {:s, 16} -> {:s, 8} will be casted to {:s, 16} . In case of {:u, 16}, {:u, 32}, {:u, 16} will be casted to {:u, 32}
- When different numeric types are involved, it is casted to {:f, 64}. I have not changed the existing behaviour. Will update this behaviour later.
- numeric and other types - error

As the discussion of type casting is going on, I have not written tests.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -2092,20 +2092,47 @@ defmodule Explorer.Series do
   @doc type: :shape
   @spec concat([Series.t()]) :: Series.t()
   def concat([%Series{} | _t] = series) do
-    dtypes = series |> Enum.map(& &1.dtype) |> Enum.uniq()
+    {null?, dtypes_map} = 
+      Enum.reduce(series, {false, %{}}, fn 
+        %Series{dtype: :null}, {true, dtypes} = acc -> acc
+        %Series{dtype: :null}, {_, dtypes} = acc -> {true, dtypes}
+        %Series{dtype: dt}, {null?, dtypes} when is_atom(dt) -> {null?, Map.put_new(dtypes, dt, dt)}
+        %Series{dtype: {type, n} = dt}, {null?, dtypes} -> 
+          dtypes = 
+            case dtypes[type] do
+              nil -> Map.put(dtypes, type, dt)
+              {_t, tn} when n > tn -> Map.put(dtypes, type, dt)
+              _ -> dtypes
+            end
+          {null?, dtypes}
+      end)
+    
+    dtypes = Map.values(dtypes_map)
 
-    case dtypes do
-      [_dtype] ->
-        impl!(series).concat(series)
+    series = 
+      series
+      |> maybe_raise_mismatched!(dtypes)
+      |> maybe_cast(null?, dtypes)
 
-      [a, b] when K.and(is_numeric_dtype(a), is_numeric_dtype(b)) ->
-        series = Enum.map(series, &cast(&1, {:f, 64}))
-        impl!(series).concat(series)
+    impl!(series).concat(series)
+  end
 
-      incompatible ->
-        raise ArgumentError,
-              "cannot concatenate series with mismatched dtypes: #{inspect(incompatible)}. " <>
-                "First cast the series to the desired dtype."
+  defp maybe_cast(series, true, []), do: series
+
+  defp maybe_cast(series, _, [dtype]), do: Enum.map(series, &cast(&1, dtype))
+
+  defp maybe_cast(series, _, _), do: Enum.map(series, &cast(&1, {:f, 64}))
+
+  defp maybe_raise_mismatched!(series, [dtype]), do: series
+
+  defp maybe_raise_mismatched!(series, dtypes) do
+    if Enum.all?(dtypes, &is_numeric_dtype/1) do
+      series
+    else
+      dtypes = Enum.map(series, &(&1.dtype))
+      raise ArgumentError,
+            "cannot concatenate series with mismatched dtypes: #{inspect(dtypes)}. " <>
+              "First cast the series to the desired dtype."
     end
   end
 


lkarthee:
```elixir
ss = Series.from_list(["a", "b", "c"])
s8 = Series.from_list([1, 2, 3], dtype: {:s, 8})
s16 = Series.from_list([16, 17, 18], dtype: {:s, 16})
f32 = Series.from_list([32.0, 33.0, 34.0], dtype: {:f, 32})
f64 = Series.from_list([64.0, 65.0, 6.4], dtype: {:f, 64})
u8 = Series.from_list([10, 11, 12], dtype: {:u, 8})
Series.concat([s8, sn])
Series.concat([ss, sn]) 
# Series.concat([ss, s8]) # fails 
# Series.concat([ss, f32]) # fails 
Series.concat([sn, f32])
Series.concat([f32, sn])
Series.concat([sn, s8, s16])
Series.concat([sn, s8, f32])
Series.concat([u8, s8])  # cast to {:f, 64}
```

josevalim:
Polars logic is that it must preserve the type on the left side, rather than trying to find a magic type that satisfies both sides, so it performs fewer casting. I'd say that's a sane we can choose to mimic if we want to. Overall, we have to choose between:

1. Define our own casting logic that aims to unify both sides (this PR)
2. Align with Polars and document that the type of the left side is preserved

Any opinions?

lkarthee:
Null ++ Type and Type ++ Null can be done. We can cast to a same higher type if there is no loss of information. For example, `{:s, _}` to `{:u, _}` casting should be explicit.

cigrainger:
I'm definitely tempted to cast in such a way that we preserve information. That being said, it can end up being surprising, even when well documented. There's something to be said for the simplicity of 'always left side' vs. 'here's a hierarchy'. I'm not entirely sure. I think the hierarchy/preservation approach can become confusing if, say, you're reducing over a number of concatenations. But the opposite is also true: let's say you have a number of Series and one of them has default to `null` because it was empty and you didn't know (this happens plenty for me). I would find it surprising and frustrating if I was getting a `null` type at the end of some function and I couldn't figure out why.

All of this is to say I think there are pros and cons both ways and I'm just not sure. How complex do we think juggling a hierarchy will be going forward?

billylanchantin:
Whoops I didn't mean to summon you, @alexander-beedie! But thank you for the extra context! And of course all the work on Polars, which we get to benefit from :)

philss:
I was following the approach of casting to preserve information for when the user is mixing different numeric dtypes in https://github.com/elixir-explorer/explorer/issues/794.
I notice that Polars (Rust) does some casting - for example, the subtraction `u8 - f64` of series results in a `f64` series -, but sometimes I have to do it myself - in the case of `u8 - u8`, I have to cast the left side to signed int in order to perform the operation.

> However, my pragmatic side says that following Polars will be easier in the long term. Left-side-wins is a simple take that's easy to apply.

I agree with @billylanchantin here! I think in the long term, it will be probably easier to maintain. But I have my doubts if this is going to lead to a pleasant experience for the users. My take - and it's not a strong opinion - is that we would want to keep casting for binary operations (like add or subtract), but we would keep APIs like `concat/1` more restrictive, requiring the user to explicitly cast the series to a single dtype.

But yeah, I agree with @cigrainger that this can be confusing even if we document it - and probably more confusing mixing up the behaviours like I'm suggesting here :sweat_smile:. But it may be a nicer experience IMHO.

lkarthee:
### Summarizing discussion
Here are the main themes:
 - Are operations commutative ?  a ++ b = b ++ a 
 - Surprise element or unintuitive to user.
 - Preserve information.
 - Complexity of rules.
 - Maintainability w.r.to Polars alignment.
 - Precedent for diverging from polars.
 - Effort required to implement and maintain say commutative for null & type casting.

My opinion:
 - For a user coming from elixir, it might surprise that b  ++ a is not allowed ?
 - Say if nil is valid in a column, but when appending order matters for a list of nils is unintuitive.
 - Casting in this case does not result in any loss of info.
 - Same type casting to a higher rank should be allowed for convenience.
 - Can be restrictive w.r.to casting to a different numeric type {:s, n} to {:u, n} or {:s, n} to {:f, n} or {:u, n} to {:f, n} and vice versa.
 - There is a precedent of diverging from polars date time and duration.
 - Can there be a flag to disallow/opt out of the divergent behaviour ?

cigrainger:
Good feedback from a trusted friend who is a heavy R/dplyr user: "math not being commutative would break my f***ing brain".

billylanchantin:
@cigrainger I agree with your friend! Although `++` isn't commutative, it _is_ commutative WRT the output type (or ought to be IMHO).

@lkarthee Excellent summary.

> Can there be a flag to disallow/opt out of the divergent behaviour ?

I vote no because 1) it would be a maintenance burden and 2) you can use explicit casting to get a different behavior if you want.

billylanchantin:
@philss I just re-read your comment. To clarify, **I think we could diverge from Polars and cast**. I was trying to outline the pitfalls of that decision in my comment, but I can see how it came off as me arguing for keeping in line with Polars. Sorry if I muddled the convo!

josevalim:
It seems everyone is in favor of this change, so let's go ahead! Only tests + docs missing. :) Thank you @lkarthee

lkarthee:
Need clarity on when types don't match - fail or cast ?
 - int to float 
 - unsigned to signed  

```elixir
Series.concat([u8, s8]) # should fail ?
Series.concat([u8, f32]) # ??? fail or cast ?
```

josevalim:
U8 with S8 becomes S16. Anything with float becomes float. + probably already works like that.

billylanchantin:
@josevalim For my own edification, is the logic there that `{:u, x} == {:s, 2x}`?

josevalim:
@billylanchantin correct!

josevalim:

PR #816, opened 2024-01-13
add nil expr for summarise and mutate - Py Polars `lit(None)`

lkarthee:
This PR add `lit(None)`.  Can be used to align `describe` with Py Polars. 

```elixir
df = DF.new(strs: ["a", "b", "c"], nums: [1, 2, 3])
DF.mutate(df, c: nil)
#Explorer.DataFrame<
  Polars[3 x 3]
  strs string ["a", "b", "c"]
  nums s64 [1, 2, 3]
  c null [nil, nil, nil]
>
DF.summarise(df, c: nil)
#Explorer.DataFrame<
  Polars[1 x 1]
  c null [nil]
>
```

I will add test cases after review.

### DIFF
--- a/test/explorer/data_frame_test.exs
+++ b/test/explorer/data_frame_test.exs
@@ -663,6 +663,16 @@ defmodule Explorer.DataFrameTest do
   end
 
   describe "mutate/2" do
+    test "with nil" do
+      df = DF.new(strs: ["a", "b", "c"], nums: [1, 2, 3])
+      assert DF.mutate(df, c: nil) |> inspect() == ~s(#Explorer.DataFrame<
+  Polars[3 x 3]
+  strs string ["a", "b", "c"]
+  nums s64 [1, 2, 3]
+  c null [nil, nil, nil]
+>)
+    end
+
     test "adds new columns" do
       df = DF.new(a: [1, 2, 3], b: ["a", "b", "c"])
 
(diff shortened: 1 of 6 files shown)


josevalim (approved):

test/explorer/data_frame_test.exs
```diff
@@ -663,6 +663,16 @@ defmodule Explorer.DataFrameTest do
...
 
   describe "mutate/2" do
+    test "with nil" do
+      df = DF.new(strs: ["a", "b", "c"], nums: [1, 2, 3])
+      assert DF.mutate(df, c: nil) |> inspect() == ~s(#Explorer.DataFrame<
+  Polars[3 x 3]
+  strs string ["a", "b", "c"]
+  nums s64 [1, 2, 3]
+  c null [nil, nil, nil]
+>)
```
josevalim, line 673: We should not rely on inspect for tests. We should use something like to_columns or assert on individual series. The tests below are a good example!

lkarthee: what is preferred in `DF.dtypes(df1)` or `df1.dtypes`

josevalim: Any is fine. See the one we use the most in test and follow that template :)

josevalim:

test/explorer/data_frame_test.exs
lkarthee: i have rewritten tests and pushed..

PR #817, opened 2024-01-13
typespec - add :null to dtype

lkarthee:


josevalim:

PR #820, opened 2024-01-14
add DataFrame.transpose/2

lkarthee:
```elixir
df = Explorer.DataFrame.new(a: [32.0, 33.0], b: [1, 2], c: ["a", "b"])
DF.print(df)
dft = DF.transpose(df, include_header: true, header_name: "names", column_names: ["x", "y"]) #, 

+--------------------------------------------+
| Explorer DataFrame: [rows: 3, columns: 3]  |
+--------------+--------------+--------------+
|    names     |      x       |      y       |
|   <string>   |   <string>   |   <string>   |
+==============+==============+==============+
| a            | 32.0         | 33.0         |
+--------------+--------------+--------------+
| b            | 1            | 2            |
+--------------+--------------+--------------+
| c            | a            | b            |
+--------------+--------------+--------------+


```

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -4206,6 +4206,77 @@ defmodule Explorer.DataFrame do
   end
 
   @doc """
+  Transpose a DataFrame.
+
+  > #### Warning {: .warning}
+  >
+  > This is an expensive operation since data is stored in a columnar format.
+
+  ## Options
+
+    * `:include_header` - if `true` column names are added as first column or header column. (default: `false`)
+    * `:header_name` - when `include_header` is `true`, sets name of the header column. (default: `column`)
+    * `:column_names` - names for non header columns. Length of column_names should match the row count of data frame. (default: `nil`)
+
+  ## Examples
+
+      iex> df = Explorer.DataFrame.new(a: ["d", nil], b: [1, 2], c: ["a", "b"])
+      iex> Explorer.DataFrame.transpose(df)
+      #Explorer.DataFrame<
+        Polars[3 x 2]
+        column_0 string ["d", "1", "a"]
+        column_1 string [nil, "2", "b"]
+      >
+
+      iex> df = Explorer.DataFrame.new(a: ["d", nil], b: [1, 2], c: ["a", "b"])
+      iex> Explorer.DataFrame.transpose(df, column_names: ["x", "y"])
+      #Explorer.DataFrame<
+        Polars[3 x 2]
+        x string ["d", "1", "a"]
+        y string [nil, "2", "b"]
+      >
+
+      iex> df = Explorer.DataFrame.new(a: ["d", nil], b: [1, 2], c: ["a", "b"])
+      iex> Explorer.DataFrame.transpose(df, include_header: true, header_name: "name")
+      #Explorer.DataFrame<
+        Polars[3 x 3]
+        name string ["a", "b", "c"]
+        column_0 string ["d", "1", "a"]
+        column_1 string [nil, "2", "b"]
+      >
+  """
+  @doc type: :single
+  @spec transpose(df :: DataFrame.t(), opts :: Keyword.t()) :: DataFrame.t()
+  def transpose(df, opts \\ []) do
+    opts =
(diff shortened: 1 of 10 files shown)


lib/explorer/data_frame.ex
```diff
@@ -4205,6 +4205,77 @@ defmodule Explorer.DataFrame do
...
+  Transpose a DataFrame.
+
+  > #### Warning {: .warning}
+  >
+  > This is an expensive operation since data is stored in a columnar format.
+
+  ## Options
+
+    * `:include_header` - if `true` column names are added as first column or header column. (default: `false`)
+    * `:header_name` - when `include_header` is `true`, sets name of the header column. (default: `column`)
```
josevalim, line 4218: We don't need both options. We can have a `:header`option. It may be false (default), true (default to column) or a string with the header name.

lkarthee: should `:column_names` be `columns` ?  I see it as a common parameter for many DF functions ?

lkarthee: i have changed `:column_names` to `column`. I can revert it..

lib/explorer/data_frame.ex
```diff
@@ -4205,6 +4205,82 @@ defmodule Explorer.DataFrame do
...
+  @doc type: :single
+  @spec transpose(df :: DataFrame.t(), opts :: Keyword.t()) :: DataFrame.t()
+  def transpose(df, opts \\ []) do
+    opts =
+      Keyword.validate!(opts, header: false, columns: nil)
+
+    header =
+      case opts[:header] do
+        false -> false
+        nil -> nil
```
josevalim, line 4256:
```suggestion
```

lkarthee: To avoid `to_column_name` convert it to string, I added that clause -> `nil -> nil` ?

josevalim: When it would ever be nil? The accepted values should be Boolean or string, no?

lkarthee: Atoms are also accepted- they are converted to strings by to_column(). So when user passes nil, it will be converted to "nil". Just to guard it I put that condition. Is nil to "nil" conversion is acceptable?

billylanchantin: Handling atoms makes sense. Would an `ArgumentError` be more appropriate for `header: nil`?

Also I think guards on the `to_column_name` clause might be helpful.

lkarthee: If we disallow `nil`, user may pass `"nil"`.  `df[nil]` is valid, even though it looks funny. 
```elixir
iex> dft = DF.transpose(df, header: "nil", columns: ["x", "y"])
iex> dft[nil]
#Explorer.Series<
  Polars[3]
  string ["a", "b", "c"]
>
```
Not much we can do  here. I have accepted Jose change code suggestion and removed the statement.

test/explorer/data_frame_test.exs
```diff
@@ -2722,6 +2722,82 @@ defmodule Explorer.DataFrameTest do
...
+               "y" => ["33.0", "2", "b"]
+             }
+    end
+
+    test "with column names < length of rows raises" do
+      df = Explorer.DataFrame.new(a: [32.0, 33.0], b: [1, 2], c: ["a", "b"])
+      assert DF.shape(df) == {2, 3}
+
+      assert_raise RuntimeError,
+                   "Polars Error: lengths don't match: Length of new column names must be the same as the row count",
```
josevalim, line 2791: Ideally we want to validate what we can on Explorer side. We only let Polars raise when the error requires traversing the series itself. Can you validate this one in Elixir too? Thank you!

josevalim (approved):
Looks great to me! @cigrainger should decide on the API (name and options). :)

cigrainger (approved):
Yeah this can be tough to do with `pivot_wider`. It's a bit niche, but all good. I'm cool with `transpose`, and I think you guys got to the ideal set of opts w/ `header` and `columns`. I'm happy with this!

test/explorer/data_frame_test.exs
lkarthee: addressed in #823

PR #821, opened 2024-01-14
Fix mismatched types in `Series.pow`

billylanchantin:
Fixes:

* https://github.com/elixir-explorer/explorer/issues/374

New dtype handling as of Rust Polars 0.36.2:

```
pow(uint,  uint)  = uint
pow(uint,  sint)  = float
pow(sint,  uint)  = sint
pow(sint,  sint)  = float
pow(float, *int)  = float
pow(*int,  float) = float
pow(float, float) = float
```

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -3772,8 +3772,6 @@ defmodule Explorer.Series do
   def atan(%Series{dtype: dtype}),
     do: dtype_error("atan/1", dtype, [{:f, 32}, {:f, 64}])
 
-  defp basic_numeric_operation(operation, left, right, args \\ [])
-
   defp basic_numeric_operation(operation, %Series{} = left, right, args) when is_numeric(right),
     do: basic_numeric_operation(operation, left, from_same_value(left, right), args)
 
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -3402,7 +3402,7 @@ defmodule Explorer.Series do
       iex> Explorer.Series.pow(s, 3)
       #Explorer.Series<
         Polars[3]
-        s64 [8, 64, 216]
+        f64 [8.0, 64.0, 216.0]
       >
 
       iex> s = [2, 4, 6] |> Explorer.Series.from_list()
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -3428,19 +3428,7 @@ defmodule Explorer.Series do
   """
   @doc type: :element_wise
   @spec pow(left :: Series.t() | number(), right :: Series.t() | number()) :: Series.t()
-  def pow(left, right) do
-    if K.or(match?(%Series{}, left), match?(%Series{}, right)) do
-      apply_series_list(:pow, [cast_for_pow(left), cast_for_pow(right)])
-    else
-      raise ArgumentError, "at least one input must be a series"
-    end
-  end
-
-  @non_finite [:nan, :infinity, :neg_infinity]
-  defp cast_for_pow(%Series{} = series), do: series
-  defp cast_for_pow(u) when K.and(is_integer(u), u >= 0), do: from_list([u], dtype: {:u, 64})
-  defp cast_for_pow(s) when K.and(is_integer(s), s < 0), do: from_list([s], dtype: {:s, 64})
-  defp cast_for_pow(f) when K.or(is_float(f), K.in(f, @non_finite)), do: from_list([f])
+  def pow(left, right), do: basic_numeric_operation(:pow, left, right)
 
   @doc """
   Calculates the natural logarithm.
@@ -3763,6 +3751,8 @@ defmodule Explorer.Series do
 
   def atan(%Series{dtype: dtype}),
     do: dtype_error("atan/1", dtype, [{:f, 32}, {:f, 64}])
+
+  defp basic_numeric_operation(operation, left, right, args \\ [])
 
   defp basic_numeric_operation(operation, %Series{} = left, right, args) when is_numeric(right),
     do: basic_numeric_operation(operation, left, from_same_value(left, right), args)
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -3428,7 +3428,27 @@ defmodule Explorer.Series do
   """
   @doc type: :element_wise
   @spec pow(left :: Series.t() | number(), right :: Series.t() | number()) :: Series.t()
-  def pow(left, right), do: basic_numeric_operation(:pow, left, right)
+  def pow(left, right) do
+    [left, right] = cast_for_arithmetic("pow/2", [left, right])
+
+    if out_dtype = cast_to_pow(dtype(left), dtype(right)) do
+      apply_series_list(:pow, [out_dtype, left, right])
+    else
+      dtype_mismatch_error("pow/2", left, right)
+    end
+  end
+
+  # TODO: ensure these types are correct.
+  defp cast_to_pow({:u, l}, {:u, r}), do: {:u, max(l, r)}
+  defp cast_to_pow({:u, _}, {:s, _}), do: {:f, 64}
+  defp cast_to_pow({:u, _}, {:f, f}), do: {:f, f}
+  defp cast_to_pow({:s, s}, {:u, u}), do: {:s, min(64, max(2 * u, s))}
+  defp cast_to_pow({:s, _}, {:s, _}), do: {:f, 64}
+  defp cast_to_pow({:s, _}, {:f, f}), do: {:f, f}
+  defp cast_to_pow({:f, f}, {:u, _}), do: {:f, f}
+  defp cast_to_pow({:f, f}, {:s, _}), do: {:f, f}
+  defp cast_to_pow({:f, f}, {:f, _}), do: {:f, f}
+  defp cast_to_pow(_, _), do: nil
 
   @doc """
   Calculates the natural logarithm.
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -3766,8 +3766,6 @@ defmodule Explorer.Series do
   def atan(%Series{dtype: dtype}),
     do: dtype_error("atan/1", dtype, [{:f, 32}, {:f, 64}])
 
-  defp basic_numeric_operation(operation, left, right, args \\ [])
-
   defp basic_numeric_operation(operation, %Series{} = left, right, args) when is_numeric(right),
     do: basic_numeric_operation(operation, left, from_same_value(left, right), args)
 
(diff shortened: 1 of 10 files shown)


native/explorer/src/series.rs
```diff
@@ -1231,65 +1231,55 @@ pub fn s_n_distinct(s: ExSeries) -> Result<usize, ExplorerError> {
...
-                iter1
-                    .zip(iter2)
-                    .map(|(v1, v2)| v1.and_then(|left| v2.map(|right| left.powf(right))))
-                    .collect()
-            };
+pub fn s_pow(
+    base_exseries: ExSeries,
+    base_exdtype: ExSeriesDtype,
+    exponent_exseries: ExSeries,
+    exponent_exdtype: ExSeriesDtype,
```
billylanchantin, line 1238: Is there a way to pull the dtype off an `ExSeries`? I couldn't figure it out, so I just passed them in as extra args. 😅

test/explorer/series_test.exs
```diff
@@ -2091,7 +2154,7 @@ defmodule Explorer.SeriesTest do
 
       result = Series.pow(s1, s2)
 
-      assert result.dtype == {:s, 64}
+      assert result.dtype == {:f, 64}
```
billylanchantin, line 2157: Since we infer `Series.from_list([1, 2, 3])` to have type `{:s, 64}`, several result dtypes had to switch to float.

lib/explorer/series.ex
```diff
@@ -3428,7 +3428,19 @@ defmodule Explorer.Series do
...
+    if K.or(match?(%Series{}, left), match?(%Series{}, right)) do
+      apply_series_list(:pow, [cast_for_pow(left), cast_for_pow(right)])
+    else
+      raise ArgumentError, "at least one input must be a series"
+    end
+  end
+
+  @non_finite [:nan, :infinity, :neg_infinity]
+  defp cast_for_pow(%Series{} = series), do: series
+  defp cast_for_pow(u) when K.and(is_integer(u), u >= 0), do: from_list([u], dtype: {:u, 64})
```
billylanchantin, line 3441: I added the `cast_for_pow` function entirely for this clause. It allows you to get:

```elixir
iex> s1 = Explorer.Series.from_list([1, 2, 3])
iex> Explorer.Series.pow(s1, 2)
# #Explorer.Series<
#   Polars[3]
#   s64 [1, 4, 9]
# >
```

I.e. raising a signed int series to an unsigned int literal keeps the result dtype as signed int instead of casting to a float.

Per out conversation on slack about unsigned ints, I'm wondering if this actually desirable. It's at odds with:

```elixir
iex> s1 = Explorer.Series.from_list([1, 2, 3])
iex> s2 = Explorer.Series.from_list([2])
iex> Explorer.Series.pow(s1, s2)
# #Explorer.Series<
#   Polars[3]
#   f64 [1.0, 4.0, 9.0]
# >
```

If we didn't have that clause, though, you'd have to do this to keep the dtype as signed int:

```elixir
iex> s1 = Explorer.Series.from_list([1, 2, 3])
iex> s3 = Explorer.Series.from_list([2], dtype: :u64)
iex> Explorer.Series.pow(s1, s3)
# #Explorer.Series<
#   Polars[3]
#   s64 [1, 4, 9]
# >
```

which is less ergonomic than just `Explorer.Series.pow(s1, 2)`.

Thoughts?

native/explorer/src/series.rs
lkarthee: > Is there a way to pull the dtype off an ExSeries?
`s.dtype()` or `e.dtype()`

https://github.com/elixir-explorer/explorer/blob/f06700ccb98f2a55f3a64b114d5be1b673185f33/native/explorer/src/datatypes.rs#L28-L41

https://docs.rs/polars/latest/polars/series/trait.SeriesTrait.html#method.dtype

lkarthee, line 1238:
```suggestion
pub fn s_pow(s: ExSeries, other: ExSeries) -> Result<ExSeries, ExplorerError> {
```

lib/explorer/series.ex
josevalim: Could we, by any chance, have Nx behaviour? If integers are given, it raises if the right side is negative, and always returns integers. If you want a float response, you just cast any of the results to float. I guess this could be achieved by casting the right side to `{:u, precision}` always when a `{:s, precision}` is given?

native/explorer/src/series.rs
billylanchantin: Thank you!

* [66b45ce](https://github.com/elixir-explorer/explorer/pull/821/commits/66b45ce4228252efdd02a55fba775816ed261ccd)
* [ed5338f](https://github.com/elixir-explorer/explorer/pull/821/commits/ed5338f6085507e7e4780dd31026a19dedb27f03)

lib/explorer/series.ex
billylanchantin: > Could we, by any chance, have Nx behaviour? If integers are given, it raises if the right side is negative, and always returns integers.

I think this would break with Polars since they allow `pow(uint, sint) == float`.

My reason for this cast was that our integers are "untyped" (no `s` vs. `u`). I tried to make the UX a little smoother by assuming any integer > 0 was unsigned. This breaks with our normal type inference, though.

> If you want a float response, you just cast any of the results to float. I guess this could be achieved by casting the right side to {:u, precision} always when a {:s, precision} is given?

Yeah a user can always achieve a more specific result by casting.

native/explorer/src/series.rs
```diff
@@ -1232,64 +1232,49 @@ pub fn s_n_distinct(s: ExSeries) -> Result<usize, ExplorerError> {
...
-        }
+    let result = df_with_result.collect()?.column("result")?.clone();
+
+    Ok(ExSeries::new(result))
+}
+
+fn first(exseries: ExSeries) -> LiteralValue {
+    let dtype = exseries.dtype().clone();
+
+    match dtype {
```
lkarthee, line 1266: May be ?
```suggestion
fn first(s: ExSeries) -> LiteralValue {
    match s.dtype() {
```

billylanchantin: Oh yeah that's better :)

lib/explorer/series.ex
billylanchantin: @josevalim We could of course impose our own type restrictions if we want. We'd just need to pattern match on the dtypes.

Definitely doable :)

native/explorer/src/series.rs
```diff
@@ -1232,64 +1232,47 @@ pub fn s_n_distinct(s: ExSeries) -> Result<usize, ExplorerError> {
...
-                iter1
-                    .map(|v1| v1.and_then(|left| v2.map(|right| left.powf(right))))
-                    .collect()
-            } else {
-                iter1
-                    .zip(iter2)
-                    .map(|(v1, v2)| v1.and_then(|left| v2.map(|right| left.powf(right))))
-                    .collect()
-            };
+    let df_with_result = if s.len() == other.len() {
```
lkarthee, line 1235: maybe use match instead of if ?

billylanchantin: Is that possible in Rust? Here's how I'd do it in Elixir:

```elixir
case {length(base), length(exponent)} do
  {n, n} -> pow(base, exponent)        # <- Rust equivalent?
  {1, _} -> pow(first(base), exponent)
  {_, 1} -> pow(base, first(exponent))
end
```

lkarthee: https://github.com/elixir-explorer/explorer/pull/821/files#r1452655112

billylanchantin: Ahh, I didn't realize I could add predicates. Thanks!

billylanchantin:
Hey sorry for the churn! I was playing with a version that kept everything in Elixir but I didn't mean to push it yet.

billylanchantin:
Ok I think this new Elixir-only direction is better.

1. We're more in control of the types. @josevalim, it should now be easy to follow Nx if we want (we may need a post-cast in some cases though).
2. It helped catch some inconsistencies in other tests.
3. Less Rust code for me to mess up. 😛

lib/explorer/series.ex
```diff
@@ -3428,7 +3428,27 @@ defmodule Explorer.Series do
...
+  # TODO: ensure these types are correct.
+  defp cast_to_pow({:u, l}, {:u, r}), do: {:u, max(l, r)}
+  defp cast_to_pow({:u, _}, {:s, _}), do: {:f, 64}
+  defp cast_to_pow({:u, _}, {:f, f}), do: {:f, f}
+  defp cast_to_pow({:s, s}, {:u, u}), do: {:s, min(64, max(2 * u, s))}
+  defp cast_to_pow({:s, _}, {:s, _}), do: {:f, 64}
+  defp cast_to_pow({:s, _}, {:f, f}), do: {:f, f}
+  defp cast_to_pow({:f, f}, {:u, _}), do: {:f, f}
+  defp cast_to_pow({:f, f}, {:s, _}), do: {:f, f}
+  defp cast_to_pow({:f, f}, {:f, _}), do: {:f, f}
```
josevalim, line 3450:
```suggestion
  defp cast_to_pow({:u, l}, {:u, r}), do: {:u, max(l, r)}
  defp cast_to_pow({:s, s}, {:u, u}), do: {:s, min(64, max(2 * u, s))}
  defp cast_to_pow({:f, l}, {:f, r}), do: {:f, max(l, r)}
  defp cast_to_pow({:f, l}, {n, _}) when n in [:u, :s], do: {:f, l}
  defp cast_to_pow({n, _}, {:f, r}) when n in [:u, :s], do: {:f, r}
  defp cast_to_pow({n, _}, {:s, _}) when n in [:u, :s], do: {:f, 64}
  defp cast_to_pow(_, _), do: nil
```

josevalim (approved):
Smart use of DFs!

lib/explorer/series.ex
billylanchantin: This change will match Nx, yeah? I'm good with this stance. But I want to document it before we merge. Otherwise anyone coming in with Polars expectations will be thrown off.

billylanchantin:
Thanks, y'all!

> The idea here is to document for the end users, right? I'm 👍 for this.

@philss Yeah exactly. Raising on `pow(unsigned, signed)` is a totally sensible choice, but it differs from what Polars is doing (and Pandas I think) so I want to call it out in the docs.

lib/explorer/series.ex
billylanchantin: [058fea6](https://github.com/elixir-explorer/explorer/pull/821/commits/058fea6a0597054a35e3a9e21e33ba0100f80608)

I did this on a separate commit because we can't use infix `in` in this file.

Also I misunderstood what you were doing: you were just consolidating the cases. You weren't making `pow` work like Nx.

billylanchantin:
Note: I misunderstood what @josevalim was getting at with his suggestion above. He was only consolidating the cases, not making `pow` work like it does in Nx.

We can still do that if we want, but perhaps that should be a separate PR. If so, I think this one is good to go.

Give this comment a 👍 if we're good with that and I'll merge it. (Or someone can just merge it themselves 😄)

PR #822, opened 2024-01-15
Fix up `from_query` docs

cigrainger:
The examples for `DataFrame.from_query` show `from_query/2`, but this isn't defined. Params need to be passed, even if empty.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -8,11 +8,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1699770036,
-        "narHash": "sha256-bZmI7ytPAYLpyFNgj5xirDkKuAniOkj1xHdv5aIJ5GM=",
+        "lastModified": 1705299878,
+        "narHash": "sha256-UxAt2aUZFEZaAoNY1SrXgVJAkIXAKQN+dk/2nvoa3mY=",
         "owner": "nix-community",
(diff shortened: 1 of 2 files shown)


josevalim (commented):
At the time, we chose to require the parameters to avoid having two optional arguments that default to an empty list, which can be confusing.

Anyway, I am OK with merging this, your call!

cigrainger:
Yeah I agree it can be confusing. Perhaps better to document it correctly instead of changing it? Or, could `params` just be an `opt`?

josevalim:
:+1: for improving the docs too!

cigrainger:
Okay I've made `params` an opt and defaulted to `[]`.

josevalim:
Actually, I am not the biggest fan of this approach, sorry. :( Params are allegedly very common (or should be due for security), so I think it should be more first class than an option. Maybe we can do this:


```
@doc """
Shortcut for `query/4`.
"""
def query(conn, query, params_or_opts \\ []) when is_list(params_or_opts) do
  if Keyword.keyword?(params_or_opts) do
    query(conn, query, [], params_or_opts)
  else
    query(conn, query, params_or_opts, [])
  end
end
```

This way we can both as default but without the confusion? 🤔

cigrainger:
Ah, I didn't know they should be encouraged. In that case, I think better to leave it as is but document it -- I'll do taht.

PR #823, opened 2024-01-15
DataFrame.transpose/2 column names length check in elixir

lkarthee:


lib/explorer/data_frame.ex
```diff
@@ -4260,12 +4260,19 @@ defmodule Explorer.DataFrame do
...
+      cond do
+        columns == nil ->
+          for i <- 0..(df_length - 1), do: "column_#{i}"
+
+        length(columns) == df_length ->
+          columns
+
+        true ->
+          raise ArgumentError,
+                "lengths don't match: Length of new column names must be the same as the row count."
```
josevalim, line 4275:
```suggestion
                "length of :columns names #{length} must be the same as the row count (#{n_rows})"
```

josevalim:

PR #824, opened 2024-01-16
Support more integer dtypes in Series

philss:
This is the first part of the work I'm doing in https://github.com/elixir-explorer/explorer/issues/794

It supports the new signed and unsigned integer dtypes in operations for the Series module (and one small change in the DF). Mostly that we have here is regarding the "eager" implementation of a Series, although some bugs from the "lazy" series were solved as well.

I'm skipping the "pow/2" function because @billylanchantin is covering that in https://github.com/elixir-explorer/explorer/pull/821 :D

UPDATE 1: and `Series.concat/2` is being tackled in https://github.com/elixir-explorer/explorer/pull/812 by @lkarthee :D

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -3376,7 +3376,7 @@ defmodule Explorer.Series do
   defp cast_to_divide({:u, _}, {:s, _}), do: {:f, 64}
   defp cast_to_divide({int_type, _}, {:f, _} = float) when K.in(int_type, [:s, :u]), do: float
   defp cast_to_divide({:f, _} = float, {int_type, _}) when K.in(int_type, [:s, :u]), do: float
-  defp cast_to_divide({:f, _}, {:f, _}), do: {:f, 64}
+  defp cast_to_divide({:f, left}, {:f, right}), do: {:f, max(left, right)}
   defp cast_to_divide({:duration, p}, {:s, _}), do: {:duration, p}
   defp cast_to_divide({:duration, p}, {:f, _}), do: {:duration, p}
   defp cast_to_divide(_, _), do: nil
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -1840,7 +1840,7 @@ defmodule Explorer.Series do
       iex> Explorer.Series.rank(s, method: :ordinal)
       #Explorer.Series<
         Polars[3]
-        s64 [1, 2, 3]
+        u32 [1, 2, 3]
       >
 
       iex> s = Explorer.Series.from_list([ ~N[2022-07-07 17:44:13.020548], ~N[2022-07-07 17:43:08.473561], ~N[2022-07-07 17:45:00.116337] ])
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -3175,8 +3175,14 @@ defmodule Explorer.Series do
   defp cast_to_add({int_type, left}, {int_type, right}) when K.in(int_type, [:s, :u]),
     do: {int_type, max(left, right)}
 
-  defp cast_to_add({:s, _}, {:u, _}), do: {:s, 64}
-  defp cast_to_add({:u, _}, {:s, _}), do: {:s, 64}
+  defp cast_to_add({:s, s_size}, {:u, u_size}) when u_size >= s_size,
+    do: {:s, min(64, u_size * 2)}
+
+  defp cast_to_add({:u, u_size}, {:s, s_size}) when u_size >= s_size,
+    do: {:s, min(64, u_size * 2)}
+
+  defp cast_to_add({:s, s_size}, {:u, _}), do: {:s, s_size}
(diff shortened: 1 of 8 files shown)


lib/explorer/series.ex
```diff
@@ -3167,10 +3171,20 @@ defmodule Explorer.Series do
...
+    do: {:s, min(64, u_size * 2)}
+
+  defp cast_to_add({:u, u_size}, {:s, s_size}) when u_size >= s_size,
+    do: {:s, min(64, u_size * 2)}
+
+  defp cast_to_add({:s, s_size}, {:u, _}), do: {:s, s_size}
+  defp cast_to_add({:u, _}, {:s, s_size}), do: {:s, s_size}
+  defp cast_to_add({int_type, _}, {:f, _} = float) when K.in(int_type, [:s, :u]), do: float
+  defp cast_to_add({:f, _} = float, {int_type, _}) when K.in(int_type, [:s, :u]), do: float
   defp cast_to_add({:f, _}, {:f, _}), do: {:f, 64}
```
josevalim, line 3188: Merging between floats should be the max of their size. :)

josevalim: In Nx, the logic for merging types for +, - and * are all the same. So maybe we should add a function called `cast_numeric` that holds this shared logic. This way, we can write `cast_to_add` like this:

```
  defp cast_to_add(:date, {:duration, _}), do: :date
  defp cast_to_add({:duration, _}, :date), do: :date
  defp cast_to_add({:datetime, p}, {:duration, p}), do: {:datetime, p}
  defp cast_to_add({:duration, p}, {:datetime, p}), do: {:datetime, p}
  defp cast_to_add({:duration, p}, {:duration, p}), do: {:duration, p}
  defp cast_to_add(left, right), do: cast_numeric(left, right)
```

?

The same for `-` and `*`?

josevalim: Wait, maybe this logic should be added to `cast_for_arithmetic` instead? 🤔 That's where we already cast both sides. And then this function just check if both sides are the same? And then we likely don't need the changes in PolarsBackend.Series.subtract?

philss: I will check that! Currently these functions have different roles: `cast_for_arithmetic` really cast the series before perform the operation (at the Series module level), and `cast_to_add/subtract/divide/*` does not cast yet, but calculates the `out_dtype` so the backend (in this case `PolarsBackend.Series` or the lazy series) can cast.

lib/explorer/series.ex
```diff
@@ -3167,10 +3171,20 @@ defmodule Explorer.Series do
...
+    do: {int_type, max(left, right)}
+
+  defp cast_to_add({:s, s_size}, {:u, u_size}) when u_size >= s_size,
+    do: {:s, min(64, u_size * 2)}
+
+  defp cast_to_add({:u, u_size}, {:s, s_size}) when u_size >= s_size,
+    do: {:s, min(64, u_size * 2)}
+
+  defp cast_to_add({:s, s_size}, {:u, _}), do: {:s, s_size}
+  defp cast_to_add({:u, _}, {:s, s_size}), do: {:s, s_size}
```
billylanchantin, line 3185:
```suggestion
  defp cast_to_add({:s, s_size}, {:u, u_size}), do: min(64, max(s_size, 2 * u_size))
  defp cast_to_add({:u, s_size}, {:s, u_size}), do: min(64, max(s_size, 2 * u_size))
```

I think this works here (or wherever this logic ends up).

philss: @billylanchantin makes sense! Thanks!

lib/explorer/series.ex
```diff
@@ -3293,11 +3312,22 @@ defmodule Explorer.Series do
...
-  defp cast_to_multiply({:s, _}, {:f, _} = float), do: float
-  defp cast_to_multiply({:f, _} = float, {:s, _}), do: float
+  defp cast_to_multiply({int_type, left}, {int_type, right}) when K.in(int_type, [:s, :u]),
+    do: {int_type, max(left, right)}
+
+  defp cast_to_multiply({:s, s_size}, {:u, u_size}) when u_size >= s_size,
+    do: {:s, min(64, u_size * 2)}
+
+  defp cast_to_multiply({:u, u_size}, {:s, s_size}) when u_size >= s_size,
+    do: {:s, min(64, u_size * 2)}
```
josevalim, line 3322: If you want to remove the guard, you can write this as:

```elixir
  defp cast_to_multiply({:u, u_size}, {:s, s_size}),
    do: {:s, max(min(64, u_size * 2), s_size)}
```

lib/explorer/series.ex
philss: @josevalim I made the simplification in 566d9772216f089a728c23dd89b09816c1778893. I'm going to try that other change we talked soon.

josevalim (approved):

Issue #566, opened 2023-04-09
ISO 8601 DateTime format with timezone info can't be parsered


ray-sh:
Sample.csv
name,age,birth 
jack,11,2023-01-31T02:09:43.29Z
tom,2,2023-01-31T02:09:43.29Z

f1 = [{"name",:string},{"age",:integer},{"birth",:datetime}]
{:ok, sample} = DF.from_csv("sample.csv", max_rows: 100, dtypes: f1)

Explorer.DataFrame.table(f1), the datetime is nil from the output
+--------------------------------------------+
| Explorer DataFrame: [rows: 2, columns: 3]  |
+-------------+--------------+---------------+
|    name     |     age      |     birth     |
|  <string>   |  <integer>   |  <datetime>   |
+=============+==============+===============+
| jack     | 11           |               |
+-------------+--------------+---------------+
| tom          | 2            |               |
+-------------+--------------+---------------+

josevalim:
Can you please provide a sample `.csv` file in full. Because if your `sample.csv` is as you specify, then it means all of your column names have spaces in them.

ray-sh:
[sample.csv](https://github.com/elixir-nx/explorer/files/11185463/sample.csv)
When you remove the z from the datetime, then the issue is gone.

josevalim:
In this case can you please do a report on Polars? https://github.com/pola-rs/polars

Other than that, the alternative is to parse it as string, remove the Z, and then cast to `:date`.

lkarthee:
Someone reported it - https://github.com/pola-rs/polars/issues/6886.  

In this comment - https://github.com/pola-rs/polars/issues/6886#issuecomment-1435311157, they have changed the parse_dates to try_parse_dates. 

- some documentation about nuances of parsing dates in 10 minutes to explorer
- align the explorer flag with polars flag as it has some context to it  

This issue will be encountered when exporting data without any additional processing from table with columns which are timestamp without timezone   - as ~U[2022-10-27 04:20:15Z] serialised to 2022-10-27T04:20:15Z . Additional processing the field like NaiveDateTime.to_iso8601 will dump to polars expected format. 

```elixir
df = DF.from_csv!(file_path, opts)
df = DF.put(df, :inserted_at, Explorer.Series.transform(df[:inserted_at], &NaiveDateTime.from_iso8601!/1))
```

josevalim:
Yes, we should document or, even better, make sure we raise a nice error message.

PR #826, opened 2024-01-18
fix Series.from_list/2 bugs

lkarthee:
Read this comment for information. https://github.com/elixir-explorer/explorer/pull/826#issuecomment-1898777688

Added some tests as i could not create a series with u64 value

```elixir
s = Series.from_list([9_223_372_036_854_775_807, 9_223_372_036_854_775_806, 9_223_372_036_854_775_808], dtype: {:s, 32}) # error

s = Series.from_list([9_223_372_036_854_775_808]) # inferred as s64 and fails

s = Series.from_list([1, 2, 3, 9_223_372_036_854_775_808]) # mismatch s64 and u64

s = Series.from_list([9_223_372_036_854_775_808, 1, 2, 3]) # mismatch u64 and s64 

s = Series.from_list([1, 2, 3], dtype: {:u, 64}) # as u64 fails

s = Series.from_list([-1, 2, 3], dtype: {:u, 64}) 
```

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -251,6 +251,7 @@ defmodule Explorer.Series do
     * `:dtype` - Cast the series to a given `:dtype`. By default this is `nil`, which means
       that Explorer will infer the type from the values in the list.
       See the module docs for the list of valid dtypes and aliases.
+    * `:strict` - `boolean` defaults to `false`. When `true` infers exact data type of integers and raises on overflow and underflow
 
   ## Examples
 
@@ -406,12 +407,12 @@ defmodule Explorer.Series do
   @doc type: :conversion
   @spec from_list(list :: list(), opts :: Keyword.t()) :: Series.t()
   def from_list(list, opts \\ []) do
-    opts = Keyword.validate!(opts, [:dtype, :backend])
+    opts = Keyword.validate!(opts, [:dtype, :backend, strict: false])
     backend = backend_from_options!(opts)
 
     normalised_dtype = if opts[:dtype], do: Shared.normalise_dtype!(opts[:dtype])
 
-    type = Shared.dtype_from_list!(list, normalised_dtype)
+    type = Shared.dtype_from_list!(list, normalised_dtype, opts[:strict])
     list = Shared.cast_numerics(list, type)
 
     series = backend.from_list(list, type)
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -251,7 +251,7 @@ defmodule Explorer.Series do
     * `:dtype` - Cast the series to a given `:dtype`. By default this is `nil`, which means
       that Explorer will infer the type from the values in the list.
       See the module docs for the list of valid dtypes and aliases.
-    * `:strict` - `boolean` defaults to `false`. When `true` infers exact data type of integers and raises on overflow and underflow
+    * `:strict` - when `true` raises on overflow and underflow - defaults to `false`.
 
   ## Examples
 
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -275,9 +275,7 @@ defmodule Explorer.Shared do
   def dtype_from_list!(_list, :null, _strict), do: :null
 
   def dtype_from_list!(list, nil, strict) do
-    list
-    |> Enum.reduce(:null, &infer_type(&1, &2, nil, strict))
-    |> normalise_dtype!()
+    Enum.reduce(list, :null, &infer_type(&1, &2, nil, strict))
   end
 
   def dtype_from_list!(list, preferred_type, strict) do
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -74,7 +74,6 @@ defmodule Explorer.Shared do
   def normalise_dtype(:u16), do: {:u, 16}
   def normalise_dtype(:u32), do: {:u, 32}
   def normalise_dtype(:u64), do: {:u, 64}
-  def normalise_dtype({:s, -64}), do: {:s, 64}
   def normalise_dtype(_dtype), do: nil
 
   @doc """
@@ -258,9 +257,7 @@ defmodule Explorer.Shared do
   Gets the `dtype` of a list or raise error if not possible.
   """
   def dtype_from_list!(list) do
-    list
-    |> Enum.reduce(:null, &infer_type/2)
-    |> normalise_dtype!()
+    Enum.reduce(list, :null, &infer_type/2)
   end
 
   @doc """
@@ -279,34 +276,25 @@ defmodule Explorer.Shared do
 
   def dtype_from_list!(list, preferred_type) do
     list
-    |> Enum.reduce(:null, &infer_type/2)
-    |> merge_preferred!(preferred_type, list)
+    |> dtype_from_list!()
+    |> merge_preferred(preferred_type)
   end
 
   @non_finite [:nan, :infinity, :neg_infinity]
 
   defp infer_type(nil, type), do: type
   defp infer_type(item, :null), do: infer_type(item)
-  defp infer_type(integer, {:u, 64}) when is_integer(integer) and integer > 0, do: {:u, 64}
   defp infer_type(integer, {:f, 64}) when is_integer(integer), do: {:f, 64}
-  defp infer_type(float, {:u, _}) when is_float(float) or float in @non_finite, do: {:f, 64}
-  defp infer_type(float, {:s, _}) when is_float(float) or float in @non_finite, do: {:f, 64}
+  defp infer_type(float, {:s, 64}) when is_float(float) or float in @non_finite, do: {:f, 64}
   defp infer_type(list, {:list, type}) when is_list(list), do: infer_list(list, type)
   defp infer_type(%{} = map, {:struct, inner}), do: infer_struct(map, inner)
 
   defp infer_type(item, type) do
-    case {infer_type(item), type} do
-      {{:u, 64}, {:s, 64}} ->
-        {:u, 64}
-
-      {{:s, m}, {:s, n}} when m < 0 or n < 0 ->
-        type
-
-      {type, type} ->
-        type
-
-      _ ->
-        raise_infer_mismatch!(item, type)
+    if infer_type(item) == type do
+      type
+    else
+      raise ArgumentError,
+            "the value #{inspect(item)} does not match the inferred dtype #{inspect(type)}"
     end
   end
 
@@ -314,15 +302,7 @@ defmodule Explorer.Shared do
   defp infer_type(%Time{} = _item), do: :time
   defp infer_type(%NaiveDateTime{} = _item), do: {:datetime, :microsecond}
   defp infer_type(%Explorer.Duration{precision: precision} = _item), do: {:duration, precision}
-
-  defp infer_type(item) when is_integer(item) do
-    cond do
-      item < 0 -> {:s, -64}
-      item > 9_223_372_036_854_775_807 -> {:u, 64}
-      true -> {:s, 64}
-    end
-  end
-
+  defp infer_type(item) when is_integer(item), do: {:s, 64}
   defp infer_type(item) when is_float(item) or item in @non_finite, do: {:f, 64}
   defp infer_type(item) when is_boolean(item), do: :boolean
   defp infer_type(item) when is_binary(item), do: :string
@@ -347,102 +327,31 @@ defmodule Explorer.Shared do
             {key, infer_type(value, type)}
 
           true ->
-            raise_infer_mismatch!(map, {:struct, types})
+            raise ArgumentError,
+                  "the value #{inspect(map)} does not match the inferred dtype #{inspect({:struct, types})}"
         end
       end
 
     {:struct, types}
   end
 
-  defp infer_as_uint_type!(nil, type), do: type
+  defp merge_preferred(type, type), do: type
+  defp merge_preferred(:null, type), do: type
+  defp merge_preferred({:s, 64}, {:s, _} = type), do: type
+  defp merge_preferred({:s, 64}, {:f, _} = type), do: type
+  defp merge_preferred({:f, 64}, {:f, _} = type), do: type
+  defp merge_preferred(:string, type) when type in [:binary, :string, :category], do: type
 
-  defp infer_as_uint_type!(item, {:u, n} = type) do
-    cond do
-      item < 0 -> raise_infer_mismatch!(item, type)
-      n == 8 and item < 256 -> type
-      n == 16 and item < 2 ** 16 -> type
-      n == 32 and item < 2 ** 32 -> type
-      n == 64 and item < 2 ** 64 -> type
-      true -> raise_infer_mismatch!(item, type)
-    end
+  defp merge_preferred({:list, inferred}, {:list, preferred}) do
+    {:list, merge_preferred(inferred, preferred)}
   end
 
-  defp infer_as_int_type!(nil, type), do: type
-
-  defp infer_as_int_type!(item, {:s, n} = type) do
-    cond do
-      item < -9_223_372_036_854_775_808 or item > 9_223_372_036_854_775_807 ->
-        raise_infer_mismatch!(item, type)
-
-      n == 8 and item > -129 and item < 128 ->
-        type
-
-      n == 16 and item > -32_769 and item < 32_768 ->
-        type
-
-      n == 32 and item > -2_147_483_649 and item < 2_147_483_648 ->
-        type
-
-      n == 64 and item > -9_223_372_036_854_775_809 and item < 9_223_372_036_854_775_808 ->
-        type
-
-      true ->
-        raise_infer_mismatch!(item, type)
-    end
+  defp merge_preferred({:struct, inferred}, {:struct, preferred}) do
+    {:struct, Map.merge(inferred, preferred, fn _, v1, v2 -> merge_preferred(v1, v2) end)}
   end
 
-  defp merge_preferred!(type, type, _list), do: type
-  defp merge_preferred!(:null, type, _list), do: type
-  defp merge_preferred!({t, 64}, {:f, _} = type, _list) when t in [:s, :u], do: type
-
-  defp merge_preferred!({:s, _}, {:s, _} = type, list) do
-    Enum.reduce(list, type, &infer_as_int_type!/2)
-  end
-
-  defp merge_preferred!({:s, _}, {:u, _} = type, list) do
-    Enum.reduce(list, type, &infer_as_uint_type!/2)
-  end
-
-  defp merge_preferred!({:u, _}, {:s, _} = type, list) do
-    Enum.reduce(list, type, &infer_as_int_type!/2)
-  end
-
-  defp merge_preferred!({:u, _}, {:u, _} = type, list) do
-    Enum.reduce(list, type, &infer_as_uint_type!/2)
-  end
-
-  defp merge_preferred!({:f, 64}, {:f, _} = type, _list), do: type
-  defp merge_preferred!(:string, type, _list) when type in [:binary, :string, :category], do: type
-
-  defp merge_preferred!({:list, {x, 64} = inferred}, {:list, {y, _} = preferred}, list)
-       when x in [:s, :u] or y in [:s, :u] do
-    result = Enum.reduce(list, preferred, fn l, acc -> merge_preferred!(inferred, acc, l) end)
-    {:list, result}
-  end
-
-  defp merge_preferred!({:list, inferred}, {:list, preferred}, list) do
-    {:list, merge_preferred!(inferred, preferred, list)}
-  end
-
-  defp merge_preferred!({:struct, inferred}, {:struct, preferred}, list) do
-    result =
-      Map.merge(inferred, preferred, fn k, v1, v2 ->
-        v_list = Enum.map(list, & &1[k])
-        merge_preferred!(v1, v2, v_list)
-      end)
-
-    {:struct, result}
-  end
-
-  defp merge_preferred!(inferred, _preferred, _list) do
+  defp merge_preferred(inferred, _preferred) do
     inferred
-  end
-
-  defp raise_infer_mismatch!(item, type) do
-    type = if type == {:s, -64}, do: {:s, 64}, else: type
-
-    raise ArgumentError,
-          "the value #{inspect(item)} does not match the inferred dtype #{inspect(type)}"
   end
 
   @doc """
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -410,7 +410,6 @@ defmodule Explorer.Shared do
     do: raise(ArgumentError, "unsupported datatype: #{inspect(item)}")
 
   defp infer_list(list, type, preferred, strict) do
-    # &infer_type/2)}
     preferred =
       case preferred do
         {:list, preferred} -> preferred
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -74,7 +74,6 @@ defmodule Explorer.Shared do
   def normalise_dtype(:u16), do: {:u, 16}
   def normalise_dtype(:u32), do: {:u, 32}
   def normalise_dtype(:u64), do: {:u, 64}
-  def normalise_dtype({:s, -64}), do: {:s, 64}
   def normalise_dtype(_dtype), do: nil
 
   @doc """
@@ -259,9 +258,6 @@ defmodule Explorer.Shared do
   """
   def dtype_from_list!(list) do
     dtype_from_list!(list, nil, false)
-    # list
-    # |> Enum.reduce(:null, &infer_type(&1, &2, nil, false))
-    # |> normalise_dtype!()
   end
 
   @doc """
@@ -275,9 +271,9 @@ defmodule Explorer.Shared do
   If no preferred type is given (nil), then the inferred type is returned.
   """
   def dtype_from_list!(list, preferred_type, strict \\ false)
+
   def dtype_from_list!(_list, :null, _strict), do: :null
 
-  # , do: dtype_from_list!(list, strict)
   def dtype_from_list!(list, nil, strict) do
     list
     |> Enum.reduce(:null, &infer_type(&1, &2, nil, strict))
@@ -295,17 +291,13 @@ defmodule Explorer.Shared do
   defp infer_type(nil, type, _preferred, _strict), do: type
   defp infer_type(item, :null, preferred, strict), do: infer_type(item, preferred, strict)
 
-  defp infer_type(integer, {:u, 64}, _preferred, _strict)
-       when is_integer(integer) and integer > 0,
-       do: {:u, 64}
-
   defp infer_type(integer, {:f, 64}, _preferred, _strict) when is_integer(integer), do: {:f, 64}
 
   defp infer_type(float, {:s, _}, _preferred, _strict)
        when is_float(float) or float in @non_finite,
        do: {:f, 64}
 
-  defp infer_type(float, {:u, 64}, _preferred, _strict)
+  defp infer_type(float, {:u, _}, _preferred, _strict)
        when is_float(float) or float in @non_finite,
        do: {:f, 64}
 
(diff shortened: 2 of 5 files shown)


lkarthee:
```
Finished in 6.2 seconds (6.2s async, 0.00s sync)
577 doctests, 1389 tests, 17 failures, 18 excluded, 1 skipped
```

https://github.com/elixir-explorer/explorer/actions/runs/7566406181/job/20603730541?pr=826

josevalim:
In my opinion, we should continue inferring those cases as s64 and continue raising if we overflow or underflow the dtype. The improvement we could do is to improve the error message.

The conversion to u64, s64, etc happens inside Rustler and I am not sure if Rustler let's us customize the error message.

@philss, do you know if we can customize the "argument error" bits here?

```
iex(1)> series = Explorer.Series.from_list([[1, 9_223_372_036_854_775_808]])
** (ArgumentError) argument error
    (explorer 0.8.0-dev) Explorer.PolarsBackend.Native.s_from_list_s64("", [1, 9223372036854775808])
```

In case we cannot, I would wrap "Explorer.PolarsBackend.Shared.from_list" into a `try/rescue ArgumentError` and raise a better message there.

josevalim:
Btw, `s = Series.from_list([1, 2, 3], dtype: {:u, 64}) # as u64 fails` works for me (and it seems it works on CI too).

lkarthee:
https://hexdocs.pm/explorer/Explorer.Series.html#from_list/2

- `:dtype` - Cast the series to a given `:dtype`. By default this is nil, which means that Explorer will infer the type from the values in the list.

I have established the base line of  what is not working scenarios in https://github.com/elixir-explorer/explorer/actions/runs/7570277474/job/20615312348 

`from_list/2` does not handle u64 and dtypes  {:s, _} and {:u, _} well. 

In this PR, I have attempted to address the below:
 - dtype is inferred properly especially {:u, 64}
 - descriptive error is thrown in case it fails.

Please review - I am pretty sure there is room for improvement . Also take a stab at it improving the PR. I can learn more if you can attempt to improvise thing by seeing how others do it in a better way. 

I have tested these scenarios in Polars and it handles well - either it throws descriptive error, or creates series as expected.

josevalim:
As per above, I don’t think we should infer u64. Operations such as subtract can be confusing, especially coming from Elixir, and it should always be opt-in. For the cases listed here, we should improve the error message. I would not change the inference code.

josevalim:
I also think doing additional traversals on the list will be expensive. Hence my suggestion of wrapping around the calls to Rust. :)

billylanchantin:
@lkarthee I had the same temptation but I think @josevalim is right: we should keep the inference code as is.

We could follow the pattern you've established where we infer the "minimal" type that can represent the integers given to `from_list`. But the problem is that many operations are not closed under those types. Unsigned subtraction (as mentioned), but even addition: U8 + U8 is very likely to result in an overflow.

Also I got the following in Polars:

* https://www.rustexplorer.com/b/psb7pj

```rust
let s1 = Series::new("s1", [1, 2]);
// shape: (2,)
// Series: 's1' [i32]
// [
// 	1
// 	2
// ]
```

So it looks like they too infer signed integers by default (their `i32` is our `s32`).

lkarthee:
Multiple parts to the story:
 - majority of new tests which are failing should pass (even with modifications) in explorer.
 - descriptive error which don't look like crashes.
 - `from_lists/2` anything relating to `{:u, 64}` data type exclusive values results in error.  
 - Inference of types - when `dtype` is `nil`.
 - Are we going to change existing behaviour which is documented about inference when `dtype` is nil, even though it might be broken . This might be a deviation from polars, we have deviated before.
 -  Preferred type merging with inference type.
 
All code snippets are using  explorer/main SHA - 62e77160ec37445be47288b8bcf341dc1d96821e 
## `from_list/2` bug in depth

### 1. Loss of information
Loss of information without any warning - `-1` to `nil`. I suspect it has something to do with rustler (might be wrong). One thing to note is the numbers are {:s , 64} not valid or exclusive {:u 64} numbers.
```elixir
alias Explorer.Series
Series.from_list([-1, 9_223_372_036_854_775_801], dtype: {:u, 64}) 
#Explorer.Series<
  Polars[2]
  u64 [nil, 9223372036854775801]
>
```
### 2. Cannot create {:u, 64}, list of {:u, 64} and struct of {:u, 64}
Now let's pass atleast the values in range and attempt to create {:u, 64} series. It is invoking rustler func which accepts only s64 hence the error.
```elixir
Series.from_list([1, 9_223_372_036_854_775_809], dtype: {:u, 64}).
** (ArgumentError) argument error
    (explorer 0.8.0-dev) Explorer.PolarsBackend.Native.s_from_list_s64("", [1, 9223372036854775809])
    (explorer 0.8.0-dev) lib/explorer/polars_backend/series.ex:24: Explorer.PolarsBackend.Series.from_list/2
    (explorer 0.8.0-dev) lib/explorer/series.ex:387: Explorer.Series.from_list/2
    #cell:phle4en4dbmtgjco:1: (file)
```
Let's attempt to create a `{:list, {:u, 64}}`. 
```elixir
series = Series.from_list([[1], [9_223_372_036_854_775_809]], dtype: {:list, {:u, 64}})
** (ArgumentError) argument error
    (explorer 0.8.0-dev) Explorer.PolarsBackend.Native.s_from_list_s64("", [9223372036854775809])
    (elixir 1.15.7) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
    (elixir 1.15.7) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
    (explorer 0.8.0-dev) lib/explorer/polars_backend/shared.ex:114: Explorer.PolarsBackend.Shared.from_list/3
    (explorer 0.8.0-dev) lib/explorer/polars_backend/series.ex:24: Explorer.PolarsBackend.Series.from_list/2
    (explorer 0.8.0-dev) lib/explorer/series.ex:387: Explorer.Series.from_list/2
    #cell:aqfn6tqdn2be45yv:1: (file)
```
Lets attempt to create a struct with field of u64 using dtype `{:struct, %{"a" => {:list, {:u, 64}}}}`.
```elixir 
Series.from_list([%{"a" => [1]}, %{"a" => [9_223_372_036_854_775_809]}], dtype: {:struct, %{"a" => {:list, {:u, 64}}}})
** (ArgumentError) argument error
    (explorer 0.8.0-dev) Explorer.PolarsBackend.Native.s_from_list_s64("a", [9223372036854775809])
    (elixir 1.15.7) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
    (elixir 1.15.7) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
    (explorer 0.8.0-dev) lib/explorer/polars_backend/shared.ex:114: Explorer.PolarsBackend.Shared.from_list/3
    (explorer 0.8.0-dev) lib/explorer/polars_backend/shared.ex:126: anonymous fn/3 in Explorer.PolarsBackend.Shared.from_list/3
    (stdlib 5.1.1) maps.erl:416: :maps.fold_1/4
    (explorer 0.8.0-dev) lib/explorer/polars_backend/shared.ex:123: Explorer.PolarsBackend.Shared.from_list/3
    #cell:qcjmwzxpdlnvdsd4:1: (file)
```

### 3. Descriptive Error messages

Lets try to pass a wrong data type as dtype - instead of `{:list, {:s, 8}}`, pass `{:s , 8}`. Cryptic Polars Error.

```elixir
series = Series.from_list([[1], [2]], dtype: {:s, 8})
** (RuntimeError) Polars Error: cannot cast List type (inner: 'Int64', to: 'Int8')
    (explorer 0.8.0-dev) lib/explorer/polars_backend/shared.ex:26: Explorer.PolarsBackend.Shared.apply_series/3
    #cell:7kssor3tml27to2y:1: (file)
```

Py Polars I get below 

```python
pl.Series([[1], [2]], dtype = pl.Int8)
... long stack trace omitted  but  has descriptive error message relating to problem.
TypeError: 'list' object cannot be interpreted as an integer

```

### 4. Inference
This is a documented behaviour in docs.  If we are changing it - documentation should be updated about divergence. 
I feel the docs are correct and divergence is not needed. Let's look at the fix in PR  w.r.to inference. We are already iterating the list and if there was an integer we assume it is {:s, 64}.  I have changed it to return 3 tuples: 
 - {:s, 64} for range 0 to 2**63-1
 - {:s, -64} if list contains negative integers
 - {:u, 64} if list has values greater than 2 ** 63 to 2 ** 64

Then normalized {:s, -64} to {:s, 64}. This is all I did w.r.to Inference. We can may be quantify overhead (using benchmarks)  approx 10 additional comparisons. These kick in only w.r.to integers.

### 5. Merge Preferred type
Even with explicit `dtype: {:u, 64}`, `from_list/2` is failing or we are losing information. We are using previous information [{:s, 64}, {:s, -64}, {:u, 64}] to check valid ranges of various integers. Only for {:struct, } we are creating additional data - I am sure you can find some solution here. Rest all cases we are iterating list again. In some cases it just raises immediately like {:u, 64} to any thing other than {:U, 64}. This overhead only for cases of integers. 

### 6. Additional tests
Current tests are inadequate as CI is passing. I have thoroughly tested add new test cases. What ever solution/fix may be majority of the test cases are valid tests. There may be modifications to error messages, but rest should pass. We can address what needs to be merged after a resolution or fix is finalised.

# TLDR:
My take on it: 
 -  descriptive error messages are important. It will be very frustrating for users to figure out things and wasting time. How we are going to achieve it is an open question. This PR is not the only way.
 - merge preferred dtype with inferred dtype has some overhead especially w.r.to structs.
 - we can have a flag to disable merge preferred dtype validations.
 - we can defer structs  validations to polars errors
 - inference changes don't add much overhead and enable us to infer {:u, 64}
 - may be we should run benchmarks on how much overhead we are adding
 - majority test cases are very much valid 
 - docs should be updated if we are diverging from documented behaviour
 - Py polars are doing it - should we look into how they are doing it ?

billylanchantin:
@lkarthee I'm totally with you when it comes to docs, error messages, and tests. The only part of this that's tripping me up is the change to the inference code. It's not about overhead, but ergonomics.

The stance we're taking is that `:s64` is the most ergonomic integer dtype to infer. If you infer a list `[1, 2, 3]` as having that dtype, you can do most addition and subtraction operations without worrying about overflow. And because you can supply a more specific dtype if you need to, this seems like a reasonable trade-off. Without it, users would regularly experience overflow gotchas.

For the specifics:

1. Loss of information
    ```elixir
    [-1, 1] |> Series.from_list(dtype: :u8) |> Series.to_list() #=> [nil, 1]
    ```
    I _think_ this is the expected behavior. Is it documented as being different elsewhere?

2. Cannot create {:u, 64}, list of {:u, 64} and struct of {:u, 64}
    ```elixir
    Series.from_list([2**63], dtype: {:u, 64}) #=> raises
    ```
    etc. ought to work. This does seem like a bug!

3. Descriptive Error messages
    I agree we should have better error messages.

4. Inference
    See above.

5. Merge Preferred type
    No opinion yet. Will need to see the code after reverting the inference changes.

6. Additional tests
    We should definitely add tests for the bugs. Though any which test the change to the inference code will need to be dropped.

lkarthee:
> @lkarthee I'm totally with you when it comes to docs, error messages, and tests. 

Thank you @billylanchantin sharing your view and for reading such a long post - I made a mistake of pushing tests before writing cover. Perhaps that caused confusion. Even this PR is complex. I tend not mix multiple things, but I could not scope it any lesser. 


> The only part of this that's tripping me up is the change to the inference code. It's not about overhead, but ergonomics.

Just to be clear - We are still defaulting to {:s, 64} unless we have a {:u, 64} value in list. We are not going into subtypes of `:s` or `:u`. There won't be inference of {s, 8} when dtype is not passed. 


> 1. Loss of information
```
[-1, 1] |> Series.from_list(dtype: :u8) |> Series.to_list() #=> [nil, 1]
```
I think this is the expected behavior. Is it documented as being different elsewhere?



https://hexdocs.pm/explorer/Explorer.Series.html#module-creating-series

> The list must consist of a single data type and nils. 

Option :dtype documentation.
> :dtype - Cast the series to a given :dtype. By default this is nil, which means that Explorer will infer the type from the values in the list.

There is no word of defaulting to s64. It says it will infer - gives example of integers, floats, etc. 

At the bottom of section it says: 
 
> Mixing non-numeric data types will raise an ArgumentError:
Explorer.Series.from_list([1, "a"])
** (ArgumentError) the value "a" does not match the inferred series dtype :integer

The precedent here is error w.r.to incompatible values not silent cast to nil. 

Integer split to subtypes is not yet released (if I am not wrong). 

We can look at Py Polars behaviour. 
```python
pl.Series([-1,1], dtype=pl.UInt8) 
# throws incompatible types error. 
```

I had personally faced a scenario where dates were not parsed - https://github.com/elixir-explorer/explorer/issues/566#issuecomment-1546508171 .

It's very frustrating when things fail silently. 

If data is wrong or I am using wrong type, I want it to fail with a big error. Else what is the point having types?

josevalim:
Since there are many bugs here, we should probably try to fix them individually instead of all at once.

As a starter, I would look into:

> `[-1, 1] |> Series.from_list(dtype: :u8) |> Series.to_list() #=> [nil, 1]`

The above should raise. It doesn't need to raise something good for now but it should raise.

> Series.from_list([2**63], dtype: {:u, 64})

This should not raise.

> There is no word of defaulting to s64. It says it will infer - gives example of integers, floats, etc.

Yeah, that's because s64 is new. We should say a list of integers are always inferred as s64. A list of floats (and integers) is always f64 by default.

josevalim:
> Series.from_list([2**63], dtype: {:u, 64})

I have pushed a fix for this. Now it works accordingly. :)

josevalim:
> Yeah, that's because s64 is new. We should say a list of integers are always inferred as s64. A list of floats (and integers) is always f64 by default.

Fixed in main too.

lkarthee:
I have changed the pr to add `strict` flag to raise if there is an overflow or underflow. Py Polars has similar flag.

Default integer is still {:s, 64}.

If strict is false, which is default - there is no performance impact. 

Just to be clear  - earlier implementation is scrapped. There is no merge logic. Only iterate list once which was already being done. No copy, no multiple iterations. 

Currently overflow and underflow can occur. 

```elixir
Series.from_list([0, 256], dtype: {:u, 8})
#Explorer.Series<
  Polars[2]
  u8 [0, 0]
>
```

<details>
<summary>Benchmark script & results</summary>

```elixir
datetime = DateTime.utc_now()
seed = DateTime.to_unix(datetime)
key = Nx.Random.key(seed)

{randints, _new_key} = Nx.Random.randint(key, -2**63, 2**63-1, shape: {100, 1000}, type: :s64)
list_1k = Nx.to_list(randints)


Benchee.run(
  %{
    "1_b_100x1000" => fn -> Enum.map(list_1k, fn x -> Explorer.Series.from_list(x) end) end,
    "3_d_100x1000_s64" => fn -> Enum.map(list_1k, fn x -> Explorer.Series.from_list(x, dtype: {:s, 64}) end) end,

    "2_b_100x1000_main" => fn -> Enum.map(list_1k, fn x -> Explorer.Series.from_list(x, old: true) end) end,
    "4_d_100x1000_s64_main" => fn -> Enum.map(list_1k, fn x -> Explorer.Series.from_list(x, dtype: {:s, 64}, old: true) end) end,

    "5_b_100x1000_strict" => fn -> Enum.map(list_1k, fn x -> Explorer.Series.from_list(x, strict: true) end) end,
    "6_d_100x1000_s64_strict" => fn -> Enum.map(list_1k, fn x -> Explorer.Series.from_list(x, dtype: {:s, 64}, strict: true) end) end,
  }
)
```
Here are the results. Don't see any abnormal increase in time. 

```
Operating System: macOS
CPU Information: Apple M1 Pro
Number of Available Cores: 10
Available memory: 16 GB
Elixir 1.15.7
Erlang 26.1.2
JIT enabled: true

warmup: 2 s
time: 5 s
memory time: 0 ns
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 42 s

Benchmarking 1_b_100x1000 ...
Benchmarking 2_b_100x1000_main ...
Benchmarking 3_d_100x1000_s64 ...
Benchmarking 4_d_100x1000_s64_main ...
Benchmarking 5_b_100x1000_strict ...
Benchmarking 6_d_100x1000_s64_strict ...
Calculating statistics...
Formatting results...

Name                              ips        average  deviation         median         99th %
4_d_100x1000_s64_main           71.18       14.05 ms     ±3.29%       13.94 ms       16.04 ms
3_d_100x1000_s64                71.02       14.08 ms     ±3.44%       13.95 ms       16.24 ms
1_b_100x1000                    70.98       14.09 ms     ±3.84%       13.95 ms       16.50 ms
2_b_100x1000_main               70.85       14.11 ms     ±4.42%       13.95 ms       16.83 ms
5_b_100x1000_strict             64.73       15.45 ms     ±2.87%       15.33 ms       17.41 ms
6_d_100x1000_s64_strict         61.77       16.19 ms     ±2.76%       16.07 ms       18.07 ms

Comparison:
4_d_100x1000_s64_main           71.18
3_d_100x1000_s64                71.02 - 1.00x slower +0.0305 ms
1_b_100x1000                    70.98 - 1.00x slower +0.0401 ms
2_b_100x1000_main               70.85 - 1.00x slower +0.0643 ms
5_b_100x1000_strict             64.73 - 1.10x slower +1.40 ms
6_d_100x1000_s64_strict         61.77 - 1.15x slower +2.14 ms
```
</details>

lkarthee:
You can close the PR if you don't feel we need strict.

lib/explorer/shared.ex
```diff
@@ -270,71 +274,176 @@ defmodule Explorer.Shared do
...
+      {:u, _} ->
+        raise_mismatched_dtype!(item, preferred)
+
+      _ ->
+        {:s, 64}
+    end
+  end
+
+  defp infer_type(item, _preferred, _strict) when is_float(item) or item in @non_finite,
+    do: {:f, 64}
```
josevalim, line 399: Btw, we also have precision issues with floats, since f32 cannot represent as large numbers as f64. :)

lib/explorer/shared.ex
```diff
@@ -270,71 +274,176 @@ defmodule Explorer.Shared do
...
-  defp infer_type(integer, {:f, 64}) when is_integer(integer), do: {:f, 64}
-  defp infer_type(float, {:s, 64}) when is_float(float) or float in @non_finite, do: {:f, 64}
-  defp infer_type(list, {:list, type}) when is_list(list), do: infer_list(list, type)
-  defp infer_type(%{} = map, {:struct, inner}), do: infer_struct(map, inner)
+  defp infer_type(nil, type, _preferred, _strict), do: type
+  defp infer_type(item, :null, preferred, strict), do: infer_type(item, preferred, strict)
+
+  defp infer_type(integer, {:u, 64}, _preferred, _strict)
+       when is_integer(integer) and integer > 0,
+       do: {:u, 64}
```
josevalim, line 300: Do we need this clause? I think we can remove it and let it fallback to infer_type/3.

lib/explorer/shared.ex
```diff
@@ -270,71 +274,176 @@ defmodule Explorer.Shared do
...
+
+  defp infer_type(integer, {:f, 64}, _preferred, _strict) when is_integer(integer), do: {:f, 64}
+
+  defp infer_type(float, {:s, _}, _preferred, _strict)
+       when is_float(float) or float in @non_finite,
+       do: {:f, 64}
+
+  defp infer_type(float, {:u, 64}, _preferred, _strict)
```
josevalim, line 308:
```suggestion
  defp infer_type(float, {:u, _}, _preferred, _strict)
```

lib/explorer/shared.ex
```diff
@@ -270,71 +274,176 @@ defmodule Explorer.Shared do
 
   If no preferred type is given (nil), then the inferred type is returned.
   """
-  def dtype_from_list!(_list, :null), do: :null
+  def dtype_from_list!(list, preferred_type, strict \\ false)
+  def dtype_from_list!(_list, :null, _strict), do: :null
 
-  def dtype_from_list!(list, nil), do: dtype_from_list!(list)
+  # , do: dtype_from_list!(list, strict)
```
josevalim, line 280:
```suggestion
```

lib/explorer/shared.ex
```diff
@@ -74,6 +74,7 @@ defmodule Explorer.Shared do
   def normalise_dtype(:u16), do: {:u, 16}
   def normalise_dtype(:u32), do: {:u, 32}
   def normalise_dtype(:u64), do: {:u, 64}
+  def normalise_dtype({:s, -64}), do: {:s, 64}
```
josevalim, line 77:
```suggestion
```

lib/explorer/shared.ex
```diff
@@ -257,7 +258,10 @@ defmodule Explorer.Shared do
   Gets the `dtype` of a list or raise error if not possible.
   """
   def dtype_from_list!(list) do
-    Enum.reduce(list, :null, &infer_type/2)
+    dtype_from_list!(list, nil, false)
+    # list
+    # |> Enum.reduce(:null, &infer_type(&1, &2, nil, false))
+    # |> normalise_dtype!()
```
josevalim, line 264:
```suggestion
```

lib/explorer/shared.ex
```diff
@@ -270,71 +274,176 @@ defmodule Explorer.Shared do
...
 
-  defp infer_list(list, type) do
-    {:list, Enum.reduce(list, type, &infer_type/2)}
+  defp infer_type(%Explorer.Duration{precision: precision} = _item, _preferred, _strict),
+    do: {:duration, precision}
+
+  defp infer_type(item, _, false) when is_integer(item), do: {:s, 64}
+
+  defp infer_type(item, nil, true) when is_integer(item) do
+    if item > 9_223_372_036_854_775_807 do
```
josevalim, line 351: We need to check the negative bound too.

lib/explorer/shared.ex
josevalim: But that should probably be another pull request.

lib/explorer/series.ex
```diff
@@ -251,6 +251,7 @@ defmodule Explorer.Series do
     * `:dtype` - Cast the series to a given `:dtype`. By default this is `nil`, which means
       that Explorer will infer the type from the values in the list.
       See the module docs for the list of valid dtypes and aliases.
+    * `:strict` - when `true` raises on overflow and underflow - defaults to `false`.
```
josevalim, line 254: When it is false, what happens? The value is simply discarded?

josevalim: I am a bit confused because it seems Rustler will raise on overflow-underflow if `strict: false`.

josevalim:
Hi @lkarthee, I have dropped some comments, but I am still not ahead we should go with this. My understanding is that Rustler will raise anyway, so will `strict: false` even work as we expect? Or Rustler will convert them to nil on underflow/overflow?

lkarthee:
@josevalim 
> My understanding is that Rustler will raise anyway, so will strict: false even work as we expect?

> I am a bit confused because it seems Rustler will raise on overflow-underflow if strict: false.

I was confused about rustler's behaviour - figured out that `argument error` occurs only when there is mismatch between iX(i8, i16, i32, i64) and uX(u8, u16, u32, u64). My assumption is some kind of typecasting is done by Rustler.

```elixir
iex(2)> Series.from_list([-129, 256], dtype: {:s, 8})
19:49:17.265 [info] dtype - {:s, 8} # <- Just to make sure we are passing before rustler is called 
type_name of val is alloc::vec::Vec<core::option::Option<i8>> # <- log from rust that it is using i8 
#Explorer.Series<
  Polars[2]
  s8 [127, 0]
>
iex(3)> Series.from_list([256], dtype: {:u, 8})
#Explorer.Series<
  Polars[1]
  u8 [0]
>
iex(4)> Series.from_list([-129, 256], dtype: {:u, 8})
** (ArgumentError) argument error
    (explorer 0.9.0-dev) Explorer.PolarsBackend.Native.s_from_list_u8("", [-129, 256])
    (explorer 0.9.0-dev) lib/explorer/polars_backend/series.ex:24: Explorer.PolarsBackend.Series.from_list/2
    (explorer 0.9.0-dev) lib/explorer/series.ex:417: Explorer.Series.from_list/2
```

Overflow of float 32 - results in crash  of VM: 
```elixir
iex(3)> Series.from_list([3.402823E+39])
#Explorer.Series<
  Polars[1]
  f64 [3.402823e39]
>
iex(4)> Series.from_list([3.402823E+39], dtype: {:f, 32})
thread '<unnamed>' panicked at src/series.rs:86:1:
    called `Result::unwrap()` on an `Err` value: {error, badarg} 
stack backtrace:iex(4)> Series.from_list([3.402823E+39], dtype: {:f, 32})
libunwind: stepWithCompactEncoding - invalid compact unwind encoding
Abort trap: 6
```

lib/explorer/series.ex
lkarthee: When false overflow happens. Discard only happens when a series is created as {:s, 64} and cast it to {:s, 8}. And cast does not happen with existing code with and without strict.
```
Series.from_list([-129, 256], dtype: {:s, 16}) |> Series.cast({:s, 8})
#Explorer.Series<
  Polars[2]
  s8 [nil, nil]
>
```
More details about rustler behaviour - https://github.com/elixir-explorer/explorer/pull/826#issuecomment-1902647024

josevalim:
Right, that’s why I think this pull request is a bit too early. We need to make sure that the default behavior can be not strict, nor seg fault the runtime. We need to understand why floating points crash and, if necessary, add casting on the Elixir side.

From the looks of it, strict: true may be the safest choice, specially as it happens by default. so my suggestion is to improve the messages and the segmentation fault before we add any new code or large change. What is the minimum work we can do to provide a good user experience with the current behavior?

lkarthee:
Alright lets into look into f32 crashes and overflow/underflow first before proceeding any further.

We can figure out rest as we go.

lkarthee:
float 32 crash is happening due to `Some(item.decode::<$type>().unwrap()), ` , it should be `match item.decode::<$type>() {...},` ?

This solves the crash - but overflow I don't know how to handle it? Any suggestions @philss @josevalim @billylanchantin 

```rust
macro_rules! from_list_float {
    ($name:ident, $type:ty, $module:ident) => {
        pub fn $name(name: &str, val: Term) -> ExSeries {
            let nan = atoms::nan();
            let infinity = atoms::infinity();
            let neg_infinity = atoms::neg_infinity();
            ExSeries::new(Series::new(
                name,
                val.decode::<ListIterator>()
                    .unwrap()
                    .map(|item| { 
                        match item.get_type() {
                        TermType::Number => match item.decode::<$type>() {Ok(f) => Some(f), Err(_) => None}, //Some(item.decode::<$type>().unwrap()), 
                        TermType::Atom => {
                            if nan.eq(&item) {
                                Some($module::NAN)
                            } else if infinity.eq(&item) {
                                Some($module::INFINITY)
                            } else if neg_infinity.eq(&item) {
                                Some($module::NEG_INFINITY)
                            } else {
                                None
                            }
                        }
                        term_type => panic!("from_list/2 not implemented for {term_type:?}"),
                    }})
                    .collect::<Vec<Option<$type>>>(),
            ))
        }
    };
}
```

josevalim:
I think crashing is fine, but we need to crash exactly in the same way as the other ones, so we can rescue it accordingly in the elixir layer

lkarthee:
https://github.com/elixir-explorer/explorer/pull/837 handles float issue.

What to do with integer overflow ?  https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow 

> When you’re compiling in release mode with the --release flag, Rust does not include checks for integer overflow that cause panics. Instead, if overflow occurs, Rust performs two’s complement wrapping. In short, values greater than the maximum value the type can hold “wrap around” to the minimum of the values the type can hold. In the case of a u8, the value 256 becomes 0, the value 257 becomes 1, and so on. The program won’t panic, but the variable will have a value that probably isn’t what you were expecting it to have. Relying on integer overflow’s wrapping behavior is considered an error.
> To explicitly handle the possibility of overflow, you can use these families of methods provided by the standard library for primitive numeric types:
>  - Wrap in all modes with the wrapping_* methods, such as wrapping_add.
>  - Return the None value if there is overflow with the checked_* methods.
>  - Return the value and a boolean indicating whether there was overflow with the overflowing_* methods.
>  - Saturate at the value’s minimum or maximum values with the saturating_* methods.

josevalim:
> Instead, if overflow occurs, Rust performs two’s complement wrapping.

This is exactly how Nx behaves. So I am glad to call it a day.

josevalim:
To be clear, Nx doesn't deal with overflows/underflows, they always happen across all operations, and I'd say it is fine to have the same behaviour here too. One of the main points of those libraries is to enable high performance, so the trade-off makes sense?

lkarthee:
I will :
- document overflow/underflow behaviour 
- wrap "Explorer.PolarsBackend.Shared.from_list" into a try/rescue ArgumentError and raise a better message there
- close this pr

PR #830, opened 2024-01-19
Fix functions with result/dtypes mismatch

philss:
This is a fix of some functions that returns something different from what Polars is calculating.

To find them, I had to enable the `:check_polars_frames` application env in `config/config.exs`.


josevalim (approved):
Ship it! Don't forget to leave the check polars config enabled, if the whole suite passes with it!

PR #841, opened 2024-01-29
Add Struct json_decode/3 for decoding json from string

lkarthee:
```elixir
iex(1)> require Explorer.DataFrame, as: DF
Explorer.DataFrame
iex(2)> d = DF.new([%{a: "{\"n\": 1}"}])
#Explorer.DataFrame<
  Polars[1 x 1]
  a string ["{\"n\": 1}"]
>
iex(3)> DF.mutate(d, aj: json_decode(a))
#Explorer.DataFrame<
  Polars[1 x 2]
  a string ["{\"n\": 1}"]
  aj struct[1] [%{"n" => 1}]
>
```


josevalim (commented):
What happens if the string is not valid JSON?

lkarthee:
Currently I get error unless I pass `dtype: {:struct, %{"n" => {:s, 64}}}` - due to out_df and check_df mismatch. When dtype is not passed, it is equivalent of from_list() scenario. 

Please advise on how to handle this: 
- Should we accept the `check_df.dtypes` returned by polars in case of `:df_mutate_with_exprs` ?
- Should we narrow down check  to don't worry about `{:struct, _}` for `out_df.dtypes != check_df.dtypes` and accept `check_df` if everything else matches ? 

```elixir 
** (RuntimeError) DataFrame mismatch.

expected:

    names: ["a", "aj"]
    dtypes: %{"a" => :string, "aj" => nil}

got:

    names: ["a", "aj"]
    dtypes: %{"a" => :string, "aj" => {:struct, %{"n" => {:s, 64}}}}

    (explorer 0.9.0-dev) lib/explorer/polars_backend/shared.ex:62: Explorer.PolarsBackend.Shared.apply_dataframe/4
    iex:3: (file)
```

https://github.com/elixir-explorer/explorer/blob/8b0372644a91c9f962cce83c73494f8d97bb60a1/lib/explorer/polars_backend/shared.ex#L42-L79

lkarthee:
> What happens if the string is not valid JSON?

Raises a polars error.
```elixir
iex(6)> d = DF.new([%{a: "{\"n\": 1"}])
iex(7)> DF.mutate(d, aj: json_decode(a, dtype: {:struct, %{"n" => {:s, 64}}}))
** (RuntimeError) Polars Error: error deserializing JSON: json parsing error: 'ExpectedObjectContent at character 8 (']')'
    (explorer 0.9.0-dev) lib/explorer/polars_backend/shared.ex:81: Explorer.PolarsBackend.Shared.apply_dataframe/4
    iex:7: (file)
```

josevalim:
I would start with the simplest API possible:

1. no options
2. document it errors on invalid json

native/explorer/Cargo.toml
```diff
@@ -36,6 +36,9 @@ object_store = { version = "0.8", default-features = false, optional = true }
 [target.'cfg(not(any(all(windows, target_env = "gnu"), all(target_os = "linux", target_env = "musl"))))'.dependencies]
 mimalloc = { version = "*", default-features = false }
 
+[patch.crates-io]
```
lkarthee, line 39: context here - https://github.com/pola-rs/polars/issues/14008

lib/explorer/polars_backend/shared.ex
```diff
@@ -42,36 +42,55 @@ defmodule Explorer.PolarsBackend.Shared do
...
+            # We need to collect here, because the lazy frame may not have
+            # the full picture of the result yet.
+            check_df =
+              if match?(%PolarsLazyFrame{}, new_df) do
+                {:ok, new_df} = Native.lf_collect(new_df)
+                create_dataframe(new_df)
+              else
+                create_dataframe(new_df)
+              end
+
```
lkarthee, line 56: DF.mutate for struct related ops are special case as we may not know dtype ahead of op. I am just accepting the returned dtype. I made sure that the exception is narrowed down to  :df_mutate_with_exprs and only happens for structs.

josevalim: Hrm, this can be a big deal and a blocker for us. :( We need to know the operations so we can build lazy queries, which make sense to be impossible in this case. So we'd need to require the dtypes.

Other options are:

1. Introduce `DF.from_json` and add a mechanism to convert a DF to a series (i.e. a struct) (either DF.to_structs or Series.from_dataframe) (this means we can support infer schema length and dtype options as the other conversion functions)

2. Add `Series.from_json`

I am more inclined to go with option 1 for now. Thoughts?

lkarthee: I think supporting option 2 is a low hanging fruit: 

- raise error when from_json is used as data frame operation
- rename `json_decode`  to `from_json`

```elixir
d = DF.new([%{a: "{\"n\": 1}"}])
aj = Series.from_json(df["a"])
DF.put(df, "aj", aj) 
```

lib/explorer/polars_backend/expression.ex
```diff
@@ -172,6 +173,11 @@ defmodule Explorer.PolarsBackend.Expression do
     raise ArgumentError, "missing #{inspect(__MODULE__)} nodes: #{inspect(missing)}"
   end
 
+  def to_expr(%LazySeries{op: :from_json, args: _}) do
```
lkarthee, line 176:
```suggestion
  def to_expr(%LazySeries{op: :from_json}) do
```

@josevalim If i drop `args: _` i get a lot of warnings even though all tests pass. Why does this happen ?
<details>

<summary>Log containing warnings</summary>
```
09:59 $ mix ci
    Finished dev [unoptimized + debuginfo] target(s) in 0.42s
Compiling 3 files (.ex)
     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

     warning: this clause cannot match because a previous clause at line 265 always matches
     │
 314 │     def to_expr(%LazySeries{op: unquote(op), args: unquote(args)}) do
     │     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ lib/explorer/polars_backend/expression.ex:314

Compiling crate explorer in debug mode (native/explorer)
   Compiling explorer v0.1.0 (/Users/kartheek/erlang-ws/github-ws/latest/explorer/native/explorer)
    Finished dev [unoptimized + debuginfo] target(s) in 5.26s
Excluding tags: [:cloud_integration]

.................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Finished in 2.8 seconds (2.8s async, 0.00s sync)
582 doctests, 1390 tests, 0 failures, 18 excluded, 1 skipped
</details>

**Comment by josevalim** [comment]
> I think supporting option 2 is a low hanging fruit:

Unfortunately I think this is inconsistent. All of our `from_*` operations work on data types outside of Series and DataFrame. So we would be breaking a rule here.

Therefore, if we introduce `from_json` (or `read_json`), the argument must be either a file or a string. Ideally we would support it in the dataframe too, because all of our `from_json` (or `read_json`) functions are defined in the dataframe. However, `from_json` supports files and URLs, so it is quite some work, but it would be consistent with everything else including polars itself: https://docs.pola.rs/py-polars/html/reference/api/polars.read_json.html

So, after sleeping on it, I think we should first add read_json/write_json, and then from_json and to_json. If it is too much work, let us know, and @philss can pick it up soon (he pretty much wrote all file integration stuff :D).

**Comment by lkarthee** [comment]
@josevalim I can contribute an hour a day for few more months. Now that I am familiar explorer/polars - it does not take much time for implementing things from polars. Maintainers can feel free to tag me or assign PRs/bugs to me - i will try my best to resolve them. 

I am confused about your comments and from_json - feel free to correct me.

### Json Decode 

`example_file.csv`
```
active,ask_code,ask_id,ask_price,code,data,expiry,id,index,inserted_at,item_id,level,project_id,properties,state,type,updated_at,user_id,value
True,,,,pull,{},1970-01-01T00:00:00Z,752e072c-9907-4fd0-9276-8fd30f7d1d1c,{},2023-02-24 15:41:57+00:00,e71a8d18-16c0-4efa-8442-dc17344ae5f8,1,890d0555-f266-42b2-a1ba-66925d8d47f6,"{'r_total': 0, 'u_total': 1, 'xp_total': 1}",ok,durable,2023-03-20 09:09:44+00:00,5755aa4d-d91e-4ab4-830e-791996f41a99,1
```
Description about data:
 -  json_decode is useful in scenario where data frame is already loaded from a non json file like from .csv
 -  Not all columns are json, rather some columns are json columns - rest are primitive types. 

Possible Operations:
 - convert those json columns to struct `pl.col('data').str.json_decode()` .
 - extract data from a json path using `pl.col('data').str.json_path_match('$. xp_total').alias('xp')` .

Problems with having `Series.json_decode` as DataFrame operation:
- we don't know `dtype` in advance, hence we can't use `DF.mutate` as `apply_dataframe` validates out_df and check_df.

Solution: 
- implement `Series.json_decode` as a series  operation which wont work as data frame operation/expression.
- raise error suggesting user to use `Series.json_decode/2` rather than `json_decode/2` in data frame mutate.

### `from_json` 
This is a complete different operation from `json_decode`:
-  `from_json` whole file is json 
-  `from_ndjson` every line is json
-  `json_decode` one or more variables/columns are json. there might be columns which are just primitives, example csv file 

`example.json`
```json
{
  "columns": [
    {"name": "integer",  "datatype": "Int64",  "bit_settings": "",  "values": [1,  2,  3 ]},
    {"name": "float", "datatype": "Float64", "bit_settings": "",  "values": [4, 5, 6 ]},
    {"name": "string",  "datatype": "String", "bit_settings": "", "values": ["a", "b", "c"] }
  ]
}
``` 

```python
df = pl.DataFrame(
    {
        "integer": [1, 2, 3],
        "float": [4.0, 5.0, 6.0],
        "string": ["a", "b", "c"],
    }
)

df.write_json("output.json")
df_json = pl.read_json("output.json")
print(df_json)
```

josevalim:
Thank you for the context. I see what you mean and I don't have a good answer for it.

The best I can think of is to call it "json_decode" but always expect the "schema"/"dtype" to be given, so we don't have to guess.

lkarthee:
One final question - can we have series only ops and combine it with DF.put ? That way user won't have to pass dtype every time.

lkarthee:
> We could but I would like to avoid "branching" for now, if that makes any sense.

Jose reply is to my question - whether we could have Series only op that way we could have avoid passing dtype?

I felt my question was partly in the previous reply and deleted it. Posting it again after I seen Jose's reply.

lkarthee:
I removed `infer_schema_length`, we don't need it as we are passing dtype.

lib/explorer/polars_backend/expression.ex
```diff
@@ -142,7 +142,8 @@ defmodule Explorer.PolarsBackend.Expression do
     member: 3,
 
     # Structs
-    field: 2
+    field: 2,
+    json_decode: 2
```
josevalim, line 146: It is really a string operation, not a struct one. :) It is about the input!

lib/explorer/backend/lazy_series.ex
```diff
@@ -143,7 +143,8 @@ defmodule Explorer.Backend.LazySeries do
     lengths: 1,
     member: 3,
     # Struct functions
-    field: 2
+    field: 2,
+    json_decode: 2
```
josevalim, line 147: It is really a string operation, not a struct one. :) It is about the input!

lib/explorer/series.ex
```diff
@@ -6043,6 +6043,28 @@ defmodule Explorer.Series do
...
+  ## Examples
+
+      iex> s = Series.from_list(["{\\"a\\":1}"])
+      iex> Series.json_decode(s, {:struct, %{"a" => {:s, 64}}})
+      #Explorer.Series<
+        Polars[1]
+        struct[1] [%{"a" => 1}]
+      >
+
+  Will raise `RuntimeError` for invalid json.
```
josevalim, line 6058: Should we have the dytpe also be a list in case the json is a list?

test/explorer/series_test.exs
```diff
@@ -5866,4 +5866,20 @@ defmodule Explorer.SeriesTest do
...
     end
   end
+
+  describe "json_decode/2" do
+    test "extracts struct from json" do
+      s = Series.from_list(["{\"n\": 1}"])
+      sj = Series.json_decode(s, {:struct, %{"n" => {:s, 64}}})
+      assert sj.dtype == {:struct, %{"n" => {:s, 64}}}
+      assert Series.to_list(sj) == [%{"n" => 1}]
+    end
```
josevalim, line 5876: These tests are redundant now.

test/explorer/series_test.exs
```diff
@@ -5866,4 +5866,20 @@ defmodule Explorer.SeriesTest do
...
+
+    test "extracts struct from json with dtype" do
+      s = Series.from_list(["{\"n\": 1}"])
+      sj = Series.json_decode(s, {:struct, %{"n" => {:f, 64}}})
+      assert sj.dtype == {:struct, %{"n" => {:f, 64}}}
+      assert Series.to_list(sj) == [%{"n" => 1.0}]
+    end
```
josevalim, line 5883: Let's add a test for the dtype not matching. Let's also add the same tests but inside a DF.mutate (i.e. lazy).

lib/explorer/series.ex
josevalim: Let's also document happens if the dtype does not match.

lkarthee: > Should we have the dytpe also be a list in case the json is a list?
Something like this ?
```elixir
s = Series.from_list(["[{\\"a\\":1}]"])
```

josevalim: Correct.

lib/explorer/series.ex
```diff
@@ -6043,6 +6043,28 @@ defmodule Explorer.Series do
...
+  """
+  @doc type: :struct_wise
```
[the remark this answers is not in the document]
josevalim: Yes!

test/explorer/series_test.exs
lkarthee: I just realised json_decode decodes to even primitives not just :struct and :list. It coerces to null for incompatible dtype. Error occurs when the string is not valid json.

```python
df = pl.DataFrame([{"a": "[\"a\"]"}])
df = df.with_columns(aj=pl.col("a").str.json_decode(dtype=pl.Int64,infer_schema_length=10), ajl=pl.col("a").str.json_decode(dtype=pl.List(pl.String),infer_schema_length=10))
shape: (1, 3)
┌───────┬──────┬───────────┐
│ a     ┆ aj   ┆ ajl       │
│ ---   ┆ ---  ┆ ---       │
│ str   ┆ i64  ┆ list[str] │
╞═══════╪══════╪═══════════╡
│ ["a"] ┆ null ┆ ["a"]     │
└───────┴──────┴───────────┘
```

josevalim: Excellent!

lib/explorer/series.ex
```diff
@@ -6043,6 +6043,46 @@ defmodule Explorer.Series do
...
+  Decoded value will be nil incase of a dtype mismatch e.g. string and integer.
+
+  ## Examples
+
+      iex> s = Series.from_list(["\\"1\\""])
+      iex> Series.json_decode(s, {:s, 64})
+      #Explorer.Series<
+        Polars[1]
+        s64 [nil]
+      >
```
josevalim, line 6076:
```suggestion
  Decodes a string series containing valid JSON according to `dtype`.

  ## Examples

      iex> s = Series.from_list(["1"])
      iex> Series.json_decode(s, {:s, 64})
      #Explorer.Series<
        Polars[1]
        s64 [1]
      >

      iex> s = Series.from_list(["{\\"a\\":1}"])
      iex> Series.json_decode(s, {:struct, %{"a" => {:s, 64}}})
      #Explorer.Series<
        Polars[1]
        struct[1] [%{"a" => 1}]
      >

  If the decoded value does not match the given `dtype`,
  nil is returned for the given entry:

      iex> s = Series.from_list(["\\"1\\""])
      iex> Series.json_decode(s, {:s, 64})
      #Explorer.Series<
        Polars[1]
        s64 [nil]
      >

  It raises an exception if the string is invalid JSON. 
```

josevalim:
Thank you! It was a bit bumpy but I am glad with where we arrived!

josevalim:

lkarthee:
> Thank you! It was a bit bumpy but I am glad with where we arrived!

Thank you Jose, I don't mind bumpy rides.

PR #843, opened 2024-01-31
[WIP] Adding support for deltalake file

### LINKED ISSUE #752: [Feature Request] Add delta lake file support
Python Polars has support for delta lake files already though I can't find the function exposed in the rust package.

cocoa-xu:
This PR tries to add support for deltalake file, and should close #752 once it's done.

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -79,6 +79,9 @@ defmodule Explorer.Backend.DataFrame do
               eol_delimiter :: option(String.t())
             ) :: io_result(df)
 
+  # IO: deltalake
+  @callback to_delta(df, table_uri :: fs_entry()) :: ok_result()
+
   # IO: Parquet
   @callback from_parquet(
               entry :: fs_entry(),
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1393,6 +1393,33 @@ defmodule Explorer.DataFrame do
   end
 
   @doc """
+  Writes a dataframe to a deltalake format file.
+  """
+  @doc tyoe: :io
+  @spec to_delta(df :: DataFrame.t(), table_uri :: String.t() | fs_entry()) :: :ok | {:error, Exception.t()}
+  def to_delta(df, table_uri) do
+    Shared.apply_impl(df, :to_delta, [table_uri])
+  end
+
+  @doc """
+  Similar to `to_delta/2`, but raises in case of error.
+  """
+  @doc type: :io
+  @spec to_delta!(df :: DataFrame.t(), table_uri :: String.t() | fs_entry()) :: :ok
+  def to_delta!(df, filename) do
+    case to_delta(df, filename) do
+      :ok ->
+        :ok
+
+      {:error, %module{} = e} when module in [ArgumentError, RuntimeError] ->
+        raise module, "to_delta failed: #{inspect(e.message)}"
+
+      {:error, error} ->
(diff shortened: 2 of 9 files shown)


philss:
hey @cocoa-xu , thank you for this PR!

I'm going to try to fix the missing parts if you don't mind. I'm going to open another branch to avoid conflicting with yours, but I'm going to be based on this branch, ok?

cocoa-xu:
> hey @cocoa-xu , thank you for this PR!
> 
> I'm going to try to fix the missing parts if you don't mind. I'm going to open another branch to avoid conflicting with yours, but I'm going to be based on this branch, ok?

No problem!

josevalim:
@cocoa-xu could we perhaps use a library that converts delta lake to an arrow stream, and then we can use the existing logic in `df_from_arrow_stream_pointer` to convert it to a dataframe?

Alternatively, we could even have delta lake as a separate library, and all it need to do is to expose a pointer to arrow stream, similar to what we do in Adbc. In Adbc we do:

      Adbc.Connection.query_pointer(conn, query, params, fn pointer, _num_rows ->
        Explorer.PolarsBackend.Native.df_from_arrow_stream_pointer(pointer)
      end)

But we could export `df_from_arrow_stream_pointer(pointer)` as a public function? This way you could do:

      DeltaLake.transfer_arrow_stream_pointer(lake, fn pointer ->
        Explorer.PolarsBackend.from_arrow_stream_pointer(pointer)
      end)

WDYT?

cocoa-xu:
> @cocoa-xu could we perhaps use a library that converts delta lake to an arrow stream, and then we can use the existing logic in df_from_arrow_stream_pointer to convert it to a dataframe?
>
> Alternatively, we could even have delta lake as a separate library, and all it need to do is to expose a pointer to arrow stream, similar to what we do in Adbc.

Yeah, I think perhaps I'll try to have delta lake as a separate library, at least it would be a good start point, and probably easier to debug (and possibly integrate it into explorer if it won't cause too much trouble to do so)

cocoa-xu:
So I read more in the deltalake repo and other relevant repos, and I found that we basically have to wait for the upstream to support ADBC.

According to [delta-rs issue #954](https://github.com/delta-io/delta-rs/issues/954) and [the design doc of Delta Lake ADBC](https://docs.google.com/document/d/1ud-iBPg8VVz2N3HxySz9qbrffw6a9I7TiGZJ2MBs7ZE/edit?usp=sharing),  

> We would aim to use this to support integration with DuckDB and Polars.

> Polars has a different problem: it uses arrow2, which is a separate Arrow implementation from apache/arrow-rs (used by DataFusion).

This was one of the difficulties I encountered here -- there's no easy way to cast from `polars_arrow`'s array to the one deltalake is using, i.e., `apache/arrow-rs`' array, or the other way around.

Currently, I think we should wait for 
- the Rust version of ADBC driver manager, which will be implemented either in 
    - https://github.com/apache/arrow-adbc/pull/416
    - https://github.com/apache/arrow-adbc/pull/1496
- ADBC connector for deltalake, https://github.com/delta-io/delta-rs/issues/954

Once they're available, we can easily add support for deltalake to explorer.

Besides that, we can try to replace current ADBC binding for Elixir https://github.com/elixir-explorer/adbc with the Rust version of ADBC if we want to and if it's possible/suitable for our use cases.

josevalim:
ADBC sounds like a solid route forward. Shall we close this one then?

cocoa-xu:
> ADBC sounds like a solid route forward. Shall we close this one then?

Yeah I think we can close this one and wait for the ADBC version for it.

PR #845, opened 2024-02-02
remove  deprecated funcs from docs

lkarthee:
Edit: Removed from docs 

Original: 
Deprecated function docs  do not have links - the names  are just displayed as text. 

https://hexdocs.pm/explorer/Explorer.DataFrame.html#functions-deprecated

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1622,7 +1622,7 @@ defmodule Explorer.DataFrame do
   def lazy(df), do: Shared.apply_impl(df, :lazy)
 
   @deprecated "Use lazy/1 instead"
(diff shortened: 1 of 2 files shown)


josevalim:
I'd rather `@doc false` them and not have them show up in the docs. Thoughts?

lkarthee:
Personally speaking, I would like to see deprecated functions show up in docs until they are removed. 

Do you think mentioning in changelog and compiler warnings are sufficient?

josevalim:
The idea is that only old code based call it and in their case the deprecation warning has the relevant into to migrate.

lkarthee:
removed deprecated from docs

Issue #847, opened 2024-02-06
Invalid Struct contents with `{:list, {:struct, %{...}}`


maennchen:
## Versions

Explorer: `0.8.0`
Elixir: `1.16.0`
OTP: `26.2.1`

## Reproduction

```elixir
dbg Series.from_list([[%{"a" => 1}]], dtype: {:list, {:struct, %{"a" => :integer}}})
# Series.from_list([[%{"a" => 1}]], dtype: {:list, {:struct, %{"a" => :integer}}}) #=> #Explorer.Series<
#   Polars[1]
#   list[struct[1]] [[%{"a" => 1}]]
# >
dbg Series.from_list([[], [%{"a" => 1}]], dtype: {:list, {:struct, %{"a" => :integer}}})
# Series.from_list([[], [%{"a" => 1}]], dtype: {:list, {:struct, %{"a" => :integer}}}) #=> #Explorer.Series<
#   Polars[2]
#   list[struct[1]] [[], [%{"" => nil}]]
# >
```

## Expected

The series with an empty list at the start should have `[%{"a" => 1}]` as the value for the second item and not ` [%{"" => nil}]`.

philss:
I'm investigating this. It looks like a problem when we try to create the series in the backend (Rust side). For some reason, the series seems to lose its data (or dtype). For what I could see, the data is arriving correctly when we pass down to Polars. I will try to reproduce it in an isolated project.

maennchen:
@philss Thanks for your fix!

Unfortunately while testing the newest `main`, I now sometimes get nif panics.

**Code**

```elixir
Mix.install([{:explorer, github: "elixir-explorer/explorer"}])

[
  [],
  [%{"PHONE_NUMBER" => "17023478521", "PHONE_TYPE" => "HQPHONE"}],
  [
    %{"PHONE_NUMBER" => "911212648151", "PHONE_TYPE" => "HQPHONE"},
    %{"PHONE_NUMBER" => "911212651803", "PHONE_TYPE" => "REGFAX"},
    %{"PHONE_NUMBER" => "911212648151", "PHONE_TYPE" => "REGPHONE"}
  ]
]
|> Series.from_list(dtype: {:list, {:struct, %{"PHONE_TYPE" => :string, "PHONE_NUMBER" => :string}}})
|> dbg
```

Running this script 100 times results in 73 panics and 27 successful runs on my machine.

**Ok Output**

```
[
  [],
[...]

PR #849, opened 2024-02-07
Fix `from_list/2` of list of structs when first is empty

philss:
Closes https://github.com/elixir-explorer/explorer/issues/847

### DIFF
--- a/lib/explorer/polars_backend/shared.ex
+++ b/lib/explorer/polars_backend/shared.ex
@@ -126,7 +126,7 @@ defmodule Explorer.PolarsBackend.Shared do
     Native.s_from_list_of_series(name, series)
   end
 
-  def from_list(list, {:struct, fields} = _dtype, name) when is_list(list) do
+  def from_list(list, {:struct, fields} = dtype, name) when is_list(list) do
     series =
       for {column, values} <- Table.to_columns(list) do
         column = to_string(column)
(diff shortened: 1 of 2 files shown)


test/explorer/series/list_test.exs
```diff
@@ -168,6 +168,22 @@ defmodule Explorer.Series.ListTest do
                    "the value \"z\" does not match the inferred dtype {:s, 64}",
                    fn -> Series.from_list([[[[[1, 2], ["z", "b"]]]]]) end
     end
+
+    test "list of structs" do
+      series =
+        Series.from_list([[%{"a" => 42}], []], dtype: {:list, {:struct, %{"a" => :integer}}})
+
+      assert Series.dtype(series) == {:list, {:struct, %{"a" => {:s, 64}}}}
+      assert Series.to_list(Series.cast(series, Series.dtype(series))) == [[%{"a" => 42}], []]
```
philss, line 177: ops
```suggestion
      assert Series.to_list(series) == [[%{"a" => 42}], []]
```

lib/explorer/polars_backend/shared.ex
```diff
@@ -135,6 +135,14 @@ defmodule Explorer.PolarsBackend.Shared do
...
 
     Native.s_from_list_of_series_as_structs(name, series)
+    |> then(fn polars_series ->
+      if Native.s_dtype(polars_series) != dtype do
+        {:ok, casted} = Native.s_cast(polars_series, dtype)
+        casted
+      else
+        polars_series
+      end
+    end)
```
philss, line 145: Just to give a context, the problem was because the "empty" inner struct series were getting a different dtype, because it couldn't be "inferred" from the series alone (since it was empty).
The cast will force those empty struct series to keep the same dtype.

PR #854, opened 2024-02-10
allow nils along with struct - from_list/2

### LINKED ISSUE #853: Allow nils when creating a struct series
Today we are not accepting `nil` when creating a series of structs.

```elixir
Explorer.Series.from_list([%{a: 42, b: "s"}, nil])
```

lkarthee:
Closes #853

Attempt at fixing the bug - need some input about atom keys.

### DIFF
--- a/lib/explorer/polars_backend/shared.ex
+++ b/lib/explorer/polars_backend/shared.ex
@@ -136,10 +136,16 @@ defmodule Explorer.PolarsBackend.Shared do
     columns = Map.new(fields, fn {k, _v} -> {k, []} end)
 
     columns =
-      Enum.reduce(list, columns, fn row, columns ->
-        Enum.reduce(row, columns, fn {field, value}, columns ->
-          Map.update!(columns, field, &[value | &1])
-        end)
+      Enum.reduce(list, columns, fn
+        nil, columns ->
+          Enum.reduce(fields, columns, fn {field, _}, columns ->
+            Map.update!(columns, field, &[nil | &1])
+          end)
+
+        row, columns ->
+          Enum.reduce(row, columns, fn {field, value}, columns ->
+            Map.update!(columns, field, &[value | &1])
+          end)
       end)
 
     series =
--- a/lib/explorer/polars_backend/shared.ex
+++ b/lib/explorer/polars_backend/shared.ex
@@ -133,7 +133,13 @@ defmodule Explorer.PolarsBackend.Shared do
   end
 
   def from_list(list, {:struct, fields}, name) when is_list(list) do
-    columns = Table.to_columns(list)
+    columns =
+      list
+      |> Enum.map(fn
+        nil -> Enum.map(fields, fn {k, _} -> {k, nil} end)
+        x -> x
+      end)
+      |> Table.to_columns()
 
     series =
       for {field, inner_dtype} <- fields do
--- a/test/explorer/series/struct_test.exs
+++ b/test/explorer/series/struct_test.exs
@@ -21,6 +21,17 @@ defmodule Explorer.Series.StructTest do
              ]
     end
 
+    test "allow nils" do
+      s = Series.from_list([nil, %{"a" => 1, "b" => 2}, nil])
+      assert s.dtype == {:struct, [{"a", {:s, 64}}, {"b", {:s, 64}}]}
+
+      assert Series.to_list(s) == [
+               %{"a" => nil, "b" => nil},
+               %{"a" => 1, "b" => 2},
+               %{"a" => nil, "b" => nil}
+             ]
+    end
+
     test "allows struct values" do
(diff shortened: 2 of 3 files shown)


lib/explorer/polars_backend/shared.ex
```diff
@@ -133,7 +133,13 @@ defmodule Explorer.PolarsBackend.Shared do
...
 
   def from_list(list, {:struct, fields}, name) when is_list(list) do
-    columns = Table.to_columns(list)
+    columns =
+      list
+      |> Enum.map(fn
+        nil -> Enum.map(fields, fn {k, _} -> {k, nil} end)
+        x -> x
+      end)
+      |> Table.to_columns()
```
lkarthee, line 143: Need input -  Should check if the first non nil entry has atom keys or string keys, 

As `k` is always a string - it will fail if input has atom keys. 
```elixir
Series.from_list([%{a: 1}, nil])


** (RuntimeError) key-value records must have columns in the same order, expected :a, but got "a"
    (table 0.1.2) lib/table/reader/enumerable.ex:100: Table.Reader.Enumerable.keyval_values/2

```

lib/explorer/polars_backend/shared.ex
```diff
@@ -133,7 +133,13 @@ defmodule Explorer.PolarsBackend.Shared do
...
```
lkarthee, line 142:
```suggestion
    atom_keys? =
      Enum.reduce_while(list, false, fn
        nil, acc ->
          {:cont, acc}

        [{k, _} | _], _ ->
          {:halt, is_atom(k)}
      end)

    columns =
      list
      |> Enum.map(fn
        nil ->
          if atom_keys? do
            Enum.map(fields, fn {k, _} -> {String.to_existing_atom(k), nil} end)
          else
            Enum.map(fields, fn {k, _} -> {k, nil} end)
          end

        x ->
          x
      end)
      |> Table.to_columns()
```

test/explorer/series/struct_test.exs
```diff
@@ -21,6 +21,17 @@ defmodule Explorer.Series.StructTest do
...
 
+    test "allow nils" do
+      s = Series.from_list([nil, %{"a" => 1, "b" => 2}, nil])
+      assert s.dtype == {:struct, [{"a", {:s, 64}}, {"b", {:s, 64}}]}
+
+      assert Series.to_list(s) == [
+               %{"a" => nil, "b" => nil},
+               %{"a" => 1, "b" => 2},
+               %{"a" => nil, "b" => nil}
+             ]
```
josevalim, line 32: We tried a solution similar to your but I think the issue is that this should be:

```suggestion
      assert Series.to_list(s) == [
               nil,
               %{"a" => 1, "b" => 2},
               nil
             ]
```

And I don't know how to solve this without changing how we are creating structs from the Rust side.

lkarthee: Interesting - my understanding is that struct just a cosmetic thing in polars as docs say: 

> Polars Structs are the idiomatic way of working with multiple columns. It is also a free operation i.e. moving columns into Structs does not copy any data!

> Polars expressions always have a Fn(Series) -> Series signature and Struct is thus the data type that allows us to provide multiple columns as input/ouput of an expression. In other words, all expressions have to return a Series object, and Struct allows us to stay consistent with that requirement.

I always thought None in struct list as 'broadcasting' (in my mental model).

```python
> s = pl.Series([None, {"a":1, "b": 2}, {"a": None, "b": None}])
>s.to_list()
[{'a': None, 'b': None}, {'a': 1, 'b': 2}, {'a': None, 'b': None}]
```

If we were to expect Series.to_list([nil, %{...},]) is that possible as we cannot be sure if it was nil or %{"a" => nil, "b" => nil} in the original data ?

Is this just one of those oneway operations ?

josevalim: I see, in that case your patch is correct. I will take a look at mixed keys soon and drop a suggestion.

lkarthee: Just to clarify - mixed keys always fails with Table.to_columns() and we can't do much about it. We are trying to support all atom keys or all string keys just like `Ecto.cast`.

lib/explorer/polars_backend/shared.ex
josevalim: @lkarthee I pushed a commit to main that simplifies this code and we don't need Table.to_columns anymore. Can you please rebase your PR? You should just need to handle nils in `cast_series` (renamed from `cast_numerics`) and add a similar clause here for handling nils:

```
        nil, columns ->
          Enum.reduce(fields, columns, fn {field, _}, columns ->
            Map.update!(columns, field, &[nil | &1])
          end)
```

josevalim:

PR #855, opened 2024-02-11
add support for map as expression

lkarthee:
Add `struct` equivalent

```elixir
iex(2)> df = DF.new(%{a: [1, nil, 3], b: ["a", "b", nil]})
#Explorer.DataFrame<
  Polars[3 x 2]
  a s64 [1, nil, 3]
  b string ["a", "b", nil]
>
iex(3)> DF.mutate(df, c: %{a: a, b: b, lit: 1, null: is_nil(a)})
#Explorer.DataFrame<
  Polars[3 x 3]
  a s64 [1, nil, 3]
  b string ["a", "b", nil]
  c struct[4] [
    %{"a" => 1, "b" => "a", "lit" => 1, "null" => false},
    %{"a" => nil, "b" => "b", "lit" => 1, ...},
    %{"a" => 3, "b" => nil, ...}
  ]
>
```

Note: Hiding original text as it is stale as per https://github.com/elixir-explorer/explorer/pull/855#issuecomment-1937523811
<details>
  <summary>Original text</summary>
Add `struct` expression.

```elixir 
df = DF.new(%{a: [1, 2, 3], b: ["a", "b", "c"]})
#Explorer.DataFrame<
  Polars[3 x 2]
  a s64 [1, 2, 3]
  b string ["a", "b", "c"]
>
DF.mutate(df, c: struct([a: a, b: b]))
#Explorer.DataFrame<
  Polars[3 x 3]
  a s64 [1, 2, 3]
  b string ["a", "b", "c"]
  c struct[2] [
    %{"a" => 1, "b" => "a"},
    %{"a" => 2, "b" => "b"},
    %{"a" => 3, "b" => "c"}
  ]
>

Explorer.Series.struct(a: df["a"], b: df["b"])
#Explorer.Series<
  Polars[3]
  struct[2] [
    %{"a" => 1, "b" => "a"},
    %{"a" => 2, "b" => "b"},
    %{"a" => 3, "b" => "c"}
  ]
>
```
</details>

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2800,63 +2800,7 @@ defmodule Explorer.DataFrame do
 
     result = fun.(ldf)
 
-    column_pairs =
-      to_column_pairs(df, result, fn value ->
-        case value do
-          %Series{data: %LazySeries{}} = lazy_series ->
-            lazy_series
-
-          %Series{data: _other} ->
-            raise ArgumentError,
-                  "expecting a lazy series. Consider using `Explorer.DataFrame.put/3` " <>
-                    "to add eager series to your dataframe."
-
-          list when is_list(list) ->
-            raise ArgumentError,
-                  "expecting a lazy series or scalar value, but instead got a list. " <>
-                    "consider using `Explorer.Series.from_list/2` to create a `Series`, " <>
-                    "and then `Explorer.DataFrame.put/3` to add the series to your dataframe."
-
-          nil ->
-            lazy_s = LazySeries.new(:lazy, [nil], :null)
-            Explorer.Backend.Series.new(lazy_s, :null)
-
-          number when is_number(number) ->
-            dtype = if is_integer(number), do: {:s, 64}, else: {:f, 64}
-            lazy_s = LazySeries.new(:lazy, [number], dtype)
-
-            Explorer.Backend.Series.new(lazy_s, dtype)
-
-          string when is_binary(string) ->
-            lazy_s = LazySeries.new(:lazy, [string], :string)
-
-            Explorer.Backend.Series.new(lazy_s, :string)
-
-          boolean when is_boolean(boolean) ->
-            lazy_s = LazySeries.new(:lazy, [boolean], :boolean)
-
-            Explorer.Backend.Series.new(lazy_s, :boolean)
-
-          date = %Date{} ->
-            lazy_s = LazySeries.new(:lazy, [date], :date)
-
-            Explorer.Backend.Series.new(lazy_s, :date)
-
-          datetime = %NaiveDateTime{} ->
-            lazy_s = LazySeries.new(:lazy, [datetime], {:datetime, :nanosecond})
-
-            Explorer.Backend.Series.new(lazy_s, {:datetime, :nanosecond})
-
-          duration = %Explorer.Duration{precision: precision} ->
-            lazy_s = LazySeries.new(:lazy, [duration], {:duration, precision})
-
-            Explorer.Backend.Series.new(lazy_s, {:duration, precision})
-
-          other ->
-            raise ArgumentError,
-                  "expecting a lazy series or scalar value, but instead got #{inspect(other)}"
-        end
-      end)
+    column_pairs = to_column_pairs(df, result, &value!/1)
 
     new_dtypes =
       for {column_name, series} <- column_pairs, into: %{} do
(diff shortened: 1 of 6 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -145,7 +145,8 @@ defmodule Explorer.Backend.LazySeries do
     lengths: 1,
```
lkarthee, line 145: will add tests after review..

lib/explorer/series.ex
```diff
@@ -6118,6 +6118,29 @@ defmodule Explorer.Series do
     apply_series(series, :json_decode, [dtype])
   end
 
+  @doc type: :string_wise
```
lkarthee, line 6121: - should we cast to the dtype which is passed ?
- can we infer dtype from `LazySeries`/  `Series` ?

josevalim:
Could we implement this without adding a `struct` function? Could we automatically convert maps to structs instead?

josevalim (approved):

lkarthee:
@josevalim currently %{}  works in `mutate` as a top-level expression. 

```elixir 
DF.mutate(df, c: %{a: a, b: b}) # works
DF.mutate(df, c: %{a: is_nil(a), b: is_nil(b)}) # works
DF.mutate(df, c: is_nil(%{a: a, b: b})) # fails
```
Can you suggest on how to tackle this ?

josevalim:
I am on my phone, but somewhere in lazy series we handle all literals, we should probably add map handling in there. The code will probably be pretty similar to the one you added to data frame, so we should probably find a way of sharing those as well.

josevalim:
I took a Quick Look and I was wrong. We only allow casting in specific operations in series.ex. For example, we could begin supporting maps in the comparison operators, if comparison is supported between structs. Outside of that, we most likely won’t support passing maps. There may be an argument we should allow literal (such as integers and maps) on is_nil, but that’s probably not the case today

lkarthee:
There are some convenient use cases of https://docs.pola.rs/user-guide/expressions/structs/#practical-use-cases-of-struct-columns .

Should we support passing struct to a series ? These would add value to mutating, filtering without mutating, etc

josevalim:
The question is: which operations should we support in on? For example, it doesn't make sense to support them on add or multiply. So I'd do operation per operation, at least initially.

lkarthee:
Ok, let me explore more on this question and come back later.

I think this PR is complete for now.

lib/explorer/data_frame.ex
```diff
@@ -2800,77 +2800,105 @@ defmodule Explorer.DataFrame do
...
-            Explorer.Backend.Series.new(lazy_s, :string)
+  defp value!(%Series{data: _other}) do
+    raise ArgumentError,
+          "expecting a lazy series. Consider using `Explorer.DataFrame.put/3` " <>
+            "to add eager series to your dataframe."
+  end
 
-          boolean when is_boolean(boolean) ->
-            lazy_s = LazySeries.new(:lazy, [boolean], :boolean)
+  defp value!(list) when is_list(list) do
```
lkarthee, line 2830: lists can be converted to map, name will be inferred if it is a plain column without op. For everything else, name should be specified i.e {name, Series}. 

```elixir
DF.mutate(df, c: [a, b])
#Explorer.DataFrame<
  Polars[3 x 3]
  a s64 [1, nil, 3]
  b string ["a", "b", nil]
  c struct[2] [
    %{"a" => 1, "b" => "a"},
    %{"a" => nil, "b" => "b"},
    %{"a" => 3, "b" => nil}
  ]
>

DF.mutate(df, c: [a, b, a_nil: is_nil(a)])
#Explorer.DataFrame<
  Polars[3 x 3]
  a s64 [1, nil, 3]
  b string ["a", "b", nil]
  c struct[3] [
    %{"a" => 1, "a_nil" => false, "b" => "a"},
    %{"a" => nil, "a_nil" => true, "b" => "b"},
    %{"a" => 3, "a_nil" => false, ...}
  ]
>
```

lkarthee: can revert it if you want.

lib/explorer/data_frame.ex
```diff
@@ -2800,77 +2800,105 @@ defmodule Explorer.DataFrame do
...
+        _, _ ->
+          raise ArgumentError,
+                "expecting list elements to be columns. Consider using %{a: a, ...} instead of list [a, ...]"
+      end)
 
-            Explorer.Backend.Series.new(lazy_s, :date)
+    if map do
+      value!(map)
+    else
+      raise ArgumentError,
```
lkarthee, line 2847: this can be deleted if we are going to keep lists  ?

lib/explorer/data_frame.ex
josevalim: Yes, let's please revert it, as we don't support this anywhere in Elixir and we can use the list syntax for creating literal lists as well.

test/explorer/data_frame_test.exs
```diff
@@ -691,6 +691,30 @@ defmodule Explorer.DataFrameTest do
              }
     end
 
+    test "with map " do
```
josevalim, line 694: Let's also add a separate test of maps with maps inside?

test/explorer/data_frame_test.exs
```diff
@@ -691,6 +691,30 @@ defmodule Explorer.DataFrameTest do
...
+      df = DF.new(%{a: [1, nil, 3], b: ["a", "b", nil]})
+      df1 = DF.mutate(df, c: %{a: a, b: b, lit: 1, null: is_nil(a)})
```
josevalim, line 696: We should also add a column which is a regular series. We actually supported this originally in the first commit but it was lost in the refactor.

lkarthee: Something like this ?
```
s = Series.from_list(["x", "y", "z"])
DF.mutate(df, c: s)
```

josevalim: Yes!

lkarthee: Can i look at this sometime later this week ?

josevalim: There is no rush! We can wait and merge it once you have time to look at it. :)

josevalim:

PR #856, opened 2024-02-11
Inferred dtype property tests

billylanchantin:
### Description

We've had a few issues with the inferred dtype logic (unsurprising, it's tricky!). I thought we may benefit from some property tests moving forward.

### Basic Idea

I've written a generator for dtypes and a generator for series of a specific dtype. I'm currently testing roughly the following property:

```
dtype |> series_of_dtype |> Series.dtype |> sub_dtype_of?(dtype)
```

where `sub_dtype_of?` deals with the fact that empty lists result in the `:null` dtype.

### First hit

I believe I've already found a bug with these tests. I've included it and the code which generated it as a demonstration.

### Considerations before merging

* More scalar types?
* Fix bugs before merging?
* How to deal with slow tests?


josevalim:
:+1:

> How to deal with slow tests?

Skip them by default (I think `exclude: [test: :property]` works) and then always include them on CI.

josevalim:
> Is the logic of sub_dtype_of?/2 correct? 

It looks so to me!

test/explorer/series/inferred_dtype_property_test.exs
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  defp dtype_generator do
+    scalar_dtype_generator = StreamData.constant({:s, 64})
+
+    # We don't need complicated keys: single letter strings should suffice.
+    key_generator = StreamData.string(?a..?z, min_length: 1, max_length: 1)
+
+    dtype_generator =
+      StreamData.tree(scalar_dtype_generator, fn generator ->
```
josevalim, line 37: It is up to you, but maybe StreamData should be imported. :)

josevalim (approved):
I am happy with the PR. Feel free to merge or ask others opinions. :)

test/explorer/series/inferred_dtype_property_test.exs
billylanchantin: Oh yeah that's more readable: [a78b283](https://github.com/elixir-explorer/explorer/pull/856/commits/a78b283acbc7bdf495e3ae13b287033a7cbe21db)

PR #833, opened 2024-01-20
Add DataFrame.with_row_count/2

iurimateus:
Do note that `with_row_count` was renamed in polars/main to `with_row_index` (ref https://github.com/pola-rs/polars/pull/13494), so perhaps we should wait for a new polars release?

PR #862, opened 2024-02-18
Series.row_index/1

iurimateus:
>  Interesting. So what if we:
> 
> 1. Add a `Series.row_index()`
> 2. Make it raise for the default backend saying it is only supported in dataframes
> 3. Implement it properly for expressions
> 
> Would that work? Perhaps we try that as a separate PR so we can compare them side by side?

_Originally posted by @josevalim in https://github.com/elixir-explorer/explorer/issues/833#issuecomment-1902213418_

@josevalim is this what you had in mind? I think it's the first `/0` function in Series; I'm also not sure on the implementation.

Should we rename `len()` to `size()` for consistency?

cc @cigrainger

### DIFF
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -236,6 +236,13 @@ defmodule Explorer.PolarsBackend.Series do
   @impl true
   def any?(series), do: Shared.apply_series(series, :s_any)
 
+  @impl true
+  def row_index(series) do
+    Explorer.DataFrame.new(series: series)
+    |> Explorer.DataFrame.mutate_with(&[row_index: Series.row_index(&1[:series])])
+    |> Explorer.DataFrame.pull(:row_index)
+  end
+
   # Cumulative
 
   @impl true
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -237,11 +237,7 @@ defmodule Explorer.PolarsBackend.Series do
   def any?(series), do: Shared.apply_series(series, :s_any)
 
   @impl true
-  def row_index(series) do
-    Explorer.DataFrame.new(series: series)
-    |> Explorer.DataFrame.mutate_with(&[row_index: Series.row_index(&1[:series])])
-    |> Explorer.DataFrame.pull(:row_index)
-  end
+  def row_index(series), do: Shared.apply_series(series, :s_row_index)
 
   # Cumulative
 
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -2909,6 +2909,44 @@ defmodule Explorer.Series do
   def any?(%Series{dtype: :boolean} = series), do: apply_series(series, :any?)
   def any?(%Series{dtype: dtype}), do: dtype_error("any?/1", dtype, [:boolean])
 
+  @doc """
+
+  Returns a series of N consecutive numbers, starting from 0.
+
+  ## Examples
+      iex> s = Series.from_list([nil, true, true])
+      iex> Series.row_index(s)
+      #Explorer.Series<
+        Polars[3]
+        u32 [0, 1, 2]
+      >
+
+  Can be used to add a row index as the first column in the DataFrame.
+  The resulting column does not have any special properties. It is a regular column of type
+  UInt32.
+
+      iex> require Explorer.DataFrame, as: DF
+      iex> df = DF.new(a: [1, 3, 5], b: [2, 4, 6])
+      iex> DF.mutate(df, index: row_index(a)) |> DF.relocate("index", before: 0)
+      #Explorer.DataFrame<
+        Polars[3 x 3]
+        index u32 [0, 1, 2]
+        a s64 [1, 3, 5]
+        b s64 [2, 4, 6]
+      >
+      iex> df = DF.new(a: [1, 3, 5], b: [2, 4, 6])
+      iex> DF.mutate(df, id: row_index(a) + 1000)
+      #Explorer.DataFrame<
+        Polars[3 x 3]
+        a s64 [1, 3, 5]
+        b s64 [2, 4, 6]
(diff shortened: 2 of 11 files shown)


josevalim:
Thank you @iurimateus. Quick question: do we need to use expr_len? couldn't we use `size(...)` and expect the series as argument for `row_index`?

billylanchantin:
@iurimateus I think `expr_len` is already present, it's just named `expr_size`:

https://github.com/elixir-explorer/explorer/blob/main/native/explorer/src/expressions.rs#L576-L580

iurimateus:
For the record, these changes were before we had `size`.

> @iurimateus I think `expr_len` is already present, it's just named `expr_size`:

Thanks for pointing it out

> Thank you @iurimateus. Quick question: do we need to use expr_len? couldn't we use `size(...)` and expect the series as argument for `row_index`?

That works, I've pushed a new commit.

josevalim:
I would prefer to postpone int_range for now. So let's remove it from the Elixir side but we can still keep it on the Rust side.

We also need to list `row_index` as a callback here: https://github.com/elixir-explorer/explorer/blob/b901d975c4dd6dc8fe462987afbd497649b958c8/lib/explorer/backend/series.ex#L41

And then provide an implementation for regular series (which will basically return a list of u32).

lib/explorer/series.ex
```diff
@@ -6024,6 +6024,14 @@ defmodule Explorer.Series do
   def member?(%Series{dtype: dtype}, _value),
     do: dtype_error("member?/2", dtype, [{:list, :_}])
 
+  def row_index(%Series{} = series) do
+    apply_series(series, :row_index)
+  end
+
+  def int_range(start, end_, step \\ 1) do
+    Explorer.Backend.LazySeries.int_range(start, end_, step)
+  end
```
josevalim, line 6033: Let's remove this one.

lib/explorer/series.ex
```diff
@@ -6024,6 +6024,14 @@ defmodule Explorer.Series do
   def member?(%Series{dtype: dtype}, _value),
     do: dtype_error("member?/2", dtype, [{:list, :_}])
 
+  def row_index(%Series{} = series) do
```
josevalim, line 6027: Let's add docs here. :)

lib/explorer/polars_backend/series.ex
```diff
@@ -236,6 +236,13 @@ defmodule Explorer.PolarsBackend.Series do
   @impl true
   def any?(series), do: Shared.apply_series(series, :s_any)
 
+  @impl true
```
iurimateus, line 239: is that ok?

lib/explorer/polars_backend/series.ex
```diff
@@ -236,6 +236,13 @@ defmodule Explorer.PolarsBackend.Series do
...
+  def row_index(series) do
```
iurimateus, line 240: is this ok?

lib/explorer/polars_backend/series.ex
```diff
@@ -236,6 +236,13 @@ defmodule Explorer.PolarsBackend.Series do
...
+    Explorer.DataFrame.new(series: series)
+    |> Explorer.DataFrame.mutate_with(&[row_index: Series.row_index(&1[:series])])
+    |> Explorer.DataFrame.pull(:row_index)
+  end
+
```
iurimateus, line 245: is this ok?

lib/explorer/polars_backend/series.ex
```diff
@@ -236,6 +236,13 @@ defmodule Explorer.PolarsBackend.Series do
   @impl true
   def any?(series), do: Shared.apply_series(series, :s_any)
 
+  @impl true
+  def row_index(series) do
+    Explorer.DataFrame.new(series: series)
+    |> Explorer.DataFrame.mutate_with(&[row_index: Series.row_index(&1[:series])])
+    |> Explorer.DataFrame.pull(:row_index)
```
josevalim, line 243: I wonder if we should expose this function in Rust instead? https://docs.rs/polars/latest/polars/datatypes/struct.UInt32Type.html#method.new 🤔 would it have any performance difference?

josevalim:
Awesome job. I have pushed one last comment and I think we are good to go!

lib/explorer/polars_backend/series.ex
iurimateus: And perhaps `new(lazy: true)`, but I'm not sure about allocations.

Thanks for the reference, I've pushed a commit exposing it in Rust.

lib/explorer/series.ex
```diff
@@ -2909,6 +2909,44 @@ defmodule Explorer.Series do
   def any?(%Series{dtype: :boolean} = series), do: apply_series(series, :any?)
   def any?(%Series{dtype: dtype}), do: dtype_error("any?/1", dtype, [:boolean])
 
+  @doc """
+
+  Returns a series of N consecutive numbers, starting from 0.
```
josevalim, line 2914:
```suggestion
  Returns a series of indexes for each item (row) in the series, starting from 0.
```

lib/explorer/series.ex
```diff
@@ -2909,6 +2909,44 @@ defmodule Explorer.Series do
...
+      iex> s = Series.from_list([nil, true, true])
+      iex> Series.row_index(s)
+      #Explorer.Series<
+        Polars[3]
+        u32 [0, 1, 2]
+      >
+
+  Can be used to add a row index as the first column in the DataFrame.
+  The resulting column does not have any special properties. It is a regular column of type
+  UInt32.
```
josevalim, line 2926:
```suggestion

      iex> s = Series.from_list([nil, true, true])
      iex> Series.row_index(s)
      #Explorer.Series<
        Polars[3]
        u32 [0, 1, 2]
      >

  This function can be used to add a row index as the first column of a dataframe.
  The resulting column is a regular column of type `:u32`.
```

josevalim:

PR #872, opened 2024-02-28
DataFrame.discard/2 - don't raise for non existent column 

### LINKED ISSUE #870: Should we always raise when a column is missing?
Feedback from a colleague: he's been tripping over `DataFrame.discard/2` and forced into patterns like `if field in DataFrame.names(df), do: DataFrame.discard(df, field), else: df`. That one feels like it should be a no-op for non-existing fields. I can see the counterargument though, so putting it out there before raising a PR.

lkarthee:
closes #870


cigrainger (approved):
I have a branch locally that I was going to push this morning that is almost this verbatim. Nice! 👍

lkarthee:
Sorry, should have asked before sending this pr.

cigrainger:
Don't be sorry! Happy to have it :)

PR #859, opened 2024-02-15
add json_path_match/2

lkarthee:
```elixir
s = Explorer.Series.from_list([~s({"a": 1}), ~s({"b": 1})])
#Explorer.Series<
  Polars[2]
  string ["{\"a\": 1}", "{\"b\": 1}"]
>
iex(3)> Explorer.Series.json_path_match(s, "$.a")
#Explorer.Series<
  Polars[2]
  string ["1", nil]
>
iex(4)> Explorer.Series.json_path_match(s, "$.")
** (RuntimeError) Polars Error: ComputeError(ErrString("error compiling JSONpath expression path error: \nEof\n"))
    (explorer 0.9.0-dev) lib/explorer/polars_backend/shared.ex:26: Explorer.PolarsBackend.Shared.apply_series/3
    iex:4: (file)
```

PR #873, opened 2024-03-03
Implements Series.split_into/3

### LINKED ISSUE #871: Split string column into multiple columns (feature request / use case)
Hi there, I have been using Explorer for some of my data cleaning / preparation, and I have stumbled upon a use case that I don't quite know how to accomplish in a performant fashion.  The datasets I'm working with can be quite large, and I'd prefer to stick to using a LazyFrame to accomplish the task.  To distill this down into the simplest possible scenario, say I'd like to split a "Name" column into "First Name" and "Last Name":

```
ID,Name
1,John Smith
2,Jane Doe
```

Would become:
```
ID,First Name,Last Name
1,John,Smith
2,Jane,Doe
```

Polars [provides a way to do this](https://docs.pola.rs/py-polars/html/reference/series/api/polars.Series.str.splitn.html), albeit not in a very clean fashion.  My next thought was to use a [mutate_with/2](https://hexdocs.pm/explorer/Explorer.DataFrame.html#mutate_with/2) along with a [replace/3](https://hexdocs.pm/explorer/Explorer.Series.html#replace/3), which would require me to perform the operation twice while also using a regex.

It does seem like based on the Explorer philosophy, having an "all-in-one" operation to split the column into multiple doesn't belong in the DataFrame API, but that does raise the question of supporting regex replacement / filtering to enable this type of operation more easily.  I know #353 is currently open, but I was hoping to bring my use case to the forefront to see what the maintainers think is the best path forward for a solution to this problem.  I'm also happy to contribute, as I was experimenting with adding a few other things in my fork of the library.

ryancurtin:
Implements the `Series.split_into/3` function, which closes #871.  String columns can be split into a struct with multiple fields.

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -127,6 +127,7 @@ defmodule Explorer.Backend.LazySeries do
     downcase: 1,
     substring: 3,
     split: 2,
+    split_into: 3,
     json_decode: 2,
     json_path_match: 2,
     # Float round
@@ -1051,6 +1052,13 @@ defmodule Explorer.Backend.LazySeries do
     data = new(:split, [lazy_series!(series), by], :string)
 
     Backend.Series.new(data, {:list, :string})
+  end
+
+  @impl true
+  def split_into(series, by, fields) do
+    data = new(:split_into, [lazy_series!(series), by, fields], :string)
+
+    Backend.Series.new(data, {:list, :struct})
   end
 
   @impl true
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -1058,7 +1058,7 @@ defmodule Explorer.Backend.LazySeries do
   def split_into(series, by, fields) do
     data = new(:split_into, [lazy_series!(series), by, fields], :string)
 
-    Backend.Series.new(data, {:list, :struct})
+    Backend.Series.new(data, {:struct, Enum.map(fields, &{&1, :string})})
   end
 
   @impl true
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -5651,6 +5651,14 @@ defmodule Explorer.Series do
   def split(%Series{dtype: dtype}, _by),
     do: dtype_error("split/2", dtype, [:string])
 
+  @doc type: :string_wise
+  @spec split_into(Series.t(), String.t(), list(String.t() | atom())) :: Series.t()
+  def split_into(%Series{dtype: :string} = series, by, fields) when is_binary(by),
+    do: apply_series(series, :split_into, [by, fields])
+
+  def split_into(%Series{dtype: dtype}, _by, _fields),
+    do: dtype_error("split_into/3", dtype, [:string])
+
   # Float
 
   @doc """
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -5651,6 +5651,23 @@ defmodule Explorer.Series do
   def split(%Series{dtype: dtype}, _by),
     do: dtype_error("split/2", dtype, [:string])
 
+  @doc """
+  Split a string Series into a struct with fields determined by the list of
+  field names provided.  The length of the field names list determines how
+  many fields the resulting struct will have.  If the string cannot be split
+  into that many separate strings, null values will be provided for the
+  remaining fields.
+
+  ## Examples
+
+      iex> s = Series.from_list(["Smith, John", "Jones, Jane"])
+      iex> Series.split_into(s, ", ", ["Last Name", "First Name"])
+      #Explorer.Series<
+        Polars[2]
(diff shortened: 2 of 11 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -1053,6 +1054,13 @@ defmodule Explorer.Backend.LazySeries do
     Backend.Series.new(data, {:list, :string})
   end
 
+  @impl true
+  def split_into(series, by, fields) do
+    data = new(:split_into, [lazy_series!(series), by, fields], :string)
+
+    Backend.Series.new(data, {:list, :struct})
```
josevalim, line 1061: I think this should be something like `{:struct, Enum.map(fields, &{&1, :string})}`

lib/explorer/series.ex
```diff
@@ -5651,6 +5651,31 @@ defmodule Explorer.Series do
   def split(%Series{dtype: dtype}, _by),
     do: dtype_error("split/2", dtype, [:string])
 
+  @doc """
+  Split a string Series into a struct with fields determined by the list of
+  field names provided.  The length of the field names list determines how
+  many fields the resulting struct will have.  If the string cannot be split
+  into that many separate strings, null values will be provided for the
+  remaining fields.
```
josevalim, line 5659:
```suggestion
  Split a string Series into a struct of string `fields`.
  
  The length of the field names list determines how many times the
  string will be split at most. If the string cannot be split into that
  many separate strings, null values will be provided for the
  remaining fields.
```

The first sentence in docs is always a short summary. :)

josevalim:
This is looking great so far @ryancurtin! The last step is to implement the feature for the lazy backend as well and add a test. The lazy backend is the one used inside DF.mutate and friends. You can see other PRs, such as this one, for examples: https://github.com/elixir-explorer/explorer/pull/859

ryancurtin:
@josevalim - Ok, I've re-requested review since I've now implemented the lazy backend and added some tests using a DataFrame.  Thank you again for your help!

native/explorer/src/series.rs
```diff
@@ -1567,6 +1567,26 @@ pub fn s_split(s1: ExSeries, by: &str) -> Result<ExSeries, ExplorerError> {
     Ok(ExSeries::new(s2))
 }
 
+#[rustler::nif(schedule = "DirtyCpu")]
```
lkarthee, line 1570: Series can be converted into lazy frame and the column/series can be manipulated using expression. That way we can avoid using ChunkedArrays.

```rust
let s2 = s1
        .clone_inner()
        .into_frame()
        .lazy()
        .select([col(s1.name())
            .str()
        // ops - mirroring expr_split_into
            .alias("res")])
        .collect()?
        .column("res")?
        .clone();
Ok(ExSeries::new(s2))
```

lib/explorer/series.ex
```diff
@@ -5651,6 +5651,32 @@ defmodule Explorer.Series do
...
+      >
+
+  """
+  @doc type: :string_wise
+  @spec split_into(Series.t(), String.t(), list(String.t() | atom())) :: Series.t()
+  def split_into(%Series{dtype: :string} = series, by, fields) when is_binary(by),
+    do: apply_series(series, :split_into, [by, fields])
+
+  def split_into(%Series{dtype: dtype}, _by, _fields),
+    do: dtype_error("split_into/3", dtype, [:string])
```
josevalim, line 5678:
```suggestion
  def split_into(%Series{dtype: :string} = series, by, [_ | _] = fields) when is_binary(by),
    do: apply_series(series, :split_into, [by, Enum.map(fields, &to_string/1)])

  def split_into(%Series{dtype: dtype}, by, [_ | _] = _fields)  when is_binary(by),
    do: dtype_error("split_into/3", dtype, [:string])
```

josevalim: Let's normalize fields to string before and make sure it is a non-empty list.

josevalim:
Three remaining comments and we will ship it :D

lib/explorer/series.ex
```diff
@@ -5651,6 +5651,32 @@ defmodule Explorer.Series do
   def split(%Series{dtype: dtype}, _by),
     do: dtype_error("split/2", dtype, [:string])
 
+  @doc """
+  Split a string Series into a struct of string `fields`.
+  
+  The length of the field names list determines how many times the
```
josevalim, line 5657:
```suggestion
  Split a string Series into a struct of string `fields`.

  The length of the field names list determines how many times the
```

lib/explorer/series.ex
ryancurtin-wk: @josevalim - The underlying polars `splitn` function takes a list of arbitrary length.  Do you think it makes sense to limit the explorer API to only use a list of two fields?  Alternatively, I could put a guard to ensure that `fields` is a list so that users can split the column however many times necessary.

lkarthee (approved):

josevalim:

PR #880, opened 2024-03-11
Add more options for cacertfile including custom file supplied by user

GenericJam:
The existing code was failing for me due to being behind corporate proxy.

I shamelessly borrowed most of this from here -https://github.com/elixir-cldr/cldr_utils/blob/2fcfefb729e1ff3fa06a4cb9d531846c2056bcf2/lib/cldr/http/http.ex#L330 - so if kipcole9 could get credit that would be great.

I haven't been able to verify it works due to other build issues but it has worked as far as I've been able to test it.

I'll keep trying to test it on my end and report back.

### DIFF
--- a/lib/explorer/fss/utils.ex
+++ b/lib/explorer/fss/utils.ex
@@ -161,7 +161,7 @@ defmodule Explorer.FSS.Utils do
     # Use secure options, see https://gist.github.com/jonatanklosko/5e20ca84127f6b31bbe3906498e1a1d7
     [
       verify: :verify_peer,
-      cacertfile: CAStore.file_path() |> String.to_charlist(),
+      cacertfile: certificate_store(),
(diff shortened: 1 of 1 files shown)


GenericJam:
This works on my own machine which is not behind a proxy. I'd say it's safe to use.
It's failing on my work machine which is behind a proxy due to rust build issues.

GenericJam:
For me, it's more about trying to delegate the location of the certs than about finding them automatically. On my work computer, the company controls most of it apart from what's in `~/code` so it's better for me if I can tell it where the certs are.

GenericJam:
For me, the very janky work around involved copy/pasting certs into the _build folder so Explorer would find them.

GenericJam:
Also a lot of corporate code is stuck on older versions of Erlang/Elixir so new fixes won't make it there for a while.

Issue #498, opened 2023-02-08
Support operations with groups in the Lazy Polars backend


philss:
The lazy frame backend that we implemented in #227 does not take into account groups.
This issue is to track the implementation of groups for this backend.

### Single table verbs

* [x] `group`
* [x] `ungroup`
* [x] `mutate`
* [x] `arrange`
* [x] `filter`
* [x] `pivot_longer` (maybe this won't change the backend)

### Row-based verbs

* [x] `head`
* [x] `tail`
* [x] `slice`

cigrainger:
@philss what's the status on this one? I'm keen to contribute a bit more actively again.

philss:
@cigrainger hey Chris! Sorry for not answering before. I was mostly offline today.

I couldn't figure out how to perform those operations on Lazyframes, because Polars represents grouped lazy frames in a different way, with a different struct - https://docs.rs/polars/latest/polars/prelude/struct.LazyFrame.html#method.group_by. So if we want to apply operations in groups, we would have to group + agg for every operation, and I don't know how this would work if we want to "ungroup" a lazy frame.

I don't remember the full picture, but I think the problem was related to "when" to apply the operations, since they couldn't - AFAIK - be "reverted"/ungrouped in the lazy frame.

josevalim:
If I remember correctly, Polars does not support all of our lazy grouped operations in their lazy backend. So we either need to submit PRs to Polars (very hard) or change our implementation (supposedly easier).

Today the lazy backend simply dispatch to Polars. The idea is to rewrite it to have it collect the operations on the Elixir side. Then when we collect, we call the Polars API and immediately collect it too.

So I think the first step is to keep the current features but rewrite the lazy backend to build a list of operations on Elixir land. Then we can implement group/ungroup without major changes.

This is from memory, I am not 100% confident.

cigrainger:
I think that's exactly where we ended up and it doesn't seem too crazy. How would we represent the accumulated ops, a list of MFAs? 😱

josevalim:
Yes, a list as a stack. :)

philss:
This was completed by #890 

There is one operation remaining that is `sample/3`. I think it is possible to implement that when using groups, but I'm not sure of that yet. So I should open a new issue in case I think it's feasible.

PR #882, opened 2024-03-16
Rewrite LazyFrame

philss:
This is a new way to perform lazy frame operations.

The idea is to apply all operations to the Rust LazyFrame just when collecting. And we are storing each operation in a "stack".

Related to: https://github.com/elixir-explorer/explorer/issues/498


lib/explorer/backend/data_frame.ex
```diff
@@ -291,6 +291,24 @@ defmodule Explorer.Backend.DataFrame do
     end
   end
 
+  # TODO: revisit if this is the best approach
```
josevalim, line 294: Looks good to me!

lib/explorer/polars_backend/data_frame.ex
```diff
@@ -498,7 +498,18 @@ defmodule Explorer.PolarsBackend.DataFrame do
...
 
   @impl true
-  def lazy(df), do: Shared.apply_dataframe(df, df, :df_lazy, [])
+  def lazy(df) do
+    case Native.df_lazy(df.data) do
+      {:ok, polars_df} ->
+        %{
+          df
+          | data: struct!(lazy(), polars_lazy_frame: polars_df)
+        }
```
josevalim, line 507:
```suggestion
        %{df | data: struct!(lazy(), polars_lazy_frame: polars_df)}
```

:P

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -7,16 +7,30 @@ defmodule Explorer.PolarsBackend.LazyFrame do
   alias Explorer.PolarsBackend.Native
   alias Explorer.PolarsBackend.Shared
   alias Explorer.PolarsBackend.DataFrame, as: Eager
-  alias Explorer.PolarsBackend.LazyFrame, as: PolarsLazyFrame
+  # alias Explorer.PolarsBackend.LazyFrame, as: PolarsLazyFrame
+  alias Explorer.PolarsBackend.LazyStack
```
josevalim, line 11: I would probably keep those functions in this module. :)

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -7,16 +7,30 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
   import Explorer.PolarsBackend.Expression, only: [to_expr: 1, alias_expr: 2]
 
-  defstruct resource: nil
+  defstruct polars_lazy_frame: nil, stack: []
 
-  @type t :: %__MODULE__{resource: reference()}
+  defmodule PolarsLazyFrame do
+    @moduledoc false
+
+    # This struct represents the lazy frame from the initialization
```
josevalim, line 23: Quick question: do we need to create a struct? Couldn't we just pass the reference around? Or does Rustler works better if we are passing structs?

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -7,16 +7,30 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
+  # alias Explorer.PolarsBackend.LazyFrame, as: PolarsLazyFrame
+  alias Explorer.PolarsBackend.LazyStack
 
   alias FSS.Local
   alias FSS.S3
 
   import Explorer.PolarsBackend.Expression, only: [to_expr: 1, alias_expr: 2]
 
-  defstruct resource: nil
+  defstruct polars_lazy_frame: nil, stack: []
```
josevalim, line 18: We can probably call it `frame` or something.

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -29,7 +43,64 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
+    # TODO: handle errors
+    {:ok, polars_df} = Native.lf_collect(result)
+
+    # NOTE: should we keep the groups like we do today?
+    %{ldf | data: polars_df}
+  end
+
+  defp apply_operations(ldf) do
+    ldf
+    |> LazyStack.stack_by_groups()
```
josevalim, line 58: This looks great to me! My initial suggestion is to first merge a version where we don't call `stack_by_groups`. Basically all operations become stack operations. So this:

```elixir
  def head(ldf, rows), do: Shared.apply_dataframe(ldf, ldf, :lf_head, [rows])
```

Becomes this:

```
push_operation(ldf, {:lf_head, [rows]})
```

And now, the collect is a `Enum.reduce/3` loop that converts the stack into `Shared.apply_dataframe` calls. This will solve the first part of the problem, which is to represent the operation chain in a consistent way. The next step is chunking, but then I think we can do it one by one, operation per operation.

WDYT?

lib/explorer/polars_backend/lazy_frame.ex
philss: I think we can implement the a custom type with an encoder/decoder, or use one of the other `Nif*` macros. I think I will keep it this way for the PR and try to solve this in another opportunity. WDYT?

lib/explorer/polars_backend/lazy_frame.ex
philss: Makes sense! So we just ignore the groups when stacking, and we use the groups when collecting only, right?

lib/explorer/polars_backend/lazy_frame.ex
josevalim: We can do it later, yeah.

philss:
@josevalim I think this is good for a second look :)

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -29,7 +40,45 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
+    |> stack()
+    |> Enum.reduce(ldf.data.frame, fn {operation, args}, polars_ldf ->
+      args = normalise_operation_args(operation, args, polars_ldf)
+
+      case apply(Native, operation, args) do
+        {:ok, polars_ldf} ->
+          polars_ldf
+
+        {:error, error} when is_binary(error) ->
+          RuntimeError.exception(error)
```
josevalim, line 69: I think you need to halt if you have an exception. Maybe use `reduce_while`?

josevalim: Or maybe raise? since `collect`matches on `{:ok, lazy_df}` anyway?

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -7,16 +7,27 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
   alias Explorer.PolarsBackend.DataFrame, as: Eager
-  alias Explorer.PolarsBackend.LazyFrame, as: PolarsLazyFrame
 
   alias FSS.Local
   alias FSS.S3
 
   import Explorer.PolarsBackend.Expression, only: [to_expr: 1, alias_expr: 2]
 
-  defstruct resource: nil
+  defstruct frame: nil, stack: []
```
josevalim, line 16: I wonder if there is a way to keep it as `defstruct resource: nil, stack: []` and we tell Rustler to not serialize or bother with the `stack: []` field?

lib/explorer/polars_backend/shared.ex
```diff
@@ -101,7 +101,7 @@ defmodule Explorer.PolarsBackend.Shared do
   end
 
   defp df_names(%PolarsLazyFrame{} = polars_df) do
-    {:ok, names} = Native.lf_names(polars_df)
+    {:ok, names} = Native.lf_names(polars_df.frame)
```
josevalim, line 104: Nitpick: I would pattern match on `frame` too.

lib/explorer/polars_backend/shared.ex
josevalim, line 1: Should we remove PolarsLazyFrame from this module attribute?

https://github.com/elixir-explorer/explorer/blob/3e99aa7af8d7ab01c26797a74497680aeffba793/lib/explorer/polars_backend/shared.ex#L12

Or is still used in some cases?

test/explorer/data_frame/lazy_test.exs
```diff
@@ -66,19 +66,11 @@ defmodule Explorer.DataFrame.LazyTest do
...
-  cement s64 [5, 177, 2598, 0, 204]
-  gas_flaring s64 [0, 0, 2623, 0, 3697]
-  per_capita f64 [0.08, 0.43, 0.9, 1.68, 0.37]
-  bunker_fuels s64 [9, 7, 663, 0, 321]
->)
+    new_ldf = ldf |> DF.slice(0, 5)
+    assert new_ldf |> DF.names() == DF.names(ldf)
+
+    # Should we test the inner data?
+    assert [{:lf_slice, [0, 5]}] = new_ldf.data.stack
```
josevalim, line 73: I don't think we should. We should collect it then.

josevalim (approved):
Looks great to me. Only some minor comments added.

lib/explorer/polars_backend/shared.ex
philss: It was in use in the "check dataframe dtypes" thing. We may need to write another version of that for the lazy frame now.

lib/explorer/polars_backend/lazy_frame.ex
philss: I don't think it's possible to do easily today. I will add this to my list and revisit in the near future.

josevalim (approved):

PR #887, opened 2024-03-22
Update Polars to v0.38.3

philss:
We got some minor changes, such as precision diffs and `window_standard_deviation` starting with `nil` (without weights).


philss:
The precision in the examples are diverging with my machine (works on my machine :P). I will check tomorrow :)

josevalim (approved):

PR #889, opened 2024-03-26
Implements `:streaming` option for DataFrame.to_csv/3

### LINKED ISSUE #888: Add :streaming option to DataFrame.to_csv/3
I have a use case where I would like to stream a large lazy `DataFrame` to a CSV without having to load the entire dataset in memory, and I was looking to implement a streaming option to the `to_csv/3` function, similar to `DataFrame.to_parquet/3`.  

I've already been able to make this work in a fork using the Polars [sink_csv](https://docs.rs/polars/latest/polars/prelude/struct.LazyFrame.html#method.sink_csv) function under the hood.  Please let me know what you think and I will submit my PR.  Thanks!

ryancurtin:
Closes #888 

### Description

I have added a streaming option for the `to_csv/3` that uses the [Polars `sink_csv` function](https://docs.pola.rs/docs/rust/dev/polars_lazy/frame/struct.LazyFrame.html#method.sink_csv) under the hood.  Polars does not have a `sink_csv_cloud` function similar to `sink_parquet_cloud`, so for now, I implemented it only for a LazyFrame using a local entry.

### Notes

As far as my implementation goes, I've kept the DataFrame API clean since I'm really just adding an option, but it does beg the question of how we'd want to expose some other [Polars serialization options](https://docs.pola.rs/docs/rust/dev/polars_lazy/prelude/struct.CsvWriterOptions.html) for users of the `to_csv` function.  For now I won't worry about that, though.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1329,15 +1329,15 @@ defmodule Explorer.DataFrame do
 
     * `:streaming` - Tells the backend if it should use streaming, which means
       that the dataframe is not loaded to the memory at once, and instead it is
-      written in chunks from a lazy dataframe.  This option is not supported when using an S3
-      entry.
+      written in chunks from a lazy dataframe.  Defaults to true on supported filesystems,
+      ignored on all others.
 
   """
   @doc type: :io
   @spec to_csv(df :: DataFrame.t(), filename :: fs_entry() | String.t(), opts :: Keyword.t()) ::
           :ok | {:error, Exception.t()}
   def to_csv(df, filename, opts \\ []) do
-    opts = Keyword.validate!(opts, header: true, delimiter: ",", streaming: false, config: nil)
+    opts = Keyword.validate!(opts, header: true, delimiter: ",", streaming: true, config: nil)
 
     with {:ok, entry} <- normalise_entry(filename, opts[:config]) do
       Shared.apply_impl(df, :to_csv, [entry, opts[:header], opts[:delimiter], opts[:streaming]])
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1327,15 +1327,20 @@ defmodule Explorer.DataFrame do
     * `:config` - An optional struct, keyword list or map, normally associated with remote
       file systems. See [IO section](#module-io-operations) for more details. (default: `nil`)
 
+    * `:streaming` - Tells the backend if it should use streaming, which means
+      that the dataframe is not loaded to the memory at once, and instead it is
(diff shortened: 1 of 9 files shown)


philss (approved):
Nice work! 🚀 

Regarding the options Polars provides, I think we would could expose then in a option that accepts a keyword list, like we do with `:config`. We could pass through, and let the backend validate it.

lib/explorer/data_frame.ex
```diff
@@ -1327,15 +1327,20 @@ defmodule Explorer.DataFrame do
...
+      written in chunks from a lazy dataframe.  This option is not supported when using an S3
+      entry.
+
   """
   @doc type: :io
   @spec to_csv(df :: DataFrame.t(), filename :: fs_entry() | String.t(), opts :: Keyword.t()) ::
           :ok | {:error, Exception.t()}
   def to_csv(df, filename, opts \\ []) do
-    opts = Keyword.validate!(opts, header: true, delimiter: ",", config: nil)
+    opts = Keyword.validate!(opts, header: true, delimiter: ",", streaming: false, config: nil)
```
josevalim, line 1340: We should probably default to true as in the other operations.

ryancurtin: @josevalim - Will do!  Do you think it's worth mentioning in the docs that streaming set to true for a DataFrame with an S3 entry will fail explicitly?  I could also design it so that the streaming option is ignored for a LazyFrame with an S3 entry, but for now I used the same behavior as `to_parquet/3` prior to having streaming implemented.

josevalim: @ryancurtin the most important is to default to true, so I would say "defaults to true on supported filesystems, ignored on all others".

ryancurtin: Sounds good, I just added that and tweaked the behavior so as not to raise an error when using the streaming option on an S3 filesystem given that `true` is now the default.  Tests are also updated to reflect that and try both cases.

lib/explorer/data_frame.ex
```diff
@@ -1329,15 +1329,15 @@ defmodule Explorer.DataFrame do
 
     * `:streaming` - Tells the backend if it should use streaming, which means
       that the dataframe is not loaded to the memory at once, and instead it is
-      written in chunks from a lazy dataframe.  This option is not supported when using an S3
-      entry.
+      written in chunks from a lazy dataframe.  Defaults to true on supported filesystems,
+      ignored on all others.
```
josevalim, line 1333:
```suggestion
      written in chunks from a lazy dataframe.  Defaults to true for local filesystems,
      ignored on all others.
```

josevalim (approved):

philss:
@ryancurtin thank you!

PR #890, opened 2024-04-01
Support operations with groups in the Lazy Polars backend

philss:
This is a fix for https://github.com/elixir-explorer/explorer/issues/498

Most of the operations that we had and should work with groups are now supported. The only remaining is the `sample/3`, which may add to this PR soon.

**EDIT**: `sample/3` is giving me an error that I can't understand. I will merge this without that function. I should revisit later.


josevalim (approved):

philss:
I'm merging this without the `sample/3` because of an error. I should go back to that function later.

PR #894, opened 2024-04-14
Add `re_contains/2` and `re_replace/3` to match with a regex

philss:
The regular expression must follow the `regex` crate rules, and should be a string.

The idea of keeping them separate is because we cannot use the Elixir standard regex, and we need to use strings for both literal and regex versions. To make the difference explicit, we are using the `re_` prefix for functions that accept regexes as strings.

This is related to https://github.com/elixir-explorer/explorer/issues/353


lib/explorer/series.ex
```diff
@@ -5355,12 +5359,46 @@ defmodule Explorer.Series do
...
+  @doc """
+  Detects whether a string matches a pattern.
+
+  > ### Notice {: .warning}
+  >
+  > This function matches against a regular expression. It does not expect an Elixir regex, but
+  > a escaped string - you can use the `~S` sigil for escaping - that follows the [`regex`](https://docs.rs/regex/latest/regex/)
+  > Rust crate rules. This is because our backend, Polars, expects that format.
+  >
+  > To match literal strings, you can use `contains/2`.
```
josevalim, line 5377:
```suggestion
  > This function matches against a regular expression. It does not expect an Elixir regex,
  > but a escaped string and you can use the `~S` sigil for escaping it. Since each Explorer
  > backend may have its own regular expression rules, you must consult their underlying
  > engine. For the default backend (Polars), the rules are outlined in the Rust create named
  > [`regex`](https://docs.rs/regex/latest/regex/).
  >
  > To match literal strings, you can use `contains/2`.
```

josevalim: Please mirror it below and ship it :D

josevalim (approved):

PR #895, opened 2024-04-15
Add `count_matches/2`, `re_count_matches/2`, `re_scan/2` and `re_named_captures/2` to `Series`

### LINKED ISSUE #353: Regex support?

philss:
This change is a step to support more operations with regexes.

The only detail here is that we cannot support `re_named_captures/2` inside the context of a query, because we cannot extract the groups for the regex string, since that regex is following the rules of the backend (at the point of `LazySeries`, we don't have access to the backend).

Closes #353


josevalim (approved):

lib/explorer/backend/lazy_series.ex
```diff
@@ -1151,6 +1155,43 @@ defmodule Explorer.Backend.LazySeries do
...
+
+    If you want to capture named groups from a column, you must do so outside of a query.
+    For example, instead of:
+
+        Explorer.DataFrame.mutate(df, new_column: re_named_captures(column, ~S/(a|b)/))
+
+    You must write:
+
+        Explorer.DataFrame.put(df, :new_column, Explorer.Series.re_named_captures(column, ~S/(a|b)/))
+    """
```
josevalim, line 1192: We could probably make this work by storing the original backend as a field of the the LazySeries and then asking it to return the names. But we can do this in another PR if desired, no worries for now IMO.

philss: I don't know yet how to capture that information in the context of the lazy frame, but I will research a bit. For now I'm going to merge this. Thanks!

PR #896, opened 2024-04-17
Pass down backend to lazy series and enable `re_named_captures/2` usage

philss:
This is a way to enable the usage of backend functions for series inside the context of `Query` (in the LazySeries implementation). With that, we can use `re_named_captures/2` in the functions that are using lazy series.

This is related to the following discussion: https://github.com/elixir-explorer/explorer/pull/895#discussion_r1566826282


lib/explorer/backend/lazy_frame.ex
```diff
@@ -21,7 +22,7 @@ defmodule Explorer.Backend.LazyFrame do
   @doc false
   def new(df) do
     Explorer.Backend.DataFrame.new(
-      %__MODULE__{names: df.names, dtypes: df.dtypes},
+      %__MODULE__{names: df.names, dtypes: df.dtypes, original_data: df.data},
```
josevalim, line 25: maybe we should store only the backend?

lib/explorer/backend/lazy_frame.ex
```diff
@@ -73,10 +74,21 @@ defmodule Explorer.Backend.LazyFrame do
...
+    data = %{data | backend: series_backend}
+
     Backend.Series.new(data, dtype_for_column)
   end
 
+  defp get_series_backend(%module{}) do
+    module
+    |> Module.split()
+    |> List.replace_at(-1, "Series")
+    |> Module.concat()
```
josevalim, line 89: I am not sure we can assume they have related names. So I would pass the dataframe backend instead and make it a dataframe callback?

lib/explorer/backend/lazy_frame.ex
```diff
@@ -73,10 +74,21 @@ defmodule Explorer.Backend.LazyFrame do
   @impl true
   def pull(df, column) do
     dtype_for_column = df.dtypes[column]
+    series_backend = get_series_backend(df.data.original_data)
+
     data = LazySeries.new(:column, [column], dtype_for_column)
```
josevalim, line 79: And we should probably pass the backend as a required argument here. We need to make sure we always pass it forward, otherwise something like `re_named_captures(s.foo <> s.bar, ~S"...")` won't work if we forget to pass the backend inside? Would this be too large of a change?

lib/explorer/backend/lazy_frame.ex
philss: done in 3a8d825716b5b0ee1646388d99ca0565939557e8

lib/explorer/backend/lazy_frame.ex
philss: Yeah, makes sense!

lib/explorer/backend/lazy_frame.ex
philss: I'm passing down based on the `args` - see the [`new/3` function at LazySeries](https://github.com/elixir-explorer/explorer/pull/896/files#diff-f91b78e498841f6f17ff64be6801aab96764b8d998a5d69c6b397da4d20de35fR208) -, which will mostly contain a lazy series. The only exceptions are `from_list/2` and `from_binary/2`. I honestly don't know how to solve this, because there is no point of reference to get the backend. I was thinking about getting the default backend for them. WDYT?

philss: Using the default in c708f46

lib/explorer/backend/lazy_series.ex
```diff
@@ -1177,19 +1194,28 @@ defmodule Explorer.Backend.LazySeries do
...
+  defp get_backend(%__MODULE__{} = lazy_series) do
+    lazy_series.backend || warning_with_default_backend()
+  end
+
+  defp warning_with_default_backend do
+    backend = Explorer.Backend.get()
+
+    Logger.warning(
+      "cannot get backend from LazySeries. Using the default one: #{inspect(backend)}"
+    )
```
josevalim, line 1216: When would this be the case?

philss: For this function I couldn't find a way to reproduce the exception. But if a series operation that accepts scalar values for all its arguments is called, it will call "from_list", and it will raise here. But this is unlikely to happen IMHO.

philss: This log*

josevalim: What if we raise instead then?

philss: Yeah, it makes sense! I will update it later today :)

philss: 1ae1ddd

lib/explorer/backend/lazy_series.ex
```diff
@@ -197,8 +203,18 @@ defmodule Explorer.Backend.LazySeries do
...
+  def new(op, args, dtype, aggregation \\ false, backend \\ nil) do
+    dtype = Explorer.Shared.normalise_dtype!(dtype)
+    backend = backend || backend_from_args(args)
+
+    %__MODULE__{op: op, args: args, dtype: dtype, backend: backend, aggregation: aggregation}
+  end
+
+  defp backend_from_args(args) do
+    Enum.find(args, fn
+      %__MODULE__{} = arg -> arg.backend
```
josevalim, line 215:
```suggestion
      %__MODULE__{backend: backend} -> backend
```

josevalim (approved):

PR #903, opened 2024-04-28
Initial timezone support

billylanchantin:
### Description

Adds timezone support.

```elixir
datetimes_with_timezones =
  ~U[2024-01-01T13:00:00.000000Z]
  |> DateTime.shift_zone!("America/New_York")
  |> List.wrap()

series = Explorer.Series.from_list(datetimes_with_timezones)
# #Explorer.Series<
#   Polars[1]
#   datetime[μs, America/New_York] [2024-01-01 07:00:00.000000-05:00 EST America/New_York]
# >

series.dtype
# {:datetime, :microsecond, "America/New_York"}

Explorer.Series.to_list(series)
# [#DateTime<2024-01-01 07:00:00.000000-05:00 EST America/New_York>]
```

### Dtype changes

New dtype:

* `{:datetime, precision, time_zone}`

And the old naive datetime dtype has become:

* `{:naive_datetime, precision}`

See discussion below. 

### To-do

* [ ] Support `%DateTime{}` literals in macros
* [ ] Backward-compatibility: alias `{:datetime, _}` for `{:naive_datetime, _}`


billylanchantin:
The main issue is that the new `{:datetime, precision, time_zone}` dtype can't be easily enumerated in code since any `time_zone` can be any valid time-zone string.

Before we were able to list literally every non-recursive dtype. That's no longer possible.

lib/explorer/series.ex
```diff
@@ -6550,14 +6620,23 @@ defmodule Explorer.Series do
     :"#{backend}.Series"
   end
 
-  defp dtype_error(function, dtype, valid_dtypes) when is_list(valid_dtypes) do
+  defp super_dtype_error(function, dtype, valid_super_dtypes) do
     raise(
       ArgumentError,
       "Explorer.Series.#{function} not implemented for dtype #{inspect(dtype)}. " <>
-        "Valid " <> Shared.inspect_dtypes(valid_dtypes, with_prefix: true)
+        "Valid dtypes are any subtype of #{inspect(valid_super_dtypes)}"
```
billylanchantin, line 6627: My attempt to handle the fact that we can't list all the dtypes easily anymore. Open to ideas on this.

josevalim:
> Also, as discussed on slack, I'll try to make {:datetime, _} be a valid alias for the new {:naive_datetime, _} dtype, possibly with a deprecation warning?

In my opinion it is fine to nuke it. It is a very easy change for folks to update in their own apps.

josevalim:
> Also, as discussed on slack, I'll try to make {:datetime, _} be a valid alias for the new {:naive_datetime, _} dtype, possibly with a deprecation warning?

In my opinion it is fine to nuke it. It is a very easy change for folks to update in their own apps. Alternatively, keep it as a shortcut for {:datetime

josevalim (approved):

mix.exs
```diff
@@ -43,6 +43,7 @@ defmodule Explorer.MixProject do
       {:table_rex, "~> 3.1.1 or ~> 4.0.0"},
       {:castore, "~> 1.0", optional: true},
       {:adbc, "~> 0.1", optional: true},
+      {:tz, "~> 0.26"},
```
josevalim, line 46: This should be a `only: :test` dependency. We should not enforce a TZ database onto users.

billylanchantin: Yeah I agree

billylanchantin: Addressed by: [306c789](https://github.com/elixir-explorer/explorer/pull/903/commits/306c7895e61470ef2e10927f2875a1de8212bcd2)

josevalim (approved):

native/explorer/src/datatypes.rs
```diff
@@ -352,12 +359,105 @@ impl From<NaiveDateTime> for ExDateTime {
...
+            None => duration.num_milliseconds() * 1_000,
+        }
+    }
+}
+
+// TODO-BILLY: finish this.
+impl<'a> From<DateTime<Tz>> for ExDateTime<'a> {
+    fn from(dt_tz: DateTime<Tz>) -> ExDateTime<'a> {
+        let & time_zone = dt_tz.offset().tz_id();
+        let & zone_abbr = dt_tz.offset().abbreviation();
```
billylanchantin, line 415: @philss (or anyone!) I've hit the limit of my ability to cargo cult Rust code 😭.

The `ExDateTime` struct currently looks like this:

```rust
#[derive(NifStruct, Copy, Clone, Debug)]
#[module = "DateTime"]
pub struct ExDateTime<'a> {
    // ...
    pub time_zone: &'a str,
    // ...
    pub zone_abbr: &'a str,
}
```

I need to derive these fields from the `DateTime` object, so they're not known at compile time. AFAICT I've got two options: they can have type `&'a str` or `String`. I can implement this if they have type `String`, but then we can no longer `#[derive(Copy)]`. So I've tried to implement it with `&'a str`, but I haven't figured out how yet.

Any pointers would be appreciated!

philss: Oh man, I really don't know how to solve this easily. Sorry :/
Do we need to support `Copy`? I would try to use `String` if we don't need this.

lib/explorer/polars_backend/shared.ex
```diff
@@ -187,8 +187,9 @@ defmodule Explorer.PolarsBackend.Shared do
       :category -> Native.s_from_list_categories(name, list)
       :date -> Native.s_from_list_date(name, list)
       :time -> Native.s_from_list_time(name, list)
-      {:datetime, precision} -> Native.s_from_list_datetime(name, list, Atom.to_string(precision))
-      {:duration, precision} -> Native.s_from_list_duration(name, list, Atom.to_string(precision))
+      {:naive_datetime, precision} -> Native.s_from_list_naivedatetime(name, list, precision)
```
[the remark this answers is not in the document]
billylanchantin: Ah yeah good catch. Gonna do this in a separate commit so I get the other side too.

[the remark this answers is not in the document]
billylanchantin: [6c715f9](https://github.com/elixir-explorer/explorer/pull/903/commits/6c715f90b228dc6a24e25b204597e65bb399d5b9)

native/explorer/src/datatypes.rs
billylanchantin: [8f5be24](https://github.com/elixir-explorer/explorer/pull/903/commits/8f5be2464294275774e187fe991b60e6d3b6ed18)

Ok, so we _did_ end up needing `Copy` because of how I wrote `s_from_list_datetime`. Specifically, this breaks:

```rust
val: Vec<Option<ExDateTime>>
// ...
val.iter()
    .map(|dt| dt.map(|dt| dt.into()))
//            ^^^^^^^^^^^^^^^^^^^^^^ this implicitly requires Copy
    .collect::<Vec<Option<i64>>>(),
```

I've reverted to `&'a str` for now since all the tests pass. I suspect something we try with `%DateTime{}` will eventually break because of this, but I say let's address it then.

Thoughts?

native/explorer/Cargo.toml
```diff
@@ -43,44 +44,45 @@ features = [
...
+  "rank",
+  "regex",
   "rolling_window",
+  "round_series",
   "rows",
   "simd",
+  "streaming",
+  "strings",
   "temporal",
+  "timezones",
```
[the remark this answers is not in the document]
philss: No problem. Yeah, I agree that if we document that, this is fine :)

native/explorer/src/datatypes.rs
philss: No thoughts for now :/
Let's go with this version, and we improve in the future :D

philss (approved):
This is a great addition! 🚀

billylanchantin:
Decided to go with what José suggested and nuke `:datetime` in favor of `:naive_datetime`. We can add `:datetime` back in as an alias before the next release if the change seems too disruptive.

PR #909, opened 2024-05-15
Fix DateTime -> ExDateTime lifetime issue

elbow-jason:
This PR continues the work started in #903.

@billylanchantin  and I fixed the lifetime issues for `From<DateTime<Tz>> for ExDateTime`. It was **super** close already. 

It seems to work:

```elixir
iex(1)> Explorer.PolarsBackend.Native.s_from_list_datetime("blep", [DateTime.utc_now()], :microsecond, "America/New_York") |> Explorer.PolarsBackend.Native.s_to_list()
{:ok, [#DateTime<2024-05-15 16:00:45.797338-04:00 EDT America/New_York>]}
```

I think we should consider this PR a WIP until we get some tests in place.


josevalim (approved):

billylanchantin:
I did not figure out why this code is never called, but I'm gonna merge anyway. We can always remove it later if it's unnecessary.

PR #912, opened 2024-05-25
Add `DF.transform`

billylanchantin:
## Description

Adds `DF.transform/3` which is the analogous function to `S.transform/2`. I've needed a version of this function many times in my own work.

## Example

```elixir

df = DF.new(
  numbers: [1, 2],
  datetime_local: [~N[2024-01-01 00:00:00], ~N[2024-01-01 00:00:00]],
  timezone: ["Etc/UTC", "America/New_York"]
)

DF.transform(df, [names: ["datetime_local", "timezone"]], fn row ->
  datetime_utc =
    row["datetime_local"]
    |> DateTime.from_naive!(row["timezone"])
    |> DateTime.shift_zone!("Etc/UTC")

  %{datetime_utc: datetime_utc}
end)

# #Explorer.DataFrame<
#   Polars[2 x 4]
#   numbers s64 [1, 2]
#   datetime_local naive_datetime[μs] [2024-01-01 00:00:00.000000, 2024-01-01 00:00:00.000000]
#   timezone string ["Etc/UTC", "America/New_York"]
#   datetime_utc datetime[μs, Etc/UTC] [2024-01-01 00:00:00.000000Z, 2024-01-01 05:00:00.000000Z]
# >
```

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -154,13 +154,8 @@ defmodule Explorer.Backend.DataFrame do
   @callback collect(df) :: df
   @callback from_tabular(Table.Reader.t(), dtypes) :: df
   @callback from_series([{binary(), Series.t()}]) :: df
-  @callback to_rows(df, atom_keys? :: boolean(), names :: [column_name()]) :: [map()]
-  @callback to_rows_stream(
-              df,
-              atom_keys? :: boolean(),
-              chunk_size :: integer(),
-              names :: [column_name()]
-            ) :: Enumerable.t()
+  @callback to_rows(df, atom_keys? :: boolean()) :: [map()]
+  @callback to_rows_stream(df, atom_keys? :: boolean(), chunk_size :: integer()) :: Enumerable.t()
 
   # Introspection
 
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -154,8 +154,13 @@ defmodule Explorer.Backend.DataFrame do
   @callback collect(df) :: df
   @callback from_tabular(Table.Reader.t(), dtypes) :: df
   @callback from_series([{binary(), Series.t()}]) :: df
-  @callback to_rows(df, atom_keys? :: boolean()) :: [map()]
-  @callback to_rows_stream(df, atom_keys? :: boolean(), chunk_size :: integer()) :: Enumerable.t()
+  @callback to_rows(df, atom_keys? :: boolean(), names :: [column_name()]) :: [map()]
+  @callback to_rows_stream(
+              df,
+              atom_keys? :: boolean(),
+              chunk_size :: integer(),
+              names :: [column_name()]
+            ) :: Enumerable.t()
 
   # Introspection
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -1934,11 +1934,7 @@ defmodule Explorer.DataFrame do
 
   ## Options
 
-    * `:atom_keys` - Configure if the resultant maps should have atom keys.
-      (default: `false`)
-
-    * `:names` - Column names to select for serializing.
-      By default all columns will be returned.
+    * `:atom_keys` - Configure if the resultant maps should have atom keys. (default: `false`)
 
   ## Examples
 
@@ -1953,9 +1949,9 @@ defmodule Explorer.DataFrame do
   @doc type: :conversion
   @spec to_rows(df :: DataFrame.t(), Keyword.t()) :: [map()]
   def to_rows(df, opts \\ []) do
-    opts = Keyword.validate!(opts, atom_keys: false, names: nil)
+    opts = Keyword.validate!(opts, atom_keys: false)
 
-    Shared.apply_impl(df, :to_rows, [opts[:atom_keys], opts[:names]])
+    Shared.apply_impl(df, :to_rows, [opts[:atom_keys]])
   end
 
   @doc """
@@ -1969,14 +1965,8 @@ defmodule Explorer.DataFrame do
 
   ## Options
 
-    * `:atom_keys` - Configure if the resultant maps should have atom keys.
-      (default: `false`)
-
-    * `:chunk_size` - Number of rows passed to `to_rows/2` while streaming over
-      the data. (default: `1000`)
-
-    * `:names` - Column names to select for serializing.
-      By default all columns will be returned.
+    * `:atom_keys` - Configure if the resultant maps should have atom keys. (default: `false`)
+    * `:chunk_size` - Number of rows passed to `to_rows/2` while streaming over the data. (default: `1000`)
 
   ## Examples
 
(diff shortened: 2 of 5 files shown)


lib/explorer/data_frame.ex
```diff
@@ -4227,6 +4227,103 @@ defmodule Explorer.DataFrame do
...
+  > copy the whole dataframe in memory. Prefer to use the operations in this
+  > module rather than native Elixir ones whenever possible, as this module is
+  > optimized for large series and does not need to pay the serialization cost.
+  """
+  @doc type: :single
+  @spec transform(DataFrame.t(), column_name(), Keyword.t(), fun()) ::
+          DataFrame.t()
+  def transform(df, new_col_name, opts \\ [], fun)
+      when is_column_name(new_col_name) and is_function(fun, 1) do
+    opts = Keyword.validate!(opts, atom_keys: false, columns: nil)
```
billylanchantin, line 4303: Feel free to bike shed on argument order, option names, etc. I have no strong opinions.

josevalim:
This operation is doing three things at the moment:

1. selecting
2. converting to rows
3. merging the columns (which we call concat_columns)

It also has a limitation that it computes one single column. For example, instead we could have:

```
DF.transform(df, [columns: ["datetime_local", "timezone"]], fn row ->
  [datetime_utc: row["datetime_local"]
  |> DateTime.from_naive!(row["timezone"])
  |> DateTime.shift_zone!("Etc/UTC")]
end)
```

We could also emit a custom row struct that accepts both strings and atom keys and converts fields as necessary. For example, imagine we had a `%Explorer.DataFrame.Row{index: index, df: df}`. When you called `row["datetime_local"]`, it would get that particular column and access it as `index`. Does Polars guarantee constant time access to all of its rows? If it does, then we can provide both atom/string ergonomics and only convert the necessary keys lazily.

josevalim:
However, we should benchmark the approaches. The lazy one may end-up being less efficient if we do too many trips to Rust. We should certainly have a single operation to access a given column+row.

billylanchantin:
José I didn't want to use my brain today :P

> It also has a limitation that it computes one single column.

👍 Yeah we could definitely get multiple columns with `concat_columns`. Great suggestion.

Validation may be a bit of an issue. I'd probably just want to not bother and let `concat_columns` raise. Maybe with a `try do ... reraise ...` to give a good error message?

> Does Polars guarantee constant time access to all of its rows?

I don't think so but I'm not sure. I couldn't find a definite answer in the docs.

They seem to support several kinds of index-based access and I'm not sure which is the "right" one. Following some source code led me to this file:

* https://github.com/pola-rs/polars/blob/17cf8b7e17974b51c9bdc760422005fe591bc7c5/crates/polars-core/src/chunked_array/ops/gather.rs#L149

If this is the right place, I see several references to binary searches. That makes me think it's $O(n \cdot log(n))$. Maybe they can get good amortized performance?

> However, we should benchmark the approaches. The lazy one may end-up being less efficient if we do too many trips to Rust. We should certainly have a single operation to access a given column+row.

Yeah definitely some benchmarks are in order. I suspect the most expensive part is the de-serialization step required to feed the Elixir functions. I'll try your lazy approach and get back with some numbers.

I also want to try and leverage Arrow's chunking. If de-serializing a single chunk is fast, it may be worth parallelizing over chunks on the Elixir side rather than trying to trick Polars into doing what we want. IDK how easy that level of control will be though.

josevalim:
My understanding from the Rust code is that they do a binary search only if there are several chunks. What we may want to do is to rechunk the dataframe before using it. Another potential concern here is doing the bounds check on every operation, but they do have an `_unchecked` version.

lib/explorer/data_frame.ex
```diff
@@ -4227,6 +4237,93 @@ defmodule Explorer.DataFrame do
...
+  @doc type: :single
+  @spec transform(DataFrame.t(), Keyword.t(), fun()) :: DataFrame.t()
+  def transform(df, opts \\ [], row_transform) when is_function(row_transform, 1) do
+    df_new =
+      df
+      |> to_rows_stream(opts)
+      |> Enum.map(row_transform)
+      |> new()
+
+    concat_columns(df, df_new)
```
billylanchantin, line 4324: @josevalim `DF.transform` now works with multiple columns.

Also I felt like we were optimizing in the wrong place so I tried a different approach. Now this is the whole implementation.

Instead, I added a `:names` option to `to_rows` and `to_rows_stream`. Since we need to work with the output of these functions at the end of the day, I feel like any optimization we do should happen there.

lib/explorer/backend/data_frame.ex
```diff
@@ -154,8 +154,13 @@ defmodule Explorer.Backend.DataFrame do
...
   @callback from_series([{binary(), Series.t()}]) :: df
-  @callback to_rows(df, atom_keys? :: boolean()) :: [map()]
-  @callback to_rows_stream(df, atom_keys? :: boolean(), chunk_size :: integer()) :: Enumerable.t()
+  @callback to_rows(df, atom_keys? :: boolean(), names :: [column_name()]) :: [map()]
+  @callback to_rows_stream(
+              df,
+              atom_keys? :: boolean(),
+              chunk_size :: integer(),
+              names :: [column_name()]
+            ) :: Enumerable.t()
```
josevalim, line 163: Isn't the names argument here equivalent to doing a select before?

josevalim: I think we can remove the argument here and do a DF.select inside DF.transform.

billylanchantin: `DF.select` inside `DF.transform` is how it was before. I liked the idea of moving it to `DF.to_rows` and `DF.to_rows_stream` because:

1. Those functions become more expressive.
2. `DF.transform` also becomes more performant.

Right now the `to_rows` functions zip column-by-column. If you simply tell them to serialize fewer columns, we achieve the same thing but without the extra `select`.

josevalim: > DF.transform also becomes more performant.

How does it become more performant? Select is a very cheap operation on the dataframes side. The cost of serialization will easily trump anything else?

The downside is having more ways of achieving the same thing, larger API surfaces for adapters to implement, and so on. While the differences are minimal:

```
df |> DF.select([:foo, :bar]) |> DF.to_rows()
df |> DF.to_rows(names: [:foo, :bar])
```

billylanchantin: > How does it become more performant?

I just mean you need to pay the serialization cost from `to_rows` either way, so the pre-`select` doesn't buy anything if you can tell `to_rows` to serialize fewer columns.

> The downside is having more ways of achieving the same thing, larger API surfaces for adapters to implement, and so on.

But this logic is internal to `DF.transform`, no? Or do you mean adding `:names` to `to_rows` is more to maintain?

billylanchantin: Oh wait I'm seeing your point. Since `select` is so cheap, let's have that be _the way_. 👍

josevalim: > Oh wait I'm seeing your point. Since select is so cheap, let's have that be the way. 👍

Exactly. I think `:names` makes sense for DF.transform, because you don't want to select before, but not for the smaller ones.

lib/explorer/data_frame.ex
billylanchantin: That option was later removed from `to_rows` and `to_rows_stream`.

josevalim (approved):

PR #913, opened 2024-05-27
Bump Polars to 0.40.0

lkarthee:


lkarthee:
I will continue where phils left off..

lkarthee:
I will continue  work on phils  branch https://github.com/elixir-explorer/explorer/tree/ps-bump-polars-to-v0.39.2

Issue #818, opened 2024-01-13
Expose the Polars SQL API?


cigrainger:
https://crates.io/crates/polars-sql
https://docs.rs/polars-sql/0.36.2/polars_sql/fn.sql_expr.html
https://docs.pola.rs/py-polars/html/reference/sql

I'm not a huge fan of python polars here with registering etc explicitly. It feels very... Spark-ish to me.

I like the way dplyr allows you to just write `sql(...)`. https://dbplyr.tidyverse.org/articles/sql.html

We could have `Explorer.DataFrame.sql(df|lf, "statement")`.

I'm very open to differing opinions on this!

josevalim:
At first, I am -1, unless we have a very strong use case for this. Explorer is meant to be agnostic, and we work hard for this to be the case, but the sql semantics would be necessarily tied to polars here.

cigrainger:
It definitely is a big win for OLAP use cases. I'm not sure I understand why we'd be tied to Polars SQL semantics here -- for an `explorer_sql` backend it can just be passed through directly. Backends don't necessarily _have_ to support it. And it can be documented that the accepted SQL queries depend on the backend.

josevalim:
In explorer_sql, sql is how you load data into memory. Polars SQL is a language for manipulating an existing in-memory dataframe. So they sit at different levels. And if I use Polars SQL to manipulate a Polars DataFrame, I won't be able to use to manipulate a Explorer SQL dataframe.

cigrainger:
Ah but that's not how I'm imagining `explorer_sql` -- the idea is _not_  just to load the data into memory (in which case there's no point -- just use ADBC or whatever else) but to convert Explorer functions to SQL and run the queries on the database itself until you pull it down. `dbplyr` uses tmp tables. You can also fix to actual tables.

I don't see a problem with the SQL syntax being backend dependent -- since it's passing through it would be handled by the backend anyway. This is also how `dbplyr` works.

josevalim:
Interesting. That would indeed close the gap, so I am ok with going down this path, although I wonder if we should tackle Explorer SQL first to have a better of how it will all connect. :)

cigrainger:
I'm cool with that! Though I'm pretty adamant that in `explorer_sql` this function would basically just act as a transparent 'passthrough' escape hatch, permitting you to run raw SQL on the db directly.

billylanchantin:
[...]

PR #916, opened 2024-06-01
[Proof of concept] Expr serialize

billylanchantin:
### 🚧 Do not merge

This PR is only a proof of concept.

### Description

Adds `expr_to_json` and `expr_from_json`.

The idea is to provide an escape hatch for Polars functionality we don't yet support. Included is a test case demonstrating that it's possible.

### Discussion

Explorer can lag behind what's available in Polars. This functionality would allow users to opt into behavior not otherwise available. My hope is that we can turn this idea into something like [`Ecto.Query.API.fragment/1`](https://hexdocs.pm/ecto/Ecto.Query.API.html#fragment/1).

This approach is messy for a few reasons:

* You have to somehow get your hands on the JSON
* The escape hatch is Polars-specific
* We'd have to weave it into many functions (though I have thoughts here)

But for that price, we instantly open up all expression-based operations. Please let me know if you think!

### DIFF
--- a/lib/explorer/polars_backend/expression.ex
+++ b/lib/explorer/polars_backend/expression.ex
@@ -356,6 +356,18 @@ defmodule Explorer.PolarsBackend.Expression do
     Native.expr_describe_filter_plan(polars_df, expression)
   end
 
+  def to_json(%__MODULE__{} = expression) do
+    expression
+    |> Native.expr_to_json()
+    |> Jason.decode!()
+  end
+
+  def from_json(%{} = json_map) do
+    json_map
+    |> Jason.encode!()
+    |> Native.expr_from_json()
+  end
+
   defp dtype(%LazySeries{dtype: dtype}), do: dtype
 
   defp dtype(%PolarsSeries{} = polars_series) do
--- a/lib/explorer/polars_backend/native.ex
+++ b/lib/explorer/polars_backend/native.ex
@@ -196,6 +196,8 @@ defmodule Explorer.PolarsBackend.Native do
   def expr_datetime(_datetime), do: err()
   def expr_duration(_duration), do: err()
   def expr_describe_filter_plan(_df, _expr), do: err()
+  def expr_to_json(_expr), do: err()
+  def expr_from_json(_expr), do: err()
   def expr_float(_number), do: err()
   def expr_integer(_number), do: err()
   def expr_int_range(_start, _end, _step, _dtype), do: err()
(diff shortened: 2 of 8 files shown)


josevalim:
I am not sure we should go down this route because this is serializing an internal representation, right? So there is no guarantee that, once a new Polars version lands, the JSON representation (keys and values) will remain relevant? I think relying on their SQL API should be slightly more stable, because that's a public API?

billylanchantin:
> So there is no guarantee that, once a new Polars version lands, the JSON representation (keys and values) will remain relevant?

The API is public:

* https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.meta.serialize.html
* https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.deserialize.html#polars.Expr.deserialize

And they don't _say_ you can't rely on it. But I think you're likely correct regardless. Looking at the example JSON from my test, it seems they're including a lot of what looks like internal stuff.

In theory if a particular representation breaks, a user could just get the new JSON from the new version of Polars. I'd be ok with that since this is supposed to be a backdoor. But I take your point!

> I think relying on their SQL API should be slightly more stable, because that's a public API?

That's true. However their docs say:

> As the DataFrame interface is primary, new features are typically added to the expression API first.

So I'm not sure exposing the SQL API would also expose the expressions we want, especially since:

> There is no separate SQL engine because Polars translates SQL queries into [expressions](https://docs.pola.rs/user-guide/concepts/expressions/), which are then executed using its own engine.

But we should _definitely_ pursue the SQL API regardless. How cool would it be for `Explorer.DataFrame` to implement `Ecto.Queryable`?

Still, I don't think the SQL API gets us the expressiveness I'm going for. But not exposing an API because it's internal/unreliable is potentially a deal breaker, so great feedback.

billylanchantin:
> If Polars had an API for building an expression from a SQL fragment, that would be ideal.

Seems like they do!

* https://docs.rs/polars-sql/latest/polars_sql/fn.sql_expr.html

josevalim:
@billylanchantin YES!!!

billylanchantin:
@josevalim Ok I can close this and start working on exposing a SQL-to-exprs pipeline :)

FYI you made a good point here: https://github.com/elixir-explorer/explorer/issues/818#issuecomment-1953652209

But maybe viewing the SQL as somewhat of an escape hatch changes things a bit?

josevalim:
> But maybe viewing the SQL as somewhat of an escape hatch changes things a bit?

Exactly. To me, the main question is if we can provide a consistent behaviour across backends. For example, if we have a Postgres backend, would this feature also make sense there? And I think it would, so let's go for it.

billylanchantin:
Closing in favor of: https://github.com/elixir-explorer/explorer/pull/918

PR #917, opened 2024-06-02
Bump Polars to v0.41.3

lkarthee:


lib/explorer/series.ex
```diff
@@ -3617,7 +3617,7 @@ defmodule Explorer.Series do
   defp cast_to_pow({:f, l}, {:f, r}), do: {:f, max(l, r)}
   defp cast_to_pow({:f, l}, {n, _}) when K.in(n, [:u, :s]), do: {:f, l}
   defp cast_to_pow({n, _}, {:f, r}) when K.in(n, [:u, :s]), do: {:f, r}
-  defp cast_to_pow({n, _}, {:s, _}) when K.in(n, [:u, :s]), do: {:f, 64}
+  defp cast_to_pow({n, _}, {:s, _}) when K.in(n, [:u, :s]), do: {:s, 64}
```
josevalim, line 3620: In theory this has to be a flat because the exponent can be negative, which means a float is returned. Did Polars change behaviour here and it now raises for negative exponents?

lkarthee: https://github.com/pola-rs/polars/pull/15506

billylanchantin: If I'm reading that right, we need to disallow raising to a signed int completely now.

native/explorer/rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2024-06-23"
+channel = "nightly-2024-06-24"
```
lkarthee, line 2: https://github.com/pola-rs/polars/issues/17137

test/explorer/data_frame/grouped_test.exs
```diff
@@ -578,11 +578,11 @@ defmodule Explorer.DataFrame.GroupedTest do
...
 
           [
-            b: Series.window_max(a, 2, weights: [1.0, 2.0]),
+            # b: Series.window_max(a, 2, weights: [1.0, 2.0])
             c: Series.window_mean(a, 2, weights: [0.25, 0.75]),
             d: Series.window_median(a, 2, weights: [0.25, 0.75]),
-            e: Series.window_min(a, 2, weights: [1.0, 2.0]),
-            f: Series.window_sum(a, 2, weights: [1.0, 2.0]),
+            # e: Series.window_min(a, 2, weights: [1.0, 2.0])
+            # f: Series.window_sum(a, 2, weights: [1.0, 2.0])
```
lkarthee, line 585: These should be enabled - disabled as these are  crashing polars

lkarthee (commented):
```
Finished in 3.5 seconds (3.5s async, 0.00s sync)
610 doctests, 1 property, 1485 tests, 71 failures, 20 excluded, 1 skipped

Randomized with seed 279745
```
Most of the test cases fail because:
- [x] pow dtype change
- [x] pivot_wider, pivot_longer
- [x] string slice
- [x] cross join param
- [x] window functions
- [x] commented few tests which are crashing polars

lib/explorer/series.ex
philss: I think we can still keep the same behaviour - always have a float series if we `pow(any, signed_int)` -, as I did in https://github.com/elixir-explorer/explorer/pull/917/commits/2644d270bf8dfebb22b91c0713a8be9d1c208d10

But I think we could change a little how we create series from scalar numbers. If the number is a positive integer, we could create the series as `:u32` (or adjust to the size that most fits the number). Doing this would use less memory and would remove the surprise of having the result of `pow(signed, signed)` to be a float. WDYT?

josevalim: It is fine to mirror Polars behaviour, because that's also how Nx behaves too. If integers are given (signed or unsigned), it raises for negative values. If you give a float, it preservers the float type.

billylanchantin: > But I think we could change a little how we create series from scalar numbers.

@philss I think we've had a similar discussion before but I can't find it right now. @lkarthee was also advocating inferring the "simplest" integer dtype from a list of scalars.

IIRC, we decided to not do that because it was limiting in unexpected ways. For example, if we infer `[1, 2, 3]` as `dtype: :u8`, then we can't do subtraction that series since the result may contain negative values.

If I find the discussion I'll update. Also, the possibility is still up for discussion IMO.

billylanchantin: Update, found it:

* https://github.com/elixir-explorer/explorer/pull/826

philss: @billylanchantin thank you for the context! ❤️ 
I had completely forgot about that discussion 😬

philss:
We are almost there! But a problem regarding the cloud integration is still blocking this. I'm trying to fix it by updating `object_store` and our custom `CloudWriter` (Rust side), but I suspect that the problem is more complex due to some Polars' changes. I will keep trying :)

lib/explorer/series.ex
billylanchantin: You betcha!

philss:
There is a good news and a bad news. The good news is that we have only one failing test! 🎉 🥳 
But the bad news is that we may need to submit changes to Polars in order to fix the failing test.

In a nutshell, the problem is on the `LazyFrame#sink_parquet_cloud` function from Polars that is failing **probably** because of the reasons I wrote in this issue: https://github.com/pola-rs/polars/issues/17172#issuecomment-2243328412

We can either wait for the fix to land on Polars, or deactivate the feature until there. This is going to affect people using the `:streaming` option of [`to_parquet/2`](test/explorer/data_frame/lazy_test.exs) (cloud version). WDYT @josevalim @lkarthee @billylanchantin?

josevalim:
What do we for when streaming is not supported? We raise or we just do the best we can? I am worried Polars is not ultimately interested in maintaining these features. We have another related PR that is open for quite some time.

philss:
> What do we for when streaming is not supported? We raise or we just do the best we can?

I think we can raise if it's not possible to support a given feature, but we can try to implement "hacks" when possible, until we need to remove the feature all together. Unfortunately we can't have our patches to the Polars codebase, and I don't think it's a good idea to maintain a fork to solve our issues. WDYT?

> I am worried Polars is not ultimately interested in maintaining these features. We have another related PR that is open for quite some time.

Yeah, I'm worried too. Maybe it's not their focus to maintain this cloud integration in their Rust codebase.

josevalim:
Yeah, let's raise for now and add a note to the CHANGELOG. Meanwhile, if you could send a PR to them, it would be great!

philss:
@josevalim OK, I will add a note to the change log later. [My PoC](https://github.com/elixir-explorer/explorer/pull/917/commits/d64af7ef687aaef6ef294ee2db293dbdc347c392) didn't work out, but I was wondering if we could still "enable streaming", but first stream to a temporary file, and them read from that file and send to the cloud. This could keep the convenience of saving to S3 right from Explorer, with the downside of requiring some space in disk, and probably extra time. WDYT?

> Meanwhile, if you could send a PR to them, it would be great!

Yeah, I will try that! Something similar to our `CloudWriter` may work.

kartheekusf:
+1 for disabling and documenting it in change log.

lkarthee:
+1 for disabling and we should document it in the changelog.

CHANGELOG.md
```diff
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- The `Explorer.Series.substring/3` function no more cycle to the back of string if
+  the negative offset surpass the beginning of that string. In that case, an empty
+  string is returned.
+
+- All the series functions related to "EWM" are now propagating the `nil` values instead
+  of filling the missing values with the previous one.
+
+- Saving a dataframe as a Parque file to S3 services no longer works when streaming is
+  enabled. This is temporary due to a bug in Polars. An exception should be raised instead.
```
billylanchantin, line 27:
```suggestion
- `Explorer.Series.pow/2` no longer casts to float when the exponent is a signed
  integer. We are following the way Polars works now, which is to try to execute
  the operation or raise an exception in case the exponent is negative.

- `Explorer.Series.pivot_wider/4` no longer includes the `names_from` column
  name in the new columns when `values_from` is a list of columns. This is more
  consistent with its behaviour when `values_from` is a single column.

- `Explorer.Series.substring/3` no longer cycles to the end of the string if the
  negative offset surpasses the beginning of that string. In that case, an empty
  string is returned.

- The `Explorer.Series.ewm_*` functions no longer replace `nil` values with the
  value at the previous index. They now propogate `nil` values through to the
  result series.

- Saving a dataframe as a Parquet file to S3 services no longer works when
  streaming is enabled. This is temporary due to a bug in Polars. An exception
  should be raised instead.
```

billylanchantin (changes requested):
Looks great! I rewrote the changelog a tiny bit so each bullet was similarly worded (and fixed a typo).

Also I'll be honest, I skimmed the Rust code. But my Rust skills are nascent so I'm only sorta helpful on that side of the house anyway 😅. The Elixir side seems solid, though!

billylanchantin (approved):

josevalim (approved):

PR #918, opened 2024-06-02
[WIP] Initial SQL support

billylanchantin:
### Description

This PR experiments with initial SQL support. From the doctest:

```elixir
iex> Explorer.DataFrame.execute_sql(df, "max(a)")
# #Explorer.DataFrame<
#   Polars[1 x 1]
#   a s64 [3]
# >
```

### Open questions

* [ ] Overall API structure? Right now I'm just calling `lf.select(expr)`, but we could do much more.
* [ ] Implement `Ecto.Queryable`?

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -5946,6 +5946,22 @@ defmodule Explorer.DataFrame do
     Shared.apply_impl(df, :covariance, [out_df, opts[:ddof]])
   end
 
+  # SQL
+
+  @doc """
+  ## Examples
+
+      iex> df = Explorer.DataFrame.new(a: [1, 2, 3])
+      iex> Explorer.DataFrame.execute_sql(df, "max(a)")
+      #Explorer.DataFrame<
+        Polars[1 x 1]
+        a s64 [3]
+      >
+  """
+  def execute_sql(%__MODULE__{} = df, sql_string) when is_binary(sql_string) do
+    Shared.apply_impl(df, :execute_sql, [sql_string])
+  end
+
   # Helpers
 
   defp backend_from_options!(opts) do
--- a/lib/explorer/polars_backend/lazy_frame.ex
+++ b/lib/explorer/polars_backend/lazy_frame.ex
@@ -580,6 +580,13 @@ defmodule Explorer.PolarsBackend.LazyFrame do
     Eager.re_dtype(regex_as_string)
   end
 
+  @impl true
+  def execute_sql(ldf, sql_string) do
+    ldf.data
+    |> Native.lf_execute_sql(sql_string)
+    |> Shared.create_dataframe()
+  end
+
   not_available_funs = [
     correlation: 4,
     covariance: 3,
--- a/native/explorer/src/lazyframe.rs
+++ b/native/explorer/src/lazyframe.rs
@@ -344,3 +344,11 @@ pub fn lf_concat_columns(
 
     Ok(ExLazyFrame::new(out_df.drop([id_column])))
 }
+
+#[rustler::nif]
+pub fn lf_execute_sql(lf: ExLazyFrame, sql_string: &str) -> ExLazyFrame {
+    let lf = lf.clone_inner();
+    let expr = polars::sql::sql_expr(sql_string).unwrap();
+    let lf_sql = lf.select(vec![expr]);
+    ExLazyFrame::new(lf_sql)
+}
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -241,6 +241,10 @@ defmodule Explorer.Backend.DataFrame do
   @callback summarise_with(df, out_df :: df(), aggregations :: [{column_name(), lazy_series()}]) ::
               df
 
+  # SQL
+
+  @callback execute_sql(df, sql_string :: binary()) :: df()
+
   # Functions
   alias Explorer.{DataFrame, Series}
 
--- a/lib/explorer/polars_backend/data_frame.ex
+++ b/lib/explorer/polars_backend/data_frame.ex
@@ -878,6 +878,16 @@ defmodule Explorer.PolarsBackend.DataFrame do
     |> LazyFrame.collect()
   end
 
+  # SQL
+
+  @impl true
+  def execute_sql(%DataFrame{} = df, sql_string) do
+    df
+    |> lazy()
+    |> LazyFrame.execute_sql(sql_string)
+    |> LazyFrame.collect()
+  end
+
   # Inspect
 
   @impl true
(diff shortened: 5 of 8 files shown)


lib/explorer/data_frame.ex
```diff
@@ -5946,6 +5949,52 @@ defmodule Explorer.DataFrame do
...
+      #Explorer.DataFrame<
+        Polars[2 x 2]
+        a list[s64] [[1], [2, 3]]
+        b string ["x", "y"]
+      >
+
+  Custom table name:
+
+      iex> df = Explorer.DataFrame.new(a: [1, 2, 3])
+      iex> Explorer.DataFrame.sql(df, "select a + 1 from my_table", table_name: "my_table")
```
philss, line 5983: I'm curious if it will propagate to the next SQL call involving this resultant DF. Does Polars saves it with the DF?

lib/explorer/polars_backend/lazy_frame.ex
```diff
@@ -29,7 +29,21 @@ defmodule Explorer.PolarsBackend.LazyFrame do
...
+      if :unknown in Map.values(df.dtypes) do
+        # If an `:unknown` dtype is present, we were not able to determine the
+        # full set of dtypes from the expressions themselves and so we need to
+        # fetch them from Polars.
+        Shared.create_dataframe(df.data)
+      else
+        df
+      end
+
+    df
```
philss, line 45: I think you can remove the `df` assignment here:

```suggestion
    if :unknown in Map.values(df.dtypes) do
      # If an `:unknown` dtype is present, we were not able to determine the
      # full set of dtypes from the expressions themselves and so we need to
      # fetch them from Polars.
      Shared.create_dataframe(df.data)
    else
      df
    end
```

lib/explorer/query/sql.ex
```diff
@@ -0,0 +1,23 @@
...
+  def to_lazy_series(%__MODULE__{string: sql_string}) do
+    aggregation? = true
+    dtype_out = :unknown
+
+    Explorer.Backend.LazySeries.new(
+      :sql,
+      [sql_string],
+      dtype_out,
+      aggregation?,
+      Explorer.PolarsBackend
```
philss, line 20: Maybe we want to fetch this from `Explorer.Backend.get()`. It will return exactly the same, but let's people use their backend. WDYT?

```suggestion
      Explorer.Backend.get()
```

native/explorer/src/lazyframe.rs
```diff
@@ -344,3 +344,14 @@ pub fn lf_concat_columns(
...
 }
+
+#[rustler::nif]
+pub fn lf_sql(lf: ExLazyFrame, sql_string: &str, table_name: &str) -> ExLazyFrame {
+    let mut ctx = polars::sql::SQLContext::new();
+
+    let lf = lf.clone_inner();
+    ctx.register(table_name, lf);
+
+    let lf_sql = ctx.execute(sql_string).unwrap();
```
philss, line 355: I think it's fine for this first version, but would be nice to treat the error and return a `Result<ExLazyFrame, ExplorerError>` for this function.

philss (commented):
Awesome job! 🚀 

I added some minor points :)

lib/explorer/data_frame.ex
billylanchantin: Great question. Thankfully Polars does what you'd hope:

```elixir
iex(1)> df1 = Explorer.DataFrame.new(a: [1, 2, 3])
#Explorer.DataFrame<
  Polars[3 x 1]
  a s64 [1, 2, 3]
>

iex(2)> df2 = Explorer.DataFrame.sql(df1, "select a + 1 from my_table", table_name: "my_table")
#Explorer.DataFrame<
  Polars[3 x 1]
  a s64 [2, 3, 4]
>

iex(3)> df3 = Explorer.DataFrame.sql(df2, "select a + 1 from my_table", table_name: "my_table")
#Explorer.DataFrame<
  Polars[3 x 1]
  a s64 [3, 4, 5]
>
```

You actually can use the `polars::sql::SQLContext` to remember old dataframes, join multiple dataframes, etc. I'm using it very simplistically: creating a new one on every call to `DF.sql`.

native/explorer/src/lazyframe.rs
billylanchantin: I'd be fine doing it on this version too. Any special considerations you can think of for error handling?

philss: No, the idea is just to avoid unwrap panics :)

lib/explorer/data_frame.ex
philss: I see. Nice! We can try to use that more in the future. I mean, save it to the `ExDataFrame` struct and propagate.

native/explorer/src/lazyframe.rs
billylanchantin: [1882109](https://github.com/elixir-explorer/explorer/pull/918/commits/188210988b7c13e3d95e12be405c3cb9ef721cff)

lib/explorer/data_frame.ex
billylanchantin: Yep! They have a concept of "registering" tables, which is how you do multi-table stuff. So we could either wrap that functionality in our own API or just expose the underlying functionality as is.

philss (approved):

lib/explorer/polars_backend/shared.ex
```diff
@@ -80,6 +80,21 @@ defmodule Explorer.PolarsBackend.Shared do
     end
   end
 
+  defp dtypes_match?(check_dtypes, out_dtypes) do
+    if map_size(check_dtypes) != map_size(out_dtypes) do
+      false
+    else
+      # The dtypes for each key must either be the same or `:unknown`.
+      Enum.reduce_while(check_dtypes, true, fn {key, check_value}, _ ->
```
josevalim, line 88: Maybe Enum.all? is enough here?

josevalim (commented):
@billylanchantin this is looking great! Is my understanding correct that we have two features in this pull request? One is to write SQL at the root level and another is to embed SQL inside mutate? Can I propose for us to add only the first one for now and then evolve it from there?

lib/explorer/polars_backend/shared.ex
billylanchantin: Whoops yeah definitely [fc44836](https://github.com/elixir-explorer/explorer/pull/918/commits/fc44836909c9ebfee0e62b141cd427006c2f6acc)

billylanchantin:
@josevalim I'm good with that! `DF.sql` is more powerful and has actual documentation :)

Any particular hesitation with the embedded SQL part? Or are you just favoring smaller changes since this is experimental?

josevalim:
Just favoring something smaller for now. It is always easier to add stuff than to remove them. :)

Issue #902, opened 2024-04-25
Regression in `DataFrame.concat_rows/2` in v0.8.2


maennchen:
## Script

```elixir
Mix.install([{:explorer, "0.8.2"}])

df_a = Explorer.DataFrame.new([%{a: 1, b: 2}])

df_b = Explorer.DataFrame.new([%{b: 3}])
df_b = Explorer.DataFrame.put(df_b, :a, Explorer.Series.from_list([4]))

dbg Explorer.DataFrame.concat_rows(df_a, df_b)
```

## Expected

(This is the result produced for `0.8.1`.)

```
[script.exs:8: (file)]
Explorer.DataFrame.concat_rows(df_a, df_b) #=> #Explorer.DataFrame<
  Polars[2 x 2]
  a s64 [1, 4]
  b s64 [2, 3]
>
```

## Actual

```
** (RuntimeError) Polars Error: lengths don't match: unable to vstack, column names don't match: "a" and "b"
    (explorer 0.8.2) lib/explorer/polars_backend/shared.ex:79: Explorer.PolarsBackend.Shared.apply_dataframe/4
    script.exs:8: (file)
```

philss:
@maennchen I can confirm this is a regression. Thank you!

In order to mitigate this before we launch a new version, please use the [`Explorer.DataFrame.relocate/3`](https://hexdocs.pm/explorer/Explorer.DataFrame.html#relocate/3) function as a workaround:

```elixir
 df_c = Explorer.DataFrame.relocate(df_b, "b", after: "a")

Explorer.DataFrame.concat_rows(df_a, df_c)
```

PR #920, opened 2024-06-07
Fix concat_rows with not aligned dataframes

philss:
Closes https://github.com/elixir-explorer/explorer/issues/902


philss:
This is a simple one, so I'm merging.

PR #924, opened 2024-06-08
Let `{:datetime, _}` be an alias for `{:naive_datetime, _}` for backwards compatibility

billylanchantin:
### Description

Allows `{:datetime, _}` as an alias for `{:naive_datetime, _}`. This will prevent the next release from being breaking.

### Extras

* Fix a typo the series moduledoc
* Fix a bug in the `query_to_lazy_series!` conversion fun


lib/explorer/shared.ex
```diff
@@ -87,6 +89,16 @@ defmodule Explorer.Shared do
   def normalise_dtype(:u32), do: {:u, 32}
   def normalise_dtype(:u64), do: {:u, 64}
 
+  def normalise_dtype({:datetime, precision}) do
+    :ok =
+      Logger.warning("""
+      The `{:datetime, _}` dtype has been deprecated.
+      Please use `{:naive_datetime, _}` instead.
+      """)
```
billylanchantin, line 97: I'm warning when we find `{:datetime, _}`. We can not do this as well.

lib/explorer/data_frame.ex
```diff
@@ -2865,8 +2865,8 @@ defmodule Explorer.DataFrame do
       naive_datetime = %NaiveDateTime{} ->
         LazySeries.new(:lazy, [naive_datetime], {:naive_datetime, :nanosecond})
 
-      datetime = %DateTime{} ->
-        LazySeries.new(:lazy, [datetime], {:datetime, :nanosecond})
+      datetime = %DateTime{time_zone: time_zone} ->
+        LazySeries.new(:lazy, [datetime], {:datetime, :nanosecond, time_zone})
```
billylanchantin, line 2869: I think this was a bug, right?.

Also, shouldn't the precision be `:microsecond`?

josevalim: I think it should be microsecond, yeah. We should have a test for this. It should also be microsecond above if possible.

lib/explorer/series.ex
```diff
@@ -30,9 +30,11 @@ defmodule Explorer.Series do
   and compatibility with the Elixir ecosystem:
 
     * All numeric dtypes (signed integer, unsigned integer, and floats) can
-      be specified as an atom in the form of `:s32`, `:u8`, `:f32` and o son
+      be specified as an atom in the form of `:s32`, `:u8`, `:f32` and so on
     * The atom `:float` as an alias for `{:f, 64}` to mirror Elixir's floats
     * The atom `:integer` as an alias for `{:s, 64}` to mirror Elixir's integers
+    * The tuple `{:datetime, _}` as an alias for `{:naive_datetime, _}`.
+      Note: `{:datetime, _}` has been deprecated and using it will raise a warning.
```
josevalim, line 37: I wouldn't mention this. It is deprecated and be removed next anyway.

```suggestion
```

josevalim (approved):

lib/explorer/data_frame.ex
billylanchantin: [c278300](https://github.com/elixir-explorer/explorer/pull/924/commits/c2783003764c33055c3c00faa0f8ab85a753ff9c)

josevalim (approved):
All good once CI passes!

Issue #922, opened 2024-06-08
Performance of `DataFrame.new/2` on dataframes containing list columns


spencerkent:
I'm noticing that creation of a new DataFrame that has a list column is _veeeerrry sloooow_ (in Explorer 0.8.2), so thought I'd show a little benchmark and see if there's any ideas how to make it faster :) 

Starting with two objects that support `Table.Reader`, one of them has no list columns, the other does (synthetic data):

```elixir
iex>IO.inspect(tr_object_no_list_columns)
[
  %{
    billing_code: "E1070",
    billing_code_type_version: "2024",
    billing_code_description: "rxndfe4t",
    negotiated_rate: 4348.933298145181
  },
  %{
    billing_code: "G9143",
    billing_code_type_version: "2024",
    billing_code_description: "3dc8r70lej2g",
    negotiated_rate: 3008.7836399549005
  },
  %{
    billing_code: "G0482",
    billing_code_type_version: "2024",
    billing_code_description: "bhq0y163",
    negotiated_rate: 7891.783042709335
  }
...
]
```

```elixir
iex>IO.inspect(tr_object_with_list_columns)
[
  %{
    billing_code: "G9143",
    billing_code_type_version: "2024",
    service_codes: ["57", "81", "50", "34", "32"],
    negotiated_rate: 4596.469175710615
  },
  %{
    billing_code: "A4220",
    billing_code_type_version: "2024",
    service_codes: ["71"],
    negotiated_rate: 2718.823985442407
  },
  %{
    billing_code: "A4220",
    billing_code_type_version: "2024",
    service_codes: ["62", "1", "56", "18", "23", "13", "14", "19"],
    negotiated_rate: 7437.797724472413
  }
...
]
```

These objects are essentially the same size:
```elixir
iex>IO.puts("The size of the object without a list column is: #{
  :erlang.term_to_binary(tr_object_no_list_columns) 
  |> :erlang.byte_size()) / 1_000_000} MB"
)
"The size of the object without a list column is: 0.14509 MB"
```
```elixir
iex>IO.puts("The size of the object *with* a list column is: #{
  :erlang.term_to_binary(tr_object_with_list_columns) 
  |> :erlang.byte_size()) / 1_000_000} MB"
)
"The size of the object *with* a list column is: 0.14586 MB"
```

The comparison:
```elixir
# manually-specifying the types, just in case
iex>dtypes = %{
  no_list: %{
    billing_code: :string,
    billing_code_type_version: :string,
[...]

PR #928, opened 2024-06-13
Stop inference of type if dtype is given and manually decode terms where needed

philss:
This is a continuation of https://github.com/elixir-explorer/explorer/pull/923

We are now delegating to the backend the job of decoding the terms instead of casting if the dtype is different from the inferred dtype. In fact we stopped to infer the type if the dtype is given.

In a nutshell, this PR:

1. removes the inference of type if `:dtype` is given.
2. stop casting if given dtype and inferred dtype are different, because point 1.
3. adds a manual decoding of terms for the functions that may accept special values (structs or integers).

The main goal is to improve performance, as discussed in #922.

### DIFF
--- a/lib/explorer/polars_backend/shared.ex
+++ b/lib/explorer/polars_backend/shared.ex
@@ -8,11 +8,12 @@ defmodule Explorer.PolarsBackend.Shared do
   alias Explorer.PolarsBackend.Native
   alias Explorer.PolarsBackend.Series, as: PolarsSeries
   alias Explorer.Series, as: Series
+  import Kernel, except: [apply: 2]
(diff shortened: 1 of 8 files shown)


josevalim (approved):

billylanchantin (approved):
This looks great!

Question: Do we think this test file is now obsolete?

* https://github.com/elixir-explorer/explorer/blob/main/test/explorer/series/inferred_dtype_property_test.exs

philss:
@billylanchantin Thank you! I don't think that test is obsolete because it is covering the inference algorithm, so I would keep that :D

PR #932, opened 2024-06-29
Remote series and dataframes and distributed GC

josevalim:
Automatically transfer data between nodes for remote series
and dataframes and perform distributed garbage collection.

The functions in `Explorer.DataFrame` and `Explorer.Series`
will automatically move operations on remote dataframes to
the nodes they belong to. This module provides additional
conveniences for manual placement.

## Implementation details

There is a new module called `Explorer.Remote`.
In order to understand what it does, we need
to understand the challenges in working with remote series
and dataframes.

Series and dataframes are actually NIF resources: they are
pointers to blobs of memory operated by low-level libraries.
Those are represented in Erlang/Elixir as references (the
same as the one returned by `make_ref/0`). Once the reference
is garbage collected (based on refcounting), those NIF
resources are garbage collected and the memory is reclaimed.

When using Distributed Erlang, you may write this code:

    remote_series = :erpc.call(node, Explorer.Series, :from_list, [[1, 2, 3]])

However, the code above will not work, because the series
will be allocated in the remote node and the remote node
won't hold a reference to said series! This means the series
is garbage collected and if we attempt to read it later on,
from the caller node, it will no longer exist. Therefore,
we must explicitly place these resources in remote nodes
by spawning processes to hold these refernces. That's what
the `place/2` function in this module does.

We also need to guarantee these resources are not kept
forever by these remote nodes, so `place/2` creates a
local NIF resource that notifies the remote resources
they have been GCed, effectively implementing a remote
garbage collector.

## TODO

* [x] Make `collect` in dataframe transfer to the current node
* [x] Add `collect` to series
* [x] Add `compute` to dataframe
* [x] Handle dataframes (remove Shared.apply_impl)
* [x] Handle lazy series
* [x] Add `node` option to creation functions

### DIFF
--- a/lib/explorer/remote.ex
+++ b/lib/explorer/remote.ex
@@ -135,15 +135,17 @@ defmodule Explorer.Remote do
   # If a pid is given, we assume the pid will hold those references
   # and avoid spawning new ones. If a node is given, regular placement
   # occurs.
-  def apply(pid, mod, fun, resources, args_callback) when is_pid(pid) do
-    apply(node(pid), mod, fun, resources, args_callback, &place_on_pid(&1, pid))
+  def apply(pid, mod, fun, resources, args_callback, place?) when is_pid(pid) do
+    placing_callback = if place?, do: &place_on_pid(&1, pid), else: &{&1, []}
+    do_apply(node(pid), mod, fun, resources, args_callback, placing_callback)
   end
 
-  def apply(node, mod, fun, resources, args_callback) when is_atom(node) do
-    apply(node, mod, fun, resources, args_callback, &place_on_node(&1, []))
+  def apply(node, mod, fun, resources, args_callback, place?) when is_atom(node) do
+    placing_callback = if place?, do: &place_on_node(&1, []), else: &{&1, []}
+    do_apply(node, mod, fun, resources, args_callback, placing_callback)
   end
 
-  defp apply(node, mod, fun, resources, args_callback, placing_function) do
+  defp do_apply(node, mod, fun, resources, args_callback, placing_callback) do
     resources =
       Enum.map(resources, fn
         {resource_data, resource_node} when resource_node == node or resource_node == nil ->
@@ -161,7 +163,7 @@ defmodule Explorer.Remote do
     receive do
       {^message_ref, result} ->
         Process.demonitor(monitor_ref, [:flush])
-        {term, _acc} = placing_function.(result)
+        {term, _acc} = placing_callback.(result)
         send(child, {message_ref, :ok})
         term
 
--- a/lib/explorer/remote.ex
+++ b/lib/explorer/remote.ex
@@ -1,20 +1,44 @@
 defmodule Explorer.Remote do
   @moduledoc """
-  A module responsible for tracking remote dataframes
-  and garbage collect them.
+  A module responsible for placing remote dataframes and
+  garbage collect them.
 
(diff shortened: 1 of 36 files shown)


billylanchantin (commented):
_Very_ cool stuff! 🤯

So far I've only had one question. I'll probably have more after I read through again.

lib/explorer/remote.ex
```diff
@@ -0,0 +1,197 @@
...
+  """
+
+  @doc """
+  Receives a data structure and traverses it looking
+  for remote dataframes and series.
+
+  If any is found, it spawns a process on the remote node
+  and sets up a distributed garbage collector. This function
+  only traverses maps, lists, and tuples, it does not support
+  arbitrary structs (such as map sets).
```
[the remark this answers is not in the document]
josevalim: There are two reasons why we do a traversal:

1. Convenience: it is easier if the rest of the Explorer Series/DataFrame API do not have to care about explicitly annotating what needs to be placed.

2. FLAME support: we want to integrate this with FLAME, allowing developers to run custom code within a FLAME and that may return series/dataframes which are arbitrarily nested.

jonatanklosko:
Amazing!! 🔥🐈‍⬛

lib/explorer/remote/holder.ex
```diff
@@ -0,0 +1,71 @@
...
+    refs = pop_from_list(state.refs, ref, pid)
+    pids = pop_from_list(state.pids, pid, ref)
+    noreply_or_stop(%{state | pids: pids, refs: refs})
+  end
+
+  def handle_info({:DOWN, owner_ref, _, _, reason}, %{owner_ref: owner_ref} = state) do
+    {:stop, reason, state}
+  end
+
+  def handle_info({:DOWN, _, _, pid, _}, state) do
```
jonatanklosko, line 50: Shouldn't we monitor on `hold` to receive this message?

lib/explorer/remote/holder.ex
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  @impl true
+  def handle_info({:gc, ref, pid}, state) do
+    refs = pop_from_list(state.refs, ref, pid)
+    pids = pop_from_list(state.pids, pid, ref)
+    noreply_or_stop(%{state | pids: pids, refs: refs})
+  end
+
+  def handle_info({:DOWN, owner_ref, _, _, reason}, %{owner_ref: owner_ref} = state) do
```
jonatanklosko, line 46: If the idea is to allow multiple nodes to call `hold`, why do we terminate eagerly when the owner goes down, given that we already do cleanup in the other `:DOWN` chandler?

lib/explorer/remote/holder.ex
josevalim: Yes, we should! I will try to add a test.

lib/explorer/remote/holder.ex
josevalim: We need to do this if there are no pids or no refs, I will change accordingly.

jonatanklosko: Yeah, but it should be enough to monitor on init and rely on the other clause, which wouldn't pop anything and would cause stop, no?

jonatanklosko: (the difference being, that if the initial node goes down and there are other nodes referencing, we keep the holder alive)

PR #934, opened 2024-06-30
Implement message-on-gc functionality in Rust

filmor:
@cocoa-xu, @josevalim  FYI

I'm not happy with the necessity to use two mutexes here and will see, whether `SavedTerm` can be improved.


josevalim:
Thank you @filmor! Do you think the change is worth it, given the C version does not have to deal with mutexes? Or it does have to deal with them, it is just that we cannot see them? 🤔

josevalim:
Btw, this was closed automatically by GitHub because I merged the other PR (and the branch was automatically removed). Please do submit it against main (and sorry for the inconvenience!).

filmor:
I'll resubmit. It's not quite there, yet, anyhow because `enif_send` is so ... weirdly specific about its `env` argument.

The C variant doesn't have to deal with mutexes because it is inherently unsafe. This is irrelevant for this particular resource type, as it is never actually "used" and we can trust the destructor to be only called exactly once. I can do the same in Rust by adding an `unsafe impl Sync for LocalMessage {}`.

The main reason I am looking into this is that I'd like "all" patterns to be implementable directly in Rust, everything that requires dropping down to C (or even `unsafe`) is a bug to me.

PR #936, opened 2024-06-30
Implement message-on-gc NIFs in Rust

filmor:
- Pure Rust implementation of `LocalMessage`
- Bump Rustler to 0.34.0
- Drop explicit NIF list and `on_load` callback as these are now handled by Rustler implicitly


filmor:
@josevalim The issue is actually the missing `env` parameter on the destructor. If it's okay with everyone here, I'd keep this PR open until this is fixed upstream in Rustler.

josevalim:
No issues with keeping it open!

filmor:
This not only implements the function in pure Rust (thus making explorer pure Rust again), it also makes use of other niceties to trim down `lib.rs` quite massively. If you would like to limit the impact of this PR to the destructor addition, let me know.

native/explorer/src/local_message.rs
```diff
@@ -1,29 +1,48 @@
...
-            env,
-            local_message_on_gc(env.as_c_arg(), pid.as_c_arg(), term.as_c_arg()),
-        )
+impl LocalMessage {
+    pub fn new(pid: LocalPid, term: Term<'_>) -> Self {
+        let env = OwnedEnv::new();
+        let term = env.save(term);
+        Self {
+            pid,
+            env_and_term: Mutex::new((env, term)),
```
josevalim, line 17: Is there an easy way to get the size of a data structure in Rust? My only concern is how much memory this uses, I am not very worried about the performance aspect since, as you said, it is never locked.

filmor: There is a sizeof. But I think, in this particular case we should be able to get away with just `unsafe impl Sync` on the type.

josevalim: It depends on the size. I assume it is small compared to the env and term? But if it is order of magnitude larger, then I would go with unsafe, as we are sure it is safe in other ways.

filmor: I think it's a `pthread` mutex, so ~40 bytes. Env and term are together probably around the same size. It's fine, I changed it to the `unsafe impl` now, much simpler anyway.

josevalim:

PR #942, opened 2024-07-12
Update Rust toolchain to nightly-2024-06-23

philss:
This is an attempt to fix the precomp build.


philss:
The build is passing. But `macos-11` was removed from GH Actions, so I will update to `macos-12` in a new PR.

Issue #953, opened 2024-07-31
Orders of names after CSV loading seems buggy since 9.0


clm-a:
Hello ! 

When I load  [those data](https://s3.eu-west-1.amazonaws.com/course.oc-static.com/projects/Data_Scientist_P4/2016_Building_Energy_Benchmarking.csv) from CSV, I get the col names in an unexpected order :

```ex

columns = ["OSEBuildingID", "DataYear", "BuildingType", "PrimaryPropertyType", "PropertyName", "Address", "City", "State", "ZipCode", "TaxParcelIdentificationNumber", "CouncilDistrictCode", "Neighborhood", "Latitude", "Longitude", "YearBuilt", "NumberofBuildings", "NumberofFloors", "PropertyGFATotal", "PropertyGFAParking", "PropertyGFABuilding(s)", "ListOfAllPropertyUseTypes", "LargestPropertyUseType", "LargestPropertyUseTypeGFA", "SecondLargestPropertyUseType", "SecondLargestPropertyUseTypeGFA", "ThirdLargestPropertyUseType", "ThirdLargestPropertyUseTypeGFA", "YearsENERGYSTARCertified", "ENERGYSTARScore", "SiteEUI(kBtu/sf)", "SiteEUIWN(kBtu/sf)", "SourceEUI(kBtu/sf)", "SourceEUIWN(kBtu/sf)", "SiteEnergyUse(kBtu)", "SiteEnergyUseWN(kBtu)", "SteamUse(kBtu)", "Electricity(kWh)", "Electricity(kBtu)", "NaturalGas(therms)", "NaturalGas(kBtu)", "DefaultData", "Comments", "ComplianceStatus", "Outlier", "TotalGHGEmissions", "GHGEmissionsIntensity"]
dtypes = Enum.map(columns, &{&1, :string})

df =
  DF.load_csv!(File.read!(Kino.FS.file_path("2016_Building_Energy_Benchmarking.csv")),
    delimiter: ",",
    dtypes: dtypes
  )
```

output: 
```
#Explorer.DataFrame<
  Polars[3376 x 46]
  Neighborhood string ["1", "2", "3", "5", "8", ...]
  TotalGHGEmissions string ["2016", "2016", "2016", "2016", "2016", ...]
  Electricity(kBtu) string ["NonResidential", "NonResidential", "NonResidential", "NonResidential",
   "NonResidential", ...]
.....
```
[...]

PR #955, opened 2024-08-01
Fix overwrite of dtypes for `DF.load_csv/2`

philss:
Using the `with_schema` function does not work correctly because Polars expects to receive all column names, and with the correct order. We are passing a map to the backend, and then we transform that to a list before calling the Rust code. So the schema could be out-of-order.

We are now using `with_schema_overwrite` for both `load_csv` and `from_csv`.

Closes https://github.com/elixir-explorer/explorer/issues/953

### DIFF
--- a/native/explorer/src/dataframe/io.rs
+++ b/native/explorer/src/dataframe/io.rs
@@ -47,7 +47,13 @@ pub fn df_from_csv(
         _ => CsvEncoding::Utf8,
     };
 
-    let dataframe = CsvReadOptions::default()
+    let read_options = if dtypes.is_empty() {
+        CsvReadOptions::default()
+    } else {
+        CsvReadOptions::default().with_schema_overwrite(Some(schema_from_dtypes_pairs(dtypes)?))
+    };
+
+    let dataframe = read_options
         .with_infer_schema_length(infer_schema_length)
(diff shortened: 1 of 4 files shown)


native/explorer/src/dataframe/io.rs
```diff
@@ -47,7 +47,13 @@ pub fn df_from_csv(
         _ => CsvEncoding::Utf8,
     };
 
-    let dataframe = CsvReadOptions::default()
+    let read_options = if dtypes.is_empty() {
+        CsvReadOptions::default()
+    } else {
+        CsvReadOptions::default().with_schema_overwrite(Some(schema_from_dtypes_pairs(dtypes)?))
```
josevalim, line 53: Maybe instead of a conditional, we could have `schema_from_dtypes_pairs` return Some/None accordingly, and we always call with_schema_overwrite?

josevalim (approved):

josevalim (approved):

native/explorer/src/dataframe/io.rs
philss: Good call! I changed in https://github.com/elixir-explorer/explorer/pull/955/commits/846017001330ccd05852a012c2d08c8607dba70d

PR #966, opened 2024-08-18
Add `keep: :none` argument to mutate functions

### LINKED ISSUE #965: Support dplyr-like `keep` argument for `mutate` functions
It would be nice to have the `keep` argument on `mutate` functions that [Dplyr](https://dplyr.tidyverse.org/reference/mutate.html#arguments) has. From their docs:

> .keep
> Control which columns from .data are retained in the output. Grouping columns and columns created by ... are always kept.
> 
> "all" retains all columns from .data. This is the default.
> 
> "used" retains only the columns used in ... to create new columns. This is useful for checking your work, as it displays inputs and outputs side-by-side.
> 
> "unused" retains only the columns not used in ... to create new columns. This is useful if you generate new columns, but no longer need the columns used to generate them.
> 
> "none" doesn't retain any extra columns from .data. Only the grouping variables and columns created by ... are kept.

Using myself as an example, I often find I need to create temporary working columns that I need to break up the computations to keep it readable. In those cases, `keep: :none` or `keep: :used` options really help. Especially in Livebook mode where readability trumps performance.

I am currently chaining `mutate` with `select` and it's fine, but it's a bit of work to keep them aligned with the number of mutations gets large. Being able to avoid this would be nice.

brendon9x:
This PR adds a new `keep` option to the `mutate` and `mutate_with` functions in the DataFrame module. This option allows users to control which columns are retained in the output DataFrame after a mutation operation. Closes #965 

## Examples of usage

Regular DataFrame:

```elixir
df = Explorer.DataFrame.new(a: ["a", "b", "c"], b: [1, 2, 3])
Explorer.DataFrame.mutate(df, [c: b + 1], keep: :none)
#Explorer.DataFrame<
  Polars[3 x 1]
  c s64 [2, 3, 4]
>
```

Grouped DataFrame:

```elixir
df = Explorer.Datasets.iris()
grouped = Explorer.DataFrame.group_by(df, "species")
Explorer.DataFrame.mutate(grouped, [petal_length_avg: mean(petal_length)], keep: :none)
#Explorer.DataFrame<
  Polars[150 x 2]
  Groups: ["species"]
  species string ["Iris-setosa", "Iris-setosa", "Iris-setosa", ...]
  petal_length_avg f64 [1.464, 1.464, 1.464, ...]
>
```

## Considerations

1. I decided to use a keyword argument (`:all` or `:none`) instead of a boolean for maximum forwards compatibility.
2. I was going to add unit tests, but the doctests seemed sufficient to cover the new functionality and its interactions with existing features.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2661,12 +2661,12 @@ defmodule Explorer.DataFrame do
 
   ## Options
 
-  * `:keep` – Controls which columns from the input dataframe are retained
-    in the output.
-    Possible values are:
-    - `:all`: Retains all columns from the input dataframe (default).
-    - `:none`: Retains only the columns created by the mutation (and
-               grouping columns if operating on a grouped DataFrame).
+    * `keep` - Controls which columns should be kept in the resulting dataframe.
+      Supported options for keep are:
+
+      * `:all` - Retains all columns from the input dataframe (default).
+      * `:none` - Retains only the create columns created by the mutation (and
+                  grouping columns if operating on a grouped DataFrame).
 
   > #### Notice {: .notice}
   >
@@ -2809,12 +2809,14 @@ defmodule Explorer.DataFrame do
   values using the `Explorer.Series` API directly and then add
   it to the dataframe using `put/3`.
 
-  An optional `keep` argument can be passed to control which columns
-  should be kept in the resulting dataframe. The valid options
-  for keep are:
-  * `:all` - Retains all columns from the input dataframe (default).
-  * `:none` - Retains only the create columns created by the mutation (and
-              grouping columns if operating on a grouped DataFrame).
+  ## Options
+
+    * `keep` - Controls which columns should be kept in the resulting dataframe.
+      Supported options for keep are:
+
+      * `:all` - Retains all columns from the input dataframe (default).
+      * `:none` - Retains only the create columns created by the mutation (and
+                  grouping columns if operating on a grouped DataFrame).
 
   ## Examples
 
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2815,7 +2815,7 @@ defmodule Explorer.DataFrame do
       Supported options for keep are:
 
       * `:all` - Retains all columns from the input dataframe (default).
-      * `:none` - Retains only the create columns created by the mutation (and
+      * `:none` - Retains only the columns created by the mutation (and
                   grouping columns if operating on a grouped DataFrame).
 
   ## Examples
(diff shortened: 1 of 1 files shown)


lib/explorer/data_frame.ex
```diff
@@ -2747,27 +2770,52 @@ defmodule Explorer.DataFrame do
...
   hold any values, instead it stores all operations in order to
   execute all mutations performantly.
 
-  This is a callback version of `mutate/2`. If your mutation
+  This is a callback version of `mutate/3`. If your mutation
   cannot be expressed with lazy dataframes, you may compute the
   values using the `Explorer.Series` API directly and then add
   it to the dataframe using `put/3`.
 
+  An optional `keep` argument can be passed to control which columns
```
josevalim, line 2812: Let's use the same `## Options` block here as well, WDYT?

lib/explorer/data_frame.ex
```diff
@@ -2823,13 +2871,62 @@ defmodule Explorer.DataFrame do
...
+  Note that `keep: :all` is the default behavior.
+
+  Using the `:keep` option to retain only new columns:
+
+        iex> df = Explorer.DataFrame.new(a: [4, 5, 6], b: [1, 2, 3])
+        iex> Explorer.DataFrame.mutate_with(df, &[c: Explorer.Series.add(&1["a"], &1["b"])], keep: :none)
+        #Explorer.DataFrame<
+          Polars[3 x 1]
+          c s64 [5, 7, 9]
+        >
```
josevalim, line 2897: We only need to indent with four spaces.

josevalim (approved):

lib/explorer/data_frame.ex
```diff
@@ -2654,11 +2654,25 @@ defmodule Explorer.DataFrame do
...
+    * `keep` - Controls which columns should be kept in the resulting dataframe.
+      Supported options for keep are:
+
+      * `:all` - Retains all columns from the input dataframe (default).
+      * `:none` - Retains only the columns created by the mutation (and
+                  grouping columns if operating on a grouped DataFrame).
+
+  > #### Notice {: .notice}
+  >
+  > You will need to use the `[]` to enclose your mutations in order
```
cigrainger, line 2673:
```suggestion
  > You will need to wrap your mutations in `[]` in order
```

cigrainger (approved):
I'm happy with this. IIUC it's 100% opt-in (doesn't change the current API) and provides a nice bit of sugar to avoid an additional select after mutate.

josevalim:

brendon9x:
@cigrainger @josevalim – this repo was a joy to work in. The contributor docs were great and the mix tasks worked flawlessly (and tests are super fast). I wouldn't have been able to push this over the line if any of that wasn't true. Thanks for the awesome project.

PR #975, opened 2024-08-30
Fix DataFrame.filter_with spec

mhanberg:
I believe since the callback can return a list of lazy series queries

If this isn't correct, feel free to close.

Thanks!

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2606,7 +2606,7 @@ defmodule Explorer.DataFrame do
   @doc type: :single
   @spec filter_with(
           df :: DataFrame.t(),
-          callback :: (Explorer.Backend.LazyFrame.t() -> [Series.lazy_t()])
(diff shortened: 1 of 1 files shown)


lib/explorer/data_frame.ex
```diff
@@ -2606,7 +2606,7 @@ defmodule Explorer.DataFrame do
   @doc type: :single
   @spec filter_with(
           df :: DataFrame.t(),
-          callback :: (Explorer.Backend.LazyFrame.t() -> Series.lazy_t())
+          callback :: (Explorer.Backend.LazyFrame.t() -> [Series.lazy_t()])
```
josevalim, line 2609:
```suggestion
          callback :: (Explorer.Backend.LazyFrame.t() -> Series.lazy_t() | [Series.lazy_t()])
```

mhanberg: Oh duh, good catch

josevalim:

Issue #976, opened 2024-09-03
Panic for list of lists with `nil` entries


maennchen:
## Code

```elixir
Explorer.DataFrame.new(
  [
    %{addresses: nil},
    %{
      addresses: [
        %{lines: ["Nadarzyn Mszczonowska 24"]},
        %{lines: ["Nadarzyn Mszczonowska 24"]}
      ]
    },
    %{addresses: [%{lines: nil}, %{lines: nil}]}
  ],
  dtypes: [{"addresses", {:list, {:struct, [{"lines", {:list, :string}}]}}}]
)
```

## Expected

Either an error explaining that the lists can't be `nil` or the following:

```elixir
#Explorer.DataFrame<
  Polars[3 x 1]
  addresses list[struct[1]] [
    [],
    [%{"lines" => ["Nadarzyn Mszczonowska 24"]}, %{"lines" => [...]}],
    [%{"lines" => []}, %{lines: []}]
  ]
>
```

## Actual

```
thread '<unnamed>' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-core-0.42.0/src/named_from.rs:170:56:
called `Result::unwrap()` on an `Err` value: ComputeError(ErrString("dtypes don't match, got struct[1], expected: struct[1]"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
** (ArgumentError) cannot create series "addresses": Erlang error: :nif_panicked
    (explorer 0.9.2) lib/explorer/polars_backend/data_frame.ex:578: Explorer.PolarsBackend.DataFrame.series_from_list!/3
    (explorer 0.9.2) lib/explorer/polars_backend/data_frame.ex:529: anonymous fn/3 in Explorer.PolarsBackend.DataFrame.from_tabular/2
    (elixir 1.17.2) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
    (explorer 0.9.2) lib/explorer/polars_backend/data_frame.ex:523: Explorer.PolarsBackend.DataFrame.from_tabular/2
    temp/trigger.exs:35: (file)
```

 ## Runtime

* Elixir 1.17.2
* Erlang 27.0.1
* Explorer 0.9.2

Potentially related to #857

billylanchantin:
Thanks for the report, @maennchen! We just merged a fix for this issue to `main`. Feel free to test it out now or wait until the next release.

PR #980, opened 2024-09-07
Fix issue 976

billylanchantin:
### Description

Fixes an issue with `DF.new` where a well placed `nil` could result in a panic.

Basically, the work was being done by `Native.s_from_list_of_series/2` which attempted to do following:

* create a list of individual `polars_core::series`s each with the same `dtype`
* create a final `polars_core::series` of type `{:list, dtype}` from that list

But because `s_from_list_of_series/2` didn't have the final dtype information up front, it was possible for two elements of the intermediate list to have incompatible dtypes. When this happened, the final creation would panic.

### Changes

This PR makes the dtype a required argument for:

* `s_from_list_of_series/3`
* `s_from_list_of_series_as_structs/3`

(Those two seemed like a pair so I made them the same.)

### Links

* Closes: https://github.com/elixir-explorer/explorer/issues/976

### Discussion

I think there's an opportunity for optimization since we're now passing the expected dtype, but I didn't explore it.

### DIFF
--- a/lib/explorer/polars_backend/native.ex
+++ b/lib/explorer/polars_backend/native.ex
@@ -363,7 +363,7 @@ defmodule Explorer.PolarsBackend.Native do
   def s_from_list_str(_name, _val), do: err()
   def s_from_list_binary(_name, _val), do: err()
   def s_from_list_categories(_name, _val), do: err()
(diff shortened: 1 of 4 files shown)


josevalim (approved):

philss (approved):
🚀

PR #988, opened 2024-09-20
Add support for some operations with decimals

philss:
Adding support for decimals in some of functions that work with integers and floats. This is adding support for creating decimal series using the `Explorer.Series.from_list/2` function as well.

The decimals support is still in the experimental phase in Polars, so we won't be able to support all the operations now. Some of the operations are returning `f64` series or float results, and some of the functions are raising exceptions from Polars because they are not implemented at the backend.

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -26,6 +26,8 @@ defmodule Explorer.Series do
     * `{:f, size}` - a 64-bit or 32-bit floating point number
     * `{:s, size}` - a 8-bit or 16-bit or 32-bit or 64-bit signed integer number.
     * `{:u, size}` - a 8-bit or 16-bit or 32-bit or 64-bit unsigned integer number.
+    * `{:decimal, precision, scale}` - a 128-bit signed integer number representing a decimal,
+      with a scale and precision. This unwraps to `Decimal`, using the `:decimal` package.
     * `:null` - `nil`s exclusively
     * `:string` - UTF-8 encoded binary
     * `:time` - Time type that unwraps to `Elixir.Time`
@@ -38,10 +40,11 @@ defmodule Explorer.Series do
   When passing a dtype as argument, aliases are supported for convenience
   and compatibility with the Elixir ecosystem:
 
-    * All numeric dtypes (signed integer, unsigned integer, and floats) can
-      be specified as an atom in the form of `:s32`, `:u8`, `:f32` and so on
+    * All numeric dtypes (signed integer, unsigned integer, floats and decimals) can
+      be specified as an atom in the form of `:s32`, `:u8`, `:f32` and so on.
     * The atom `:float` as an alias for `{:f, 64}` to mirror Elixir's floats
     * The atom `:integer` as an alias for `{:s, 64}` to mirror Elixir's integers
+    * The atom `:decimal` as an alias for the `{:decimal, nil, nil}`.
 
   A series must consist of a single data type only. Series may have `nil` values in them.
   The series `dtype` can be retrieved via the `dtype/1` function or directly accessed as
@@ -140,7 +143,8 @@ defmodule Explorer.Series do
   @numeric_dtypes Explorer.Shared.numeric_types()
   @numeric_or_temporal_dtypes @numeric_dtypes ++ @temporal_dtypes
 
-  @io_dtypes Shared.dtypes() -- [:binary, :string, {:list, :any}, {:struct, :any}]
+  @io_dtypes Shared.dtypes() --
+               [:binary, :string, {:list, :any}, {:struct, :any}, {:decimal, :any, :any}]
 
   @type dtype ::
           :null
@@ -150,11 +154,12 @@ defmodule Explorer.Series do
           | :date
           | :time
           | :string
-          | naive_datetime_dtype
           | datetime_dtype
+          | decimal_dtype
           | duration_dtype
           | float_dtype
           | list_dtype
+          | naive_datetime_dtype
           | signed_integer_dtype
           | struct_dtype
           | unsigned_integer_dtype
@@ -170,10 +175,12 @@ defmodule Explorer.Series do
   @type signed_integer_dtype :: {:s, 8} | {:s, 16} | {:s, 32} | {:s, 64}
   @type unsigned_integer_dtype :: {:u, 8} | {:u, 16} | {:u, 32} | {:u, 64}
   @type float_dtype :: {:f, 32} | {:f, 64}
+  @type decimal_dtype :: {:decimal, nil | pos_integer(), nil | pos_integer()}
 
-  @type dtype_alias :: integer_dtype_alias | float_dtype_alias
+  @type dtype_alias :: integer_dtype_alias | float_dtype_alias | decimal_dtype_alias
   @type float_dtype_alias :: :float | :f32 | :f64
   @type integer_dtype_alias :: :integer | :u8 | :u16 | :u32 | :u64 | :s8 | :s16 | :s32 | :s64
+  @type decimal_dtype_alias :: :decimal
 
   @type t :: %Series{data: Explorer.Backend.Series.t(), dtype: dtype()}
   @type lazy_t :: %Series{data: Explorer.Backend.LazySeries.t(), dtype: dtype()}
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -149,7 +149,7 @@ defmodule Explorer.Series do
                  :string,
                  {:list, :any},
                  {:struct, :any},
-                 {:decimal, :nil_or_pos_integer, :pos_integer}
+                 {:decimal, :pos_integer, :pos_integer}
                ]
 
   @type dtype ::
@@ -181,12 +181,12 @@ defmodule Explorer.Series do
   @type signed_integer_dtype :: {:s, 8} | {:s, 16} | {:s, 32} | {:s, 64}
   @type unsigned_integer_dtype :: {:u, 8} | {:u, 16} | {:u, 32} | {:u, 64}
   @type float_dtype :: {:f, 32} | {:f, 64}
-  @type decimal_dtype :: {:decimal, nil | pos_integer(), pos_integer()}
+  @type decimal_dtype :: {:decimal, pos_integer(), pos_integer()}
 
   @type dtype_alias :: integer_dtype_alias | float_dtype_alias | decimal_dtype_alias
   @type float_dtype_alias :: :float | :f32 | :f64
   @type integer_dtype_alias :: :integer | :u8 | :u16 | :u32 | :u64 | :s8 | :s16 | :s32 | :s64
-  @type decimal_dtype_alias :: :decimal | :d0 | :d1 | :d2 | :d3 | :d4 | :d5
+  @type decimal_dtype_alias :: :decimal
 
   @type t :: %Series{data: Explorer.Backend.Series.t(), dtype: dtype()}
   @type lazy_t :: %Series{data: Explorer.Backend.LazySeries.t(), dtype: dtype()}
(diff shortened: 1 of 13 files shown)


lib/explorer/shared.ex
```diff
@@ -99,8 +100,17 @@ defmodule Explorer.Shared do
...
+      when is_integer(scale) and (is_nil(precision) or is_integer(precision)),
+      do: dtype
+
+  def normalise_dtype(:decimal), do: {:decimal, nil, 0}
+  def normalise_dtype(:d0), do: {:decimal, nil, 0}
+  def normalise_dtype(:d1), do: {:decimal, nil, 1}
+  def normalise_dtype(:d2), do: {:decimal, nil, 2}
+  def normalise_dtype(:d3), do: {:decimal, nil, 3}
+  def normalise_dtype(:d4), do: {:decimal, nil, 4}
+  def normalise_dtype(:d5), do: {:decimal, nil, 5}
```
billylanchantin, line 113:
```suggestion
  for scale <- 0..38 do
    def normalise_dtype(unquote(:"d#{scale}")), do: {:decimal, nil, unquote(scale)}
  end
```

If we wanted to go all the way up to 38. If not pls ignore :)

billylanchantin (commented):
Great job! This looks like it was a ton of work. And I think the tradeoffs (like the not implemented errors) are very reasonable.

lib/explorer/series.ex
```diff
@@ -26,6 +26,8 @@ defmodule Explorer.Series do
     * `{:f, size}` - a 64-bit or 32-bit floating point number
     * `{:s, size}` - a 8-bit or 16-bit or 32-bit or 64-bit signed integer number.
     * `{:u, size}` - a 8-bit or 16-bit or 32-bit or 64-bit unsigned integer number.
+    * `{:decimal, precision, scale}` - a 128-bit signed integer number representing a decimal,
+      with a scale and precision. This unwraps to `Decimal`, using the `:decimal` package.
```
josevalim, line 30: What does it mean to have nil precision or nil scale?

lib/explorer/series.ex
```diff
@@ -2577,6 +2601,7 @@ defmodule Explorer.Series do
 
     * floats: #{Shared.inspect_dtypes(@float_dtypes, backsticks: true)}
     * integers: #{Shared.inspect_dtypes(@integer_types, backsticks: true)}
```
josevalim, line 2603: Btw, there is a typo here, it should be `backticks` but we fix it later. :D

lib/explorer/series.ex
```diff
@@ -2577,6 +2601,7 @@ defmodule Explorer.Series do
...
+    * `:decimal`. The result will be a float.
```
josevalim, line 2604: For consistent, it should be the following, no?

```suggestion
    * decimals: the result will be a float
```

We need to mirror it below.

lib/explorer/shared.ex
```diff
@@ -366,6 +387,14 @@ defmodule Explorer.Shared do
   def merge_numeric_dtype({:f, left}, {:f, right}), do: {:f, max(left, right)}
   def merge_numeric_dtype({:f, _} = float, :null), do: float
   def merge_numeric_dtype(:null, {:f, _} = float), do: float
+
+  def merge_numeric_dtype({:decimal, _, _} = decimal, :null), do: decimal
+  def merge_numeric_dtype(:null, {:decimal, _, _} = decimal), do: decimal
+
+  # For now, float has priority over decimals due to Polars.
```
josevalim, line 394: I think this makes sense in general. Float is imprecise, so you by definition go imprecise once you mix precise (decimal) and imprecise (float).

lib/explorer/series.ex
```diff
@@ -3395,8 +3427,18 @@ defmodule Explorer.Series do
   defp cast_to_add({:datetime, p, tz}, {:duration, p}), do: {:datetime, p, tz}
   defp cast_to_add({:duration, p}, {:datetime, p, tz}), do: {:datetime, p, tz}
   defp cast_to_add({:duration, p}, {:duration, p}), do: {:duration, p}
+
+  defp cast_to_add({:decimal, p1, s1}, {:decimal, p2, s2}),
+    do: {:decimal, maybe_max(p1, p2), maybe_max(s1, s2)}
```
josevalim, line 3432: Is there some rule we need to follow here? For example, is there a maximum value for precision and scale?

josevalim: According to ChatGPT:

---

Yes, in Apache Arrow's decimal128 type, the precision and scale are constrained as follows:

    Precision: This represents the total number of digits that can be stored, both before and after the decimal point. For decimal128, the maximum precision is 38 digits. This means that it can store up to 38 significant digits.

    Scale: This defines how many of the digits are allocated to the fractional part (i.e., after the decimal point). The scale can be any value between 0 and the precision value. For example, if you have a precision of 38 and set a scale of 10, then 28 digits can be used before the decimal point and 10 digits after.

Thus, the maximum precision is 38, and the scale can be anywhere from 0 to 38, depending on the application needs.

--- 

So I think we are good, but I'd encapsulate this logic in a function. :)

lib/explorer/shared.ex
```diff
@@ -304,6 +324,7 @@ defmodule Explorer.Shared do
   defp infer_type(%DateTime{time_zone: tz} = _item), do: {:datetime, :microsecond, tz}
   defp infer_type(%NaiveDateTime{} = _item), do: {:naive_datetime, :microsecond}
   defp infer_type(%Explorer.Duration{precision: precision} = _item), do: {:duration, precision}
+  defp infer_type(%Decimal{} = item), do: {:decimal, nil, Decimal.scale(item)}
```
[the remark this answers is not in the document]
billylanchantin: Another option is defaulting to the max (38). This is like when we default to `:s64` from a list of integers: it's the most permissive.

lib/explorer/series.ex
philss: This is not clear reading Polars' code base. But seems that:
 - precision is ignored most of the time. The only case when I see it appearing is when casting a lazy series to a decimal without a given precision (nil) - it uses the maximum value of 38. But if it's present, it will remove all numbers that are above that precision. If this is `nil`, then the maximum precision is used.
 - scale is optional, but always defaults to zero. So I think the correct default value could be zero.

josevalim: It makes sense. Since it is decimal128, the precision is always 38, then it is a matter of deciding of where to put the dot (the scale). So I think we can default everywhere to 38 instead of `nil`. WDYT?

philss: OK, makes sense! I will change that. I will do the same for the `inter_type/1` function as well, OK? https://github.com/elixir-explorer/explorer/pull/988#discussion_r1795999861

josevalim: Yes, let's try defaulting to 38 everywhere instead of computing it.

lib/explorer/shared.ex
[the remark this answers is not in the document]
philss: We are now using the max precision as default.

lib/explorer/series.ex
```diff
@@ -140,7 +143,14 @@ defmodule Explorer.Series do
...
   @numeric_or_temporal_dtypes @numeric_dtypes ++ @temporal_dtypes
 
-  @io_dtypes Shared.dtypes() -- [:binary, :string, {:list, :any}, {:struct, :any}]
+  @io_dtypes Shared.dtypes() --
+               [
+                 :binary,
+                 :string,
+                 {:list, :any},
+                 {:struct, :any},
+                 {:decimal, :nil_or_pos_integer, :pos_integer}
```
josevalim, line 152:
```suggestion
                 {:decimal, :pos_integer, :pos_integer}
```

lib/explorer/series.ex
```diff
@@ -170,10 +181,12 @@ defmodule Explorer.Series do
   @type signed_integer_dtype :: {:s, 8} | {:s, 16} | {:s, 32} | {:s, 64}
   @type unsigned_integer_dtype :: {:u, 8} | {:u, 16} | {:u, 32} | {:u, 64}
   @type float_dtype :: {:f, 32} | {:f, 64}
+  @type decimal_dtype :: {:decimal, nil | pos_integer(), pos_integer()}
```
josevalim, line 184:
```suggestion
  @type decimal_dtype :: {:decimal, pos_integer(), pos_integer()}
```

lib/explorer/series.ex
```diff
@@ -170,10 +181,12 @@ defmodule Explorer.Series do
...
 
-  @type dtype_alias :: integer_dtype_alias | float_dtype_alias
+  @type dtype_alias :: integer_dtype_alias | float_dtype_alias | decimal_dtype_alias
   @type float_dtype_alias :: :float | :f32 | :f64
   @type integer_dtype_alias :: :integer | :u8 | :u16 | :u32 | :u64 | :s8 | :s16 | :s32 | :s64
+  @type decimal_dtype_alias :: :decimal | :d0 | :d1 | :d2 | :d3 | :d4 | :d5
```
josevalim, line 189:
```suggestion
  @type decimal_dtype_alias :: :decimal
```

lib/explorer/shared.ex
```diff
@@ -497,7 +518,7 @@ defmodule Explorer.Shared do
   def dtype_to_string({:f, size}), do: "f" <> Integer.to_string(size)
   def dtype_to_string({:s, size}), do: "s" <> Integer.to_string(size)
   def dtype_to_string({:u, size}), do: "u" <> Integer.to_string(size)
-  def dtype_to_string({:decimal, precision, scale}), do: "decimal[#{precision}, #{scale}]"
+  def dtype_to_string({:decimal, precision, scale}), do: "decimal[#{precision || "*"}, #{scale}]"
```
josevalim, line 521:
```suggestion
  def dtype_to_string({:decimal, precision, scale}), do: "decimal[#{precision}, #{scale}]"
```

josevalim (approved):
Just some minor nits and ship it!!!

lib/explorer/shared.ex
```diff
@@ -366,6 +387,14 @@ defmodule Explorer.Shared do
   def merge_numeric_dtype({:f, left}, {:f, right}), do: {:f, max(left, right)}
   def merge_numeric_dtype({:f, _} = float, :null), do: float
   def merge_numeric_dtype(:null, {:f, _} = float), do: float
+
+  def merge_numeric_dtype({:decimal, _, _} = decimal, :null), do: decimal
+  def merge_numeric_dtype(:null, {:decimal, _, _} = decimal), do: decimal
+
+  # For now, float has priority over decimals due to Polars.
+  def merge_numeric_dtype({:decimal, _, _}, {:f, _} = float), do: float
+  def merge_numeric_dtype({:f, _} = float, {:decimal, _, _}), do: float
```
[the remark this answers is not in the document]
philss: I'm handle that below now :)

test/explorer/data_frame_test.exs
```diff
@@ -1687,6 +1687,32 @@ defmodule Explorer.DataFrameTest do
...
+    test "adds decimal series" do
+      df = DF.new(a: [1, 2, 4])
+
+      df1 =
+        DF.mutate(df,
+          b: from_list([Decimal.new("3.14159265359"), Decimal.new(2), nil]),
+          # Casting works, but the df will be different if we don't pass the precision,
+          # because it will take the default precision for decimals, which is "38".
+          # The second problem of not passing precision is that scale will be handled differently.
+          # This example would divide the integers by 10^2 without the precision.
```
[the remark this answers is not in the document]
philss: I think I'm covering most of what I said in the comment, specially after removing the support for `nil` precision and scale. There is still an issue when we do an arithmetic operation and the backend returns a dtype without a precision, but we are casting in the end.

josevalim (approved):

Issue #978, opened 2024-09-03
Use case:  Conditionally mutate data


ryancurtin:
My use case involves mutating data in a particular column, but only if a certain condition matches.  Mutate supports this functionality now, but since my rules can come from an external data source, I'd need to do this programmatically.   Polars has the [when](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.when.html#polars-when) function for this operation.  

I'd like to avoid creating an intermediate column to evaluate my conditions first or using `Series.transform` so I was wondering if there's a recommended approach.  I'm familiar with the library and I have combed through the docs quite a bit, but it wouldn't be the first time I've overlooked something obvious.

billylanchantin:
Does [`cond`](https://hexdocs.pm/explorer/Explorer.Query.html#module-conditionals) work?

If not, can you provide an example?

ryancurtin:
Thanks for your response.  `cond` may not really be what I want, as it has lots of the same drawbacks as `mutate`.  I've created a simple example to further elaborate on my use case:

```elixir
# Stored in DB, can have complicated logic (i.e. multiple conditons, and/or evaluation, etc).
# For simplicity's sake, here are two very basic rules
rules = [
  %Rule{column: "b", conditions: [%{column: "a", operator: ">", value: 100}], match_value: "test", result_value: "good"},
  %Rule{column: "b", conditions: [%{column: "a", operator: ">", value: 500}], match_value: "good", result_value: "great"},
]

df = DF.new(%{
  a: [10, 100, 1000],
  b: ["test", "test", "test"]
})

Enum.reduce(rules, df, fn %Rule{column: column, conditions: conditions, match_value: match, result_value: result}, acc_df ->
  DF.mutate_with(acc_df, fn ldf ->
    # Pseudo-code for what I'd want to happen below - I currently have code to parse and evaluate 
    # the rules iinto a boolean Series, but with no way of using it in `mutate_with`.
    conditions_passed = Evaluator.evaluate_rules(conditions, ldf)

    if conditions_passed do
      %{column => Series.replace(ldf[column], match, result)}
    else
      %{}
    end
  end)
end)
```

The conditions for rules don't have any bespoke operations:. It'd be just equality checks and comparison operators such as >, <, >=, <=, etc.  With that being the case, I was optimistic there is a solution buried somewhere in the existing Explorer API.  Please let me know if you'd like any further clarification.

josevalim:
[...]

PR #989, opened 2024-09-21
De-functionalize query internals

billylanchantin:
### Description

Change how `Explorer.Query` functions internally. This is step 1 of implementing the ideas from this PR:

* https://github.com/elixir-explorer/explorer/pull/944

From @josevalim:

> We can make query not return a function and that should be a relatively small change. We could even make it so accessing a lazy dataframe returns lazy series, so we can build queries without writing S.col. The semantics are well defined internally, we just chose to not expose them.

Before it worked like this:

```elixir
# User writes
DF.filter(df, expr) 

# Macro rewrites as
DF.filter_with(df, callback)

# callback has type
Explorer.Backend.LazyFrame.t() -> Series.lazy_t()

# Inside `filter_with/2`
ldf = Explorer.Backend.LazyFrame.new(df)
lazy_series = callback.(ldf)
```

Now we do:

```elixir
# User writes
DF.filter(df, expr)

# Macro rewrites as
DF.filter_with(df, lazy_series)
```

This is equivalent, but there's no longer the callback step. Before if you wanted to use `filter_with/2`, you had to write a callback. For backwards compatibility you can still do that. But now you can also do:

```elixir
df = DF.new(a: [1, 2, 3])
ldf = Explorer.Backend.LazyFrame.new(df)
lazy_series = ldf["a"] |> S.less(3)
# #Explorer.Series<
#   LazySeries[???]
#   boolean (column("a") < 3)
# >

DF.filter_with(df, lazy_series)
#=> %{a: [1, 2]}
```

This PR is all internal. Next I'd like to target adding `Access` to lazy frames.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2552,7 +2552,11 @@ defmodule Explorer.DataFrame do
   defmacro filter(df, query) do
     quote do
       require Explorer.Query
-      Explorer.DataFrame.filter_with(unquote(df), Explorer.Query.query(unquote(query)))
+
+      Explorer.DataFrame.filter_with(
+        unquote(df),
+        Explorer.Query.query(Explorer.Backend.LazyFrame.new(unquote(df)), unquote(query))
+      )
     end
   end
 
@@ -2603,15 +2607,20 @@ defmodule Explorer.DataFrame do
         species string ["Iris-setosa", "Iris-setosa", "Iris-setosa", "Iris-setosa", "Iris-setosa", ...]
       >
   """
+  @type lazy_series_or_list :: Series.lazy_t() | [Series.lazy_t()]
+  @type deprecated_filter_callback :: (Explorer.Backend.LazyFrame.t() -> lazy_series_or_list())
   @doc type: :single
   @spec filter_with(
           df :: DataFrame.t(),
-          callback :: (Explorer.Backend.LazyFrame.t() -> Series.lazy_t() | [Series.lazy_t()])
+          callback_or_lazy_series_or_list :: deprecated_filter_callback() | lazy_series_or_list()
         ) :: DataFrame.t()
   def filter_with(df, fun) when is_function(fun, 1) do
     ldf = Explorer.Backend.LazyFrame.new(df)
+    filter_with(df, fun.(ldf))
+  end
 
-    case fun.(ldf) do
+  def filter_with(df, lazy_series_or_list) do
+    case lazy_series_or_list do
       %Series{dtype: :boolean, data: %LazySeries{} = data} ->
         Shared.apply_dataframe(df, :filter_with, [df, data])
 
@@ -2796,7 +2805,7 @@ defmodule Explorer.DataFrame do
 
       Explorer.DataFrame.mutate_with(
         unquote(df),
-        Explorer.Query.query(unquote(mutations)),
+        Explorer.Query.query(Explorer.Backend.LazyFrame.new(unquote(df)), unquote(mutations)),
         unquote(opts)
       )
     end
@@ -2928,18 +2937,21 @@ defmodule Explorer.DataFrame do
           callback :: (Explorer.Backend.LazyFrame.t() -> column_pairs(Series.lazy_t())),
           opts :: keyword()
         ) :: DataFrame.t()
-  def mutate_with(%DataFrame{} = df, fun, opts \\ []) when is_function(fun) and is_list(opts) do
+  def mutate_with(df, query_or_fun, opts \\ [])
+
+  def mutate_with(%DataFrame{} = df, fun, opts) when is_function(fun) and is_list(opts) do
+    ldf = Explorer.Backend.LazyFrame.new(df)
+    mutate_with(df, fun.(ldf), opts)
+  end
+
+  def mutate_with(%DataFrame{} = df, query, opts) when is_list(opts) do
     keep = Keyword.get(opts, :keep, :all)
 
     unless keep in [:all, :none] do
       raise ArgumentError, "Invalid value for :keep option. Allowed values are :all or :none."
     end
 
-    ldf = Explorer.Backend.LazyFrame.new(df)
-
-    result = fun.(ldf)
-
-    column_pairs = to_column_pairs(df, result, &query_to_series!/1)
+    column_pairs = to_column_pairs(df, query, &query_to_series!/1)
 
     new_dtypes =
       for {column_name, series} <- column_pairs, into: %{} do
@@ -3326,7 +3338,7 @@ defmodule Explorer.DataFrame do
 
       Explorer.DataFrame.sort_with(
         unquote(df),
-        Explorer.Query.query(unquote(query)),
+        Explorer.Query.query(Explorer.Backend.LazyFrame.new(unquote(df)), unquote(query)),
         unquote(opts)
       )
     end
@@ -3428,15 +3440,19 @@ defmodule Explorer.DataFrame do
              Series.lazy_t() | [Series.lazy_t()] | [{:asc | :desc, Series.lazy_t()}]),
           opts :: [nils: :first | :last, stable: boolean()]
         ) :: DataFrame.t()
-  def sort_with(%DataFrame{} = df, fun, opts \\ []) when is_function(fun, 1) do
-    [_descending? | opts] = Shared.validate_sort_options!(opts)
+  def sort_with(df, fun, opts \\ [])
 
+  def sort_with(%DataFrame{} = df, fun, opts) when is_function(fun, 1) do
     ldf = Explorer.Backend.LazyFrame.new(df)
 
-    result = fun.(ldf)
+    sort_with(df, fun.(ldf), opts)
+  end
+
+  def sort_with(%DataFrame{} = df, column_pairs, opts) do
+    [_descending? | opts] = Shared.validate_sort_options!(opts)
 
     dir_and_lazy_series_pairs =
-      result
+      column_pairs
       |> List.wrap()
       |> Enum.map(fn
         {dir, %Series{data: %LazySeries{} = lazy_series}} when dir in [:asc, :desc] ->
@@ -5635,7 +5651,11 @@ defmodule Explorer.DataFrame do
   defmacro summarise(df, query) do
     quote do
       require Explorer.Query
-      Explorer.DataFrame.summarise_with(unquote(df), Explorer.Query.query(unquote(query)))
+
+      Explorer.DataFrame.summarise_with(
+        unquote(df),
+        Explorer.Query.query(Explorer.Backend.LazyFrame.new(unquote(df)), unquote(query))
+      )
     end
   end
 
@@ -5678,11 +5698,12 @@ defmodule Explorer.DataFrame do
         ) :: DataFrame.t()
   def summarise_with(%DataFrame{} = df, fun) when is_function(fun, 1) do
     ldf = Explorer.Backend.LazyFrame.new(df)
+    summarise_with(df, fun.(ldf))
+  end
 
-    result = fun.(ldf)
-
-    result =
-      Enum.map(result, fn
+  def summarise_with(%DataFrame{} = df, query) do
+    query =
+      Enum.map(query, fn
         {key, nil} ->
           lazy_s = LazySeries.unbacked(:lazy, [nil], :null)
           {key, Explorer.Backend.Series.new(lazy_s, :null)}
@@ -5692,7 +5713,7 @@ defmodule Explorer.DataFrame do
       end)
 
     column_pairs =
-      to_column_pairs(df, result, fn value ->
+      to_column_pairs(df, query, fn value ->
         case value do
           %Series{data: %LazySeries{op: :lazy, args: [nil], dtype: :null}} ->
             value
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -2607,12 +2607,13 @@ defmodule Explorer.DataFrame do
         species string ["Iris-setosa", "Iris-setosa", "Iris-setosa", "Iris-setosa", "Iris-setosa", ...]
       >
   """
-  @type lazy_series_or_list :: Series.lazy_t() | [Series.lazy_t()]
-  @type deprecated_filter_callback :: (Explorer.Backend.LazyFrame.t() -> lazy_series_or_list())
   @doc type: :single
   @spec filter_with(
           df :: DataFrame.t(),
-          callback_or_lazy_series_or_list :: deprecated_filter_callback() | lazy_series_or_list()
+          callback_or_lazy_series_or_list ::
+            (Explorer.Backend.LazyFrame.t() -> Series.lazy_t() | [Series.lazy_t()])
+            | Series.lazy_t()
+            | [Series.lazy_t()]
         ) :: DataFrame.t()
   def filter_with(df, fun) when is_function(fun, 1) do
     ldf = Explorer.Backend.LazyFrame.new(df)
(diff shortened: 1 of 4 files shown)


test/explorer/query_test.exs
```diff
@@ -5,11 +5,11 @@ defmodule Explorer.QueryTest do
...
-  test "allows accessing the dataframe" do
-    assert DF.new(a: [1, 2, 3])
-           |> DF.filter(df()["a"] < ^if(true, do: 3, else: 1))
-           |> DF.to_columns(atom_keys: true) == %{a: [1, 2]}
-  end
+  # test "allows accessing the dataframe" do
+  #   assert DF.new(a: [1, 2, 3])
+  #          |> DF.filter(df()["a"] < ^if(true, do: 3, else: 1))
+  #          |> DF.to_columns(atom_keys: true) == %{a: [1, 2]}
+  # end
```
billylanchantin, line 12: This test is no longer passing. It was added 2 weeks ago as part of this:

* https://github.com/elixir-explorer/explorer/issues/978#issuecomment-2329414215

Since I removed the use `Explorer.Query.df_var/0`, I'm not sure we need this test. The only reason I didn't remove it entirely was because of the docs for `col` and `across`, which I'm not sure what to do about.

lib/explorer/query.ex
```diff
@@ -299,18 +300,20 @@ defmodule Explorer.Query do
...
   """
-  defmacro query(expression) do
-    df = df_var()
-
+  defmacro query(df, expression) do
     quote do
-      fn unquote(df) ->
-        unquote(traverse(expression, df))
+      # To prevent leaking imports, we wrap this operation in a function that we
+      # immediately execute.
```
billylanchantin, line 306: I feel like there's a smarter way to do this. 🤔

lib/explorer/query.ex
```diff
@@ -299,18 +300,20 @@ defmodule Explorer.Query do
...
+      fun = fn ->
+        unquote(traverse_root(expression, df))
       end
+
+      fun.()
```
billylanchantin, line 311: I feel like there's a smarter way to do this. 🤔

lib/explorer/query.ex
```diff
@@ -409,18 +419,14 @@ defmodule Explorer.Query do
...
-        {right, vars} = traverse(right, vars, state)
-        {{left, meta, right}, vars}
+    K.if is_atom(left)
+         |> K.and(is_list(right))
+         |> K.and(special_form_defines_var?(left, right)) do
+      raise ArgumentError, "#{left}/#{length(right)} is not currently supported in Explorer.Query"
+    else
+      {left, vars} = traverse(left, vars, state)
+      {right, vars} = traverse(right, vars, state)
+      {{left, meta, right}, vars}
```
josevalim, line 429: Are these changes related to this PR or is it just changing the style? If so, can we do it in a separate PR? :)

lib/explorer/data_frame.ex
```diff
@@ -2552,7 +2552,11 @@ defmodule Explorer.DataFrame do
   defmacro filter(df, query) do
     quote do
       require Explorer.Query
-      Explorer.DataFrame.filter_with(unquote(df), Explorer.Query.query(unquote(query)))
+
+      Explorer.DataFrame.filter_with(
+        unquote(df),
+        Explorer.Query.query(Explorer.Backend.LazyFrame.new(unquote(df)), unquote(query))
```
josevalim, line 2558: We should not unquote a variable twice, if we do that, it means that if I write this code:

```
DF.new()
|> ...
|> ...
|> DF.filter(...)
```

everything on the left side will be executed twice.

I am thinking we should probably not change the `Explorer.Query` API, should we? What if we change `filter_with` instead to accept either an anonymous function or a lazy frame? And we keep Explorer.Query as it is?

josevalim:
There is another concern here related to exposing `Explorer.Backends.LazyFrame` or whatever we are going to call it. We already `to_lazy` and that returns something different. Maybe we should instead have `Explorer.Query.new(df)` and we rename `Explorer.Backends.LazyFrame` to `QueryFrame`?

Then we change `filter_with` and friends to accept either an anonymous function or the result of a `Explorer.Query.new` (which will be a Explorer.Backends.QueryFrame`). This way we keep everything related to the `Explorer.Query` API?

If so, I'd do the following changes:

* PR 1: Rename Explorer.Backends.LazyFrame to QueryFrame
* PR 2: Introduce Explorer.Query.new and allow `filter_with` to accept it as an argument beyond functions
* PR 3: Add access behaviour
* PR 4: Any pending style refactoring

lib/explorer/query.ex
billylanchantin: Haha yes absolutely, sorry for the clutter!

billylanchantin: [6dd174b](https://github.com/elixir-explorer/explorer/pull/989/commits/6dd174bb4ec53bf4180c9b71aa97e7e480befb5b)

billylanchantin:
@josevalim I think you're on the right track. Let me clarify my goal.

I want to make functionality like Polars [expressions](https://docs.pola.rs/user-guide/concepts/expressions/) more 1st class. I want users to be able to directly create/manipulate an expression-like data structure much like you can do with an `Ecto.Queryable`. And I want this data structure to replace the callbacks that are the current go-between for our `verb`/`verb_with` pairs.

We kind of have expressions already in the form of `Backend.LazySeries`. But our API is designed such that they're only an implementation detail. I want to either expose them or replace them.

I like the idea of a `Explorer.Query.new` or similar that returns this new data structure. But I'm not sure renaming `Backend.LazyFrame` as `Backend.QueryFrame` is quite right. I think they serve different purposes.

Here's my understanding of the concepts at play:

Polars | Explorer
:-- | :--
DataFrame | DataFrame
LazyFrame | LazyFrame
Series | Series
Expression | LazySeries

All that said, I'm pretty open to other suggestions on how to make this work! As I found out in the other PR, it's a bit of a tricky needle to thread.

lib/explorer/data_frame.ex
billylanchantin: Oh that's not good... 😬 Great call out.

I think this PR probably won't go forward as is because of https://github.com/elixir-explorer/explorer/pull/989#issuecomment-2366836563, but if it does I'll address this.

josevalim:
I think the table is missing one entry, which is that we need a QueryFrame that, when accessed, returns LazySeries. That’s different from a lazy frame (the result of DF.to_lazy)

billylanchantin:
Ok thanks this is good food for thought.

I think I need to play with a few options. I want to make sure `QueryFrame`s feel ergonomic. And I'm not quite sure how the interact with the other 4 concepts yet.

Once I have something I'll write it up. I'll also maybe move this to an issue. I'm realizing I'm still too much in the designing phase.

josevalim:
I think this PR is almost there. To get started, you could:

1. add Explorer.Query.new as an alias to LazyFrame.new
2. Add new clauses to the _with functions
3. Add the access behavior

And if you change nothing else, it should be what you want. Everything else I mentioned is cleanup/refactoring. It is just that I tend to think bottom-up. :)

billylanchantin:
Did the naive thing but then I realized that this is now possible:

```elixir
alias Explorer.{DataFrame, Query, Series}

df1 = DataFrame.new(a: [1, 2, 3])
df2 = DataFrame.new(b: [4, 5, 6])

qf1 = Query.new(df1)
qf2 = Query.new(df2)

c_lazy = Series.add(qf1["a"], qf2["b"])

DataFrame.mutate_with(df1, c: c_lazy)
```

```
** (RuntimeError) Polars Error: not found: b: 'with_columns' failed
    (explorer 0.10.0-dev) lib/explorer/polars_backend/shared.ex:53: Explorer.PolarsBackend.Shared.apply_dataframe/4
    (explorer 0.10.0-dev) lib/explorer/polars_backend/data_frame.ex:659: Explorer.PolarsBackend.DataFrame.mutate_with/3
    (explorer 0.10.0-dev) lib/explorer/data_frame.ex:2983: Explorer.DataFrame.mutate_with/3
    iex:13: (file)
```

test/explorer/query_test.exs
billylanchantin: [7ac1ada](https://github.com/elixir-explorer/explorer/pull/989/commits/7ac1adaef2169e7d6478f721efa62ebd0f1a0d8c)

lib/explorer/query.ex
billylanchantin: [8edd4de](https://github.com/elixir-explorer/explorer/pull/989/commits/8edd4de8c4491a221fcee426705d256786956b7b)

lib/explorer/data_frame.ex
billylanchantin: [fcdad36](https://github.com/elixir-explorer/explorer/pull/989/commits/fcdad369c481e77d636bae4eb8fde86a99c8e1f7)

lib/explorer/data_frame.ex
```diff
@@ -270,7 +270,9 @@ defmodule Explorer.DataFrame do
 
   defguardp is_column(column) when is_binary(column) or is_atom(column) or is_integer(column)
   defguardp is_column_name(column) when is_binary(column) or is_atom(column)
-  defguardp is_column_pairs(columns) when is_list(columns) or is_map(columns)
+
+  defguardp is_column_pairs(columns)
```
josevalim, line 274:
```suggestion
  # TODO: Use is_non_struct_map when we require Elixir v1.18+
  defguardp is_column_pairs(columns)
```

josevalim (approved):
Beautiful! We only need docs and this is good to go IMO!

philss (approved):
This is a nice improvement! 💯  :shipit:

billylanchantin:
Sorry for the delay!

Notes:

* I renamed `Explorer.Backend.LazyFrame` as `Explorer.Backend.QueryFrame` as discussed.
  * My first draft of the docs were super confusing otherwise.
* The "Implementation details" section is now kinda long. I'm happy to trim it if preferred :)

lib/explorer/query.ex
```diff
@@ -257,17 +257,120 @@ defmodule Explorer.Query do
...
+
+      # Callback equivalent
+      Explorer.DataFrame.mutate_with(df, fn query_frame ->
+        lazy_column_a = query_frame["a"]
+        lazy_column_b = Explorer.Series.add(lazy_column_a, 1)
+        [b: lazy_column_b]
+      end)
+
+  Although each verb expects a slightly different return value from their
+  callback, all expect a `%Explorer.Backend.QueryFrame{}` as the input. The
```
josevalim, line 286: They don't expect a Explorer.Backend.QueryFrame, it is a regular DataFrame, but backed by that given backend.

lib/explorer/query.ex
```diff
@@ -257,17 +257,120 @@ defmodule Explorer.Query do
...
+  `%Explorer.Backend.QueryFrame{}` is a special struct which acts like an
+  `%Explorer.DataFrame{}` except that the `Access` behaviour is defined to
+  return _lazy_ versions of its columns.
+
+  To see this in action, we can work directly with an
+  `%Explorer.Backend.QueryFrame{}` since the `*_with` verbs are also designed to
```
josevalim, line 292: Same here, it is still a regular DataFrame.

lib/explorer/query.ex
```diff
@@ -257,17 +257,120 @@ defmodule Explorer.Query do
...
+      If you are attempting to interpolate a value, use ^b.
+
+  However, this meant that a new `%Explorer.Backend.QueryFrame{}` struct must be
+  built for every operation to ensure the columns and dtypes are valid. Hence
+  the callback interface: rather than forcing you to create a
+  `%Explorer.Backend.QueryFrame{}` every time, you write a callback instead that
+  accepts the fresh `%Explorer.Backend.QueryFrame{}` as input.
+
+  There are some limited instances where it's more convenient to work with
+  `Explorer.Backend.QueryFrame{}`s. If you want to re-use a lazy series, you can
```
josevalim, line 345: This section is about implementation details but you are giving advice to users on how to use part of the library. I recommend extracting this out and adding documentation to the `new` function, explaining when you should prefer `Explorer.Query.new` over the callback/macro API.

lib/explorer/query.ex
```diff
@@ -292,6 +395,17 @@ defmodule Explorer.Query do
 
   @kernel_only kernel_only -- kernel_only -- kernel_all
 
+  @doc """
+  Builds an `%Explorer.Backend.QueryFrame{}` that returns lazy versions of the
```
josevalim, line 399: It returns a regular dataframe and not that struct. We should probably not mention this struct anywhere, except in the implementation detail notes. :)

josevalim:
Great @billylanchantin! I have added some feedback to the docs. In a nutshell, the implementation detail docs are going for too long and we have no documentation on how and why to use the `Explorer.Query.new` feature. I would focus on the user focusing API and minimize the changes to the implementation details. :)

lib/explorer/query.ex
billylanchantin: Oops you're right! I forgot it was a query frame "backed" dataframe:

```elixir
iex(2)> df |> Explorer.DataFrame.mutate_with(fn x ->
...(2)> x |> IO.inspect()
...(2)> [b: x["a"] |> Explorer.Series.add(1)]
...(2)> end)
#Explorer.DataFrame<
  QueryFrame[??? x 1]
  a s64
>
#Explorer.DataFrame<
  Polars[3 x 2]
  a s64 [1, 2, 3]
  b s64 [2, 3, 4]
>
```

Will fix.

billylanchantin:
Ok I think that's better. Thanks for the review :)

Do we still like the `LazyFrame` -> `QueryFrame` change? Or shall I revert that?

lib/explorer/backend/query_frame.ex
```diff
@@ -1,8 +1,8 @@
-defmodule Explorer.Backend.LazyFrame do
+defmodule Explorer.Backend.QueryFrame do
   @moduledoc """
   Represents a lazy dataframe for building query expressions.
 
-  The LazyFrame is available inside `filter_with`, `mutate_with`, and
+  The QueryFrame is available inside `filter_with`, `mutate_with`, and
```
josevalim, line 5:
```suggestion
  You may call `Explorer.Query.new` to create a query-backed dataframe,
  The QueryFrame is available inside `filter_with`, `mutate_with`, and
```

josevalim (approved):
This is great. The name QueryFrame is also much better to me. My only remaining question is if we rename to rename LazySeries to QuerySeries but I think they are not equivalent to QueryFrame (as in you can actually use a LazySeries in most series operations, but you can't do that with a QueryFrame).

billylanchantin:
> My only remaining question is if we rename to rename LazySeries to QuerySeries but I think they are not equivalent to QueryFrame (as in you can actually use a LazySeries in most series operations, but you can't do that with a QueryFrame).

Yes that's my take too.

"Query" is good in `QueryFrame` because all you can do is access, aka query, it. `LazySeries` represent computations, so "query" feels wrong to me.

lib/explorer/backend/query_frame.ex
```diff
@@ -1,8 +1,9 @@
-defmodule Explorer.Backend.LazyFrame do
+defmodule Explorer.Backend.QueryFrame do
   @moduledoc """
   Represents a lazy dataframe for building query expressions.
 
-  The LazyFrame is available inside `filter_with`, `mutate_with`, and
+  You may call `Explorer.Query.new` to create a query-backed dataframe,
```
billylanchantin, line 5:
```suggestion
  You may call `Explorer.Query.new` to create a query-backed dataframe.
```

Issue #990, opened 2024-09-23
`from_parquet` does not handle `FSS.HTTP.Entry` when `lazy: true`


aymanosman:
```
frame = Explorer.DataFrame.from_parquet!("https://huggingface.co/datasets/aqubed/kub_tickets_small/resolve/main/data/train-00000-of-00001-47868532d4f55873.parquet", lazy: true)
** (FunctionClauseError) no function clause matching in Explorer.PolarsBackend.LazyFrame.from_parquet/4

    The following arguments were given to Explorer.PolarsBackend.LazyFrame.from_parquet/4:

        # 1
        %FSS.HTTP.Entry{
          url: "https://huggingface.co/datasets/aqubed/kub_tickets_small/resolve/main/data/train-00000-of-00001-47868532d4f55873.parquet",
          config: %FSS.HTTP.Config{headers: []}
        }

        # 2
        nil

        # 3
        nil

        # 4
        false

    Attempted function clauses (showing 2 out of 2):

        def from_parquet(%FSS.S3.Entry{} = entry, max_rows, columns, _rechunk)
        def from_parquet(%FSS.Local.Entry{} = entry, max_rows, columns, _rechunk)

    (explorer 0.10.0-dev) lib/explorer/polars_backend/lazy_frame.ex:205: Explorer.PolarsBackend.LazyFrame.from_parquet/4
    (explorer 0.10.0-dev) lib/explorer/data_frame.ex:871: Explorer.DataFrame.from_parquet!/2
    iex:4: (file)
```

billylanchantin:
Thanks for this report too, @aymanosman!

I don't have the time to dig into it at the moment. Hopefully it's as simple a change as #992.

If you're inclined to dig into it, PRs are welcome :)

ceyhunkerti:
https://github.com/elixir-explorer/explorer/pull/993
here this one works with the url above needs a review and further guidance if possible

PR #993, opened 2024-09-24
LazyFrame from http endpoint

ceyhunkerti:
- Fixes the related issue => https://github.com/elixir-explorer/explorer/issues/990

- ! Couldn't add test maybe someone more knowledgeable on the repo can point me how to add it the `Bypass` setup doesn't seem to work with lazy option.
- Also I observed `DF.from_ndjson` doesn't work with `lazy` option although it's documented. I'll check further and create an issue if I can confirm it.

Tried to add a simple test like below but it's hanging

```elixir
  describe "from_parquet/2 - HTTP" do
    setup do
      [bypass: Bypass.open(), df: Explorer.Datasets.wine()]
    end

    test "reads a parquet file from an HTTP server", %{bypass: bypass, df: df} do
      Bypass.expect(bypass, "GET", "/path/to/file.parquet", fn conn ->
        bytes = Explorer.DataFrame.dump_parquet!(df)
        Plug.Conn.resp(conn, 200, bytes)
      end)

      url = http_endpoint(bypass.port) <> "/path/to/file.parquet"

      assert {:ok, ldf} = Explorer.DataFrame.from_parquet(url, lazy: true)
      df1 = DF.compute(ldf)

      assert DF.to_columns(df1) == DF.to_columns(df)
    end
  end

  defp http_endpoint(port), do: "http://localhost:#{port}"

  ```

Eventually failing with something like

```
     test/explorer/data_frame/lazy_test.exs:223
     match (=) failed
     code:  assert {:ok, ldf} = DF.from_parquet(url, lazy: true)
     left:  {:ok, ldf}
     right: {
              :error,
              %RuntimeError{message: "Polars Error: Generic HTTP error: Request error: Error after 2 retries in 180.2322859s, max_retries:10, retry_timeout:180s, source:error sending request for url (http://localhost:46017/test): 'parquet scan' failed: 'select' input failed to resolve"}
            }
     stacktrace:
       test/explorer/data_frame/lazy_test.exs:232: (test)

```

### DIFF
--- a/lib/explorer/polars_backend/lazy_frame.ex
+++ b/lib/explorer/polars_backend/lazy_frame.ex
@@ -10,6 +10,7 @@ defmodule Explorer.PolarsBackend.LazyFrame do
 
   alias FSS.Local
   alias FSS.S3
+  alias FSS.HTTP
 
   import Explorer.PolarsBackend.Expression, only: [to_expr: 1, alias_expr: 2]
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by philss** [comment]
> ! Couldn't add test maybe someone more knowledgeable on the repo can point me how to add it the Bypass setup doesn't seem to work with lazy option.

I think the problem may be to a restriction in Polars: it may require HTTPs. If you tested manually and it's working, we can move on without the test :)

**Comment by ceyhunkerti** [comment]
yes, I tested it with the following from the [original issue](https://github.com/elixir-explorer/explorer/issues/990) and it's working 
```elixir
frame = Explorer.DataFrame.from_parquet!("https://huggingface.co/datasets/aqubed/kub_tickets_small/resolve/main/data/train-00000-of-00001-47868532d4f55873.parquet", lazy: true)
```

philss (approved):

philss:
@ceyhunkerti awesome! Thank you!

josevalim:

Issue #994, opened 2024-09-26
cannot encode value Binary to term (when Binary is nested in struct)


aymanosman:
I'm getting this error with a dataset that contains `Binary` nested in a `struct`

```
thread '<unnamed>' panicked at src/encoding.rs:786:15: cannot encode value Binary([1,2.3]) to term
```

Steps to reproduce.

```python
import polars as pl
df = pl.DataFrame({"image": [b'\x01\x02\x03']})
df.write_parquet("ok.parquet")
df = pl.DataFrame({"image": [{"bytes": b'\x01\x02\x03'}]})
df.write_parquet("bad.parquet")
```

```elixir
iex(41)> Explorer.DataFrame.from_parquet!("ok.parquet")
#Explorer.DataFrame<
  Polars[1 x 1]
  image binary [<<1, 2, 3>>]
>

iex(42)> Explorer.DataFrame.from_parquet!("bad.parquet")
thread '<unnamed>' panicked at src/encoding.rs:783:15:
                                                      cannot encode value Binary([1, 2, 3]) to term
            note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  #Inspect.Error<
  got ErlangError with message:

      """
      Erlang error: :nif_panicked
      """

  while inspecting:

      %{
        data: %Explorer.PolarsBackend.DataFrame{
          resource: #Reference<0.315878801.3185704982.174156>
        },
        remote: nil,
        names: ["image"],
        __struct__: Explorer.DataFrame,
        dtypes: %{"image" => {:struct, [{"bytes", :binary}]}},
        groups: []
      }

  Stacktrace:

    (explorer 0.10.0-dev) Explorer.PolarsBackend.Native.s_to_list(#Explorer.PolarsBackend.Series<
  shape: (1,)
  Series: 'image' [struct[1]]
  [
  	{b"\x01\x02\x03"}
  ]
>)
    (explorer 0.10.0-dev) lib/explorer/polars_backend/shared.ex:24: Explorer.PolarsBackend.Shared.apply_series/3
    (explorer 0.10.0-dev) lib/explorer/backend/data_frame.ex:324: anonymous fn/3 in Explorer.Backend.DataFrame.build_cols_algebra/3
    (elixir 1.17.2) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
[...]

PR #995, opened 2024-09-26
fix - cannot encode value Binary to term

ceyhunkerti:
fixes https://github.com/elixir-explorer/explorer/issues/994

```py
import polars as pl
df = pl.DataFrame({"image": [b'\x01\x02\x03']})
df.write_parquet("ok.parquet")
df = pl.DataFrame({"image": [{"bytes": b'\x01\x02\x03'}]})
df.write_parquet("bad.parquet")
```
```elixir
iex(1)> Explorer.DataFrame.from_parquet!("bad.parquet")
#Explorer.DataFrame<
  Polars[1 x 1]
  image struct[1] [%{"bytes" => [1, 2, 3]}]
>
```


josevalim:
Looks great! Can you please add a test as well? :) You should be able to write the parquet file using a binary series.

philss:
Thanks for the PR, @ceyhunkerti! I made a fix for that in #996 returning the binary series , so I'm closing this one.

PR #997, opened 2024-09-27
Update dependencies

philss:
Rust and Elixir deps. The only one that is out of this list is Nx, that is breaking the build.


josevalim:
@philss the new Nx changed some defaults, so we may need to explicit pass `type: {..., ...}` in some cases. But other than that, it should work!

philss:
@josevalim I see. I will take a look. Thanks!

Issue #1002, opened 2024-10-13
Incorrect typespec return value for Dataframe.dtypes


mooreryan:
Here is the spec for the `Dataframe.dtypes` function:

https://github.com/elixir-explorer/explorer/blob/ca2ca1f80e4cb994f3182e8d59e9608bad95385e/lib/explorer/data_frame.ex#L2051

Here is the spec for the `dtypes` field in the `DataFrame` struct:

https://github.com/elixir-explorer/explorer/blob/ca2ca1f80e4cb994f3182e8d59e9608bad95385e/lib/explorer/data_frame.ex#L262

Given that (and the function's actual behavior), the spec for  `Dataframe.dtypes` should probably be something like this:

```elixir
@spec dtypes(df :: DataFrame.t()) :: %{String.t() => Explorer.Series.dtype()} 
```

PR #1003, opened 2024-10-14
Fix return type in `DataFrame.dtypes/1` typespec (#1002)

mooreryan:
This change uses the same type for the `dtypes` struct field and the `dtypes` function.

Resolves https://github.com/elixir-explorer/explorer/issues/1002.


josevalim:

Issue #1001, opened 2024-10-13
Potential improvement for DataFrame Access docs


mooreryan:
I was thinking that it could be useful to mention what happens when trying to access a `DataFrame` column that does not exist.

The docs mention that `DataFrame` implements `Access`: 

> [Explorer.DataFrame](https://hexdocs.pm/explorer/Explorer.DataFrame.html) also implements the [Access](https://hexdocs.pm/elixir/Access.html) behaviour (also known as the brackets syntax).

When using the bracket access syntax, an error is raised rather than returning `nil`, as the [`Access` docs](https://hexdocs.pm/elixir/Access.html) mention:

> Both return nil if the key does not exist

Here is what happens:

```elixir
df =
  DataFrame.new(
    a: [1, 2, 3],
    b: [10, 20, 30]
  )

df["c"]

** (ArgumentError) could not find column name "c". The available columns are: ["a", "b"].
If you are attempting to interpolate a value, use ^c.
```

I was thinking an example like this in the `DataFrame` docs could be helpful.  If this would be welcome, I could open a PR request for review.

(For more context, this is related to a [question](https://elixirforum.com/t/access-behaviour-for-explorer-dataframe/66691) I asked on the Elixir Forum.)

josevalim:
Please send a PR adding clarifications to our docs on Access, we can work on it together :)

Issue #723, opened 2023-10-23
Add a cheatsheet


cigrainger:
I'm excited about cheatsheets and something like this would beat "Ten minutes to Explorer", especially for those coming from dplyr or pandas who just need an easy reference. 

dplyr: https://nyu-cdsc.github.io/learningr/assets/data-transformation.pdf
also dplyr: https://www.rstudio.com/wp-content/uploads/2015/02/data-wrangling-cheatsheet.pdf
pandas: https://pandas.pydata.org/Pandas_Cheat_Sheet.pdf

I started playing over the weekend but I think the diagrams are really powerful and got myself stuck trying to figure out how to replicate them in vega lite.

billylanchantin:
I'd love to see cheatsheets too! Personally, I think it would've saved me some documentation hunting.

For example, early on I saw there was a `DataFrame.filter/2` but I couldn't find a corresponding `Series.filter/2`. I eventually found `Series.mask`, but it wasn't obvious. Something like a task-oriented cheatsheet would've made that much easier to find.

> I think the diagrams are really powerful and got myself stuck trying to figure out how to replicate them in vega lite.

I've also feel like I've lost a good bit of time trying to get vegalite to output specific visualizations. It's great for a lot of cases. But when you start doing non-data-driven things like annotating your visualization, I found that it gets tricky.

Were you thinking of doing a `.cheatmd`? Or a `.pdf`?

cigrainger:
Glad to hear it would be helpful! I was thinking of doing a `.cheatmd` so we can easily put it into the docs. I feel like it should be possible to get close with heatmaps based on specific categorical values. And I also agree that it's confusing we have `Series.mask/2` instead of `Series.filter/2`. 

@philss or @josevalim I'm sure there was a convo about this but I can't remember what the reasoning was for this anymore. https://github.com/elixir-explorer/explorer/pull/326#issue-1352162455

josevalim:
We changed the implementation and renamed at the same time but I am fine with reverting the name back to `filter`. :) It should be a quick change and we can add:

```
@deprecated "Use Explorer.Series.filter/2 instead"
def mask(s1, s2), do: filter(s1, s2) 
```

It will certainly be much easier to find.

billylanchantin:
> I feel like it should be possible to get close with heatmaps based on specific categorical values.

I agree! I was more worried about the arrows:

[...]

PR #1004, opened 2024-10-14
Add access example to DataFrame docs (#1001)

mooreryan:
To start, I added one example that shows what happens when trying to access a data frame column that does not exist.

I'm not sure how much to go into some of the comments in #1001, e.g., the fact that an error is raised instead of returning `nil` and other aspects of the `Access` behaviour that might be surprising.

What do you think, should it be kept simple like this or should I add something specifically about Access behaviour?

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -188,6 +188,17 @@ defmodule Explorer.DataFrame do
       iex> df["class"][3]
       1
 
+
+  When trying to access a column in the dataframe that does not exist, an `ArgumentError` is raised.
+
+      iex> df = Explorer.Datasets.wine()
+      iex> try do
+      ...>   {:ok, df["abcd"]}
+      ...> rescue
+      ...>   _ in ArgumentError -> :error
+      ...> end
+      :error
+
   """
 
   alias __MODULE__, as: DataFrame


lib/explorer/data_frame.ex
```diff
@@ -188,6 +188,17 @@ defmodule Explorer.DataFrame do
...
       1
 
+
+  When trying to access a column in the dataframe that does not exist, an `ArgumentError` is raised.
+
+      iex> df = Explorer.Datasets.wine()
+      iex> try do
+      ...>   {:ok, df["abcd"]}
+      ...> rescue
+      ...>   _ in ArgumentError -> :error
```
billylanchantin, line 198:
```suggestion
      ...>   ArgumentError -> :error
```

billylanchantin (changes requested):
I requested a slight simplification we can do.

What other aspects of the `Access` implementation did you maybe want to highlight?

lib/explorer/data_frame.ex
```diff
@@ -188,6 +188,17 @@ defmodule Explorer.DataFrame do
...
+      ...> end
+      :error
```
josevalim, line 200:
```suggestion
      iex> df["abcd"]
      ** (ArgumentError) could not find column name "omg"
```

lib/explorer/data_frame.ex
josevalim: Good point, I think we can even be better than the try/catch. I dropped another suggestion :)

billylanchantin: That's way better :P

mooreryan:
> What other aspects of the Access implementation did you maybe want to highlight?

Mainly some of the aspects from this discussion: https://elixirforum.com/t/access-behaviour-for-explorer-dataframe/66691. (Though, looking at the username, I think you might already be in that discussion!)

To summarize that thread, I was surprised by the fact that DataFrame had a different `Access` behavior than what the docs for the `Access` behavior would imply. E.g., raising rather than returning `nil`, `fetch` raising rather than returning `:error`, `Access.fetch!` raising `ArgumentError` rather than `KeyError`, etc.

I wasn't sure if these doc change was the spot to get into all of that though.

lib/explorer/data_frame.ex
billylanchantin: Ah I just assumed the above would work 😬 

Yeah I'd honestly go even simpler:

```elixir
iex> df =  DataFrame.new(a: [1, 2, 3])
iex> df["b"]
** (ArgumentError) could not find column name "b". The available columns are: ["a"].
If you are attempting to interpolate a value, use ^b.
```

^ Not tested...

billylanchantin:
Yep that's me!

I wanted to clarify because some points (like not needing to implement `fetch!`) were cleared up in the post but others weren't. I figured it'd be a good idea to make a fresh list all the stuff we might want to document here.

If we do want to include more than what's `["not_a_column"]` raising vs. returning `nil`, I think I'd prefer to include a brief overview of the reasoning behind the design choices. Something like:

```
### Missing columns

Explorer will generally raise an `ArgumentError` if the argument to `[]` is not
found, but not always. For example:

...your example...

This deviates from other `Access` implementations which return `nil` when the
key is not found. We raise early to help users catch errors quicker since most
DataFrame operations could not meaningfully continue if the result was `nil`.

We also raise an `ArgumentError` rather than a `KeyError`. This is because `[]`
is designed to take a wider collection of arguments than a single key, e.g.
lists and regexes.

However, note that some `[]` calls will not raise if their argument is not
found. This occurs when argument to `[]` is more of a search criteria like
a regex:

    iex> df = Explorer.DataFrame.new(col_1: [1], col_2: [2], other: ["a"])
    iex> col_n_regex = ~r/^col_\d$/
    iex> df[col_n_regex]
    #Explorer.DataFrame<
      Polars[1 x 2]
      col_1 s64 [1]
      col_2 s64 [2]
    >
    iex> nothing_will_match_regex = ~r/not_a_pattern/
    iex> df[nothing_will_match_regex]
    #Explorer.DataFrame<
      Polars[0 x 0]
    >
```

I'm not married to any of that language. But explaining the "why" might be nice (if the team agrees!).

lib/explorer/data_frame.ex
mooreryan: I'm running into an issue matching on this error message:

```
     Doctest failed: wrong message for ArgumentError
     expected:
       "could not find column name \"b\". The available columns are: [\"a\"].\nIf you are attempting to interpolate a value, use ^b."
     actual:
       "could not find column name \"b\". The available columns are: [\"a\"].\nIf you are attempting to interpolate a value, use ^b.\n"
```

It has a trailing newline and I cannot get that to work.  I have tried various combinations of spaces and newlines to try and get it to work out...but I have not been able to.  Turns out that the link I posted previosly (https://elixirforum.com/t/handling-multi-newline-exception-messages-in-doctests-elixir-v1-12-0/39912) is actually talking about the same problem the doctest matching the argument error.  And it looks like a problem isn't addressable right now.

So I'm thinking the best thing is to use the `try/rescue` way after all. What do you think?

mooreryan:
> But explaining the "why" might be nice

I like the idea of adding some clarification of the reasoning of why things were designed this way.  For me, it was pretty surprising to see the same action (trying to access a column) have different result depending on the argument, (sometimes raising, sometimes returning an empty dataframe, etc.).

(note: this got a bit rambling, there is a summary at the bottom)

## More examples

To be honest, I don't understand the reasoning behind sometimes raising and sometimes returning an empty dataframe.  E.g., for this dataframe:

```
iex(1)> df = DataFrame.new(a: [1, 2])
#Explorer.DataFrame<
  Polars[2 x 1]
  a s64 [1, 2]
>
```

This raises:

```
iex(2)> df["b"]
** (ArgumentError) could not find column name "b". The available columns are: ["a"].
If you are attempting to interpolate a value, use ^b.

    (explorer 0.9.2) lib/explorer/shared.ex:252: Explorer.Shared.maybe_raise_column_not_found/3
    (explorer 0.9.2) lib/explorer/shared.ex:187: anonymous fn/4 in Explorer.Shared.to_existing_columns/3
    (elixir 1.17.2) lib/enum.ex:1829: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (explorer 0.9.2) lib/explorer/shared.ex:175: Explorer.Shared.to_existing_columns/3
    (explorer 0.9.2) lib/explorer/data_frame.ex:3984: Explorer.DataFrame.pull/2
    (explorer 0.9.2) lib/explorer/data_frame.ex:373: Explorer.DataFrame.fetch/2
    (elixir 1.17.2) lib/access.ex:322: Access.get/3
    iex:2: (file)
```

But this returns an empty data frame:

```
iex(2)> df[~r/b/]
#Explorer.DataFrame<
  Polars[0 x 0]
>
```

Asking for the 2nd column raises:

```
iex(3)> df[1]
** (ArgumentError) no column exists at index 1
    (explorer 0.9.2) lib/explorer/shared.ex:240: Explorer.Shared.fetch_column_at!/2
    (explorer 0.9.2) lib/explorer/shared.ex:178: anonymous fn/4 in Explorer.Shared.to_existing_columns/3
    (elixir 1.17.2) lib/enum.ex:1829: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (explorer 0.9.2) lib/explorer/shared.ex:175: Explorer.Shared.to_existing_columns/3
    (explorer 0.9.2) lib/explorer/data_frame.ex:3984: Explorer.DataFrame.pull/2
    (explorer 0.9.2) lib/explorer/data_frame.ex:373: Explorer.DataFrame.fetch/2
    (elixir 1.17.2) lib/access.ex:322: Access.get/3
    iex:3: (file)
```

But this returns a data frame with just column `a` even though I'm asking for the first both the 1st and 2nd columns:

```
iex(3)> df[0..1]
#Explorer.DataFrame<
  Polars[2 x 1]
  a s64 [1, 2]
>
```

## Comparing to R's tidyverse

I'm not making a value judgement on this, rather trying to explain what I found confusing.  Consider this R [tidyverse](https://www.tidyverse.org/) example.  Here is a data frame.

```
 > df <- tibble(a = 1:2)
> df
# A tibble: 2 × 1
      a
  <int>
1     1
2     2
```

And here attempting to access a column that doesn't exist:

```
 > df["b"]
Error in `df["b"]`:
! Can't subset columns that don't exist.
✖ Column `b` doesn't exist.
```

That gives an error similar to Explorer.  And the result of selecting columns from a data frame with regex also behaves like Explorer, by giving an "empty" dataframe.

```
> df |> select(matches("b"))
# A tibble: 2 × 0
```

Same behavior when accessing a column by index when that index is out of bounds (R using 1-based indexing):

```
> df[2]
Error in `df[2]`:
! Can't subset columns past the end.
ℹ Location 2 doesn't exist.
ℹ There is only 1 column.
```

But doing the equivalent of the `df[0..1]` example from Explorer:

```
> df[1:2]
Error in `df[1:2]`:
! Can't subset columns past the end.
ℹ Location 2 doesn't exist.
ℹ There is only 1 column.
```

That raises an error in R, whereas Explorer gives the data frame with just column `a` returned.

## Summary

Explorer's access behavior is surprising to me in some ways.  Not making a judgement on how it should be, since I don't have enough experience with Explorer, just trying to think of ways that docs could make it clearer.  

I think some of this stuff is out of the scope for this small PR though.  In fact, it would probably make sense as part of a cheat sheet or something like Explorer for Tidyverse users. I may have a go at putting something like that together, but it can be a big task.

(There was a discussion [here](https://github.com/elixir-explorer/explorer/issues/723) about cheat sheets too.)

lib/explorer/data_frame.ex
billylanchantin: Hm that's a shame! I didn't know about this issue with doctests. Yeah I think going back to the `try/rescue` is a fine workaround.

BTW I tried this too but I don't like it:

```
      iex> try do
      ...>   df["b"]
      ...> rescue
      ...>   error in ArgumentError -> error.message
      ...> end
      ~s'could not find column name "b". The available columns are: ["a"].
      If you are attempting to interpolate a value, use ^b.
      '
```

josevalim: We should fix the error message by adding a `\` at the end so it removes the trailing new line.

billylanchantin: @mooreryan If you want to try adding the `\` on this PR I think that's totally fine. But if changing it here means you have to run around editing a bunch of other doctests, we (you or the team) can address it in another PR.

billylanchantin:
To summarize and expand a little, here are your uses in the three relevant libraries:

Use | Explorer | Tidy | Polars
:-- | :-- | :-- | :--
Specific column | `df["b"]` | `df["b"]` | `df["b"]`
String match | `df[~r/b/]` | `select(df, matches("b"))` | `df.select(cs.matches("b"))`
Range past end | `df[0..1]` | `df[1:2]` | `df[0:2]`

And here are their outputs:

Use | Explorer | Tidy | Polars | Agreement?
:-- | :-- | :-- | :-- | :--
Specific column | `ArgumentError` | `Error` | `ColumnNotFoundError` | ✅ Yes
String match | DataFrame | DataFrame | DataFrame | ✅ Yes
Range past end | DataFrame | `Error` | DataFrame | ❌ No

So there's only a mismatch on the range one if I'm following. And I agree it's a bit borderline (and admit I was surprised ranges didn't raise). But when designing the API, the Explorer team is balancing a few tensions: what makes sense in Elixir, what data scientists are used to, what's feasible to implement today via Polars, etc. In this case, though I can't say for sure, I think we just defaulted to how Polars handles ranges rather than the tidyverse. Both seem somewhat reasonable to me.

---

But regardless of the reasoning behind individual design decisions, clearer/better docs are almost always the right call.

And as for the cheatsheet stuff, it would be appreciated! Even a start that can be improved upon later.

josevalim:
I think we should raise for ranges out of bounds, for consistency within DataFrames and also with Nx.

billylanchantin:
Issue for raising with ranges: https://github.com/elixir-explorer/explorer/issues/1005

Thanks for pointing it out, @mooreryan!

mooreryan:
> But when designing the API, the Explorer team is balancing a few tensions: what makes sense in Elixir, what data scientists are used to, what's feasible to implement today via Polars, etc.
> ...
> But regardless of the reasoning behind individual design decisions, clearer/better docs are almost always the right call.

This sort of info would be pretty cool to have in docs somewhere--I'm just not sure where they would ideally go.

lib/explorer/data_frame.ex
billylanchantin: Oh in this case I think you can just remove the final `\n` on 268.

---

Both José and I assumed the newlines came from a docstring like this:

```elixir
"""
could not find column name "#{name}" #{did_you_mean(name, df.names)}
If you are attempting to interpolate a value, use #{name}.
"""
```

When you use docstrings like that, you get a trailing newline. The workaround is to add a `\` as the last character like so:

```elixir
"""
could not find column name "#{name}" #{did_you_mean(name, df.names)}
If you are attempting to interpolate a value, use #{name}.\
"""
```

But here the message was built by gluing strings with newline characters, so our advice wasn't applicable.

billylanchantin:
Yeah I'm not sure. Maybe our contributor's guidelines?

lib/explorer/data_frame.ex
mooreryan: > @mooreryan If you want to try adding the `\` on this PR I think that's totally fine. But if changing it here means you have to run around editing a bunch of other doctests, we (you or the team) can address it in another PR.

I removed the newline in the message.  It only affected two tests, so it seemed small enough to include in this PR.

josevalim:

Issue #1006, opened 2024-10-22
qcut error 'quantiles are not unique while allow_duplicates=False'


petrkozorezov:
It's impossible to calculate quantiles when there are duplicates.

```elixir
Explorer.Series.from_list([0.0, 0.0, 0.0, 3.0, 4.0, 5.0])
|> Explorer.Series.qcut([0.1, 0.25, 0.75])
```

```
** (RuntimeError) Polars Error: duplicate: quantiles are not unique while allow_duplicates=False
    (explorer 0.9.2) lib/explorer/polars_backend/shared.ex:18: Explorer.PolarsBackend.Shared.apply/2
    (explorer 0.9.2) lib/explorer/polars_backend/series.ex:549: Explorer.PolarsBackend.Series.qcut/5
    #cell:tug3ltfjbo7jbnjh:2: (file)
```

philss:
There is an option that we need to expose. I will fix that soon. Thanks for the report!

PR #1007, opened 2024-10-23
Expose options to Series' `cut` and `qcut`

philss:
Closes: https://github.com/elixir-explorer/explorer/issues/1006

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -4868,6 +4868,9 @@ defmodule Explorer.Series do
     * `:category_label` - The name given to the category column.
       Defaults to `category`.
 
+    * `:allow_duplicates` - If quantiles can have duplicated values.
+      Defaults to `false`.
+
   ## Examples
 
       iex> s = Explorer.Series.from_list([1.0, 2.0, 3.0, 4.0, 5.0])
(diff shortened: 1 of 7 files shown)


lib/explorer/series.ex
```diff
@@ -4881,11 +4895,24 @@ defmodule Explorer.Series do
...
   def qcut(series, quantiles, opts \\ []) do
+    opts =
+      Keyword.validate!(opts,
+        labels: nil,
+        break_point_label: nil,
+        category_label: nil,
+        allow_duplicates: false,
+        left_close: false,
+        include_breaks: true
+      )
```
philss, line 4906: I don't know how to explain these two new options - `left_close` and `include_breaks`, so I left out of docs. But if you have something in mind, feel free to suggest :)

josevalim (approved):

lib/explorer/series.ex
billylanchantin: From the Python [docs](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.cut.html#polars.Expr.cut):

* `left_closed`
  * Set the intervals to be left-closed instead of right-closed.
* `include_breaks`
  * Include a column with the right endpoint of the bin each observation falls in. This will change the data type of the output from a Categorical to a Struct.

It's probably fine if we just forward their documentation. Here's a more in depth explanation:

#### `left_closed`

The cuts which are the main argument to this function divide the real line into discrete bins. E.g. if we passed cuts `[-1, 1]`, we'd break up the real line into 3 bins like this (call them `A`, `B`, and `C`):

```
   A       B       C 
-------|-------|-------
      -1       1
```

This is a bit ambiguous though. We need to decide which bins the cuts themselves belong to. Is -1 in `A` or `B`?

By default, the bins are "right closed", meaning the bins contain their right cut but not their left. So -1 $\in$ `A` and 1 $\in$ `B`.

Including the `left_closed` argument flips this. Now the bins contain their left cut but not their right. So -1 $\in$ `B` and 1 $\in$ `C`.

#### `include_breaks`

This argument changes the return of the overall function. Instead of just mapping each value to the bin it's in, it also returns metadata about the bin it was mapped to. So you get a series of structs back instead of a category.

See their docs for an example of the structs returned.

philss: oh, thank you @billylanchantin!! :D
I will add these short descriptions :)

PR #1009, opened 2024-10-23
Fix `cut/3` and `qcut/3` when `:include_breaks` is false

philss:
Also add docs regarding the `:left_close` and `:include_breaks` options.

Thanks @billylanchantin :D
Reference: https://github.com/elixir-explorer/explorer/pull/1007#discussion_r1812946449

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -4824,10 +4824,18 @@ defmodule Explorer.Series do
       bounds (e.g. `(-inf -1.0]`, `(-1.0, 1.0]`, `(1.0, inf]`)
 
     * `:break_point_label` - The name given to the breakpoint column.
+      This is only relevant if `:include_breaks` is `true`.
       Defaults to `break_point`.
 
     * `:category_label` - The name given to the category column.
       Defaults to `category`.
+
+    * `:left_closed` - Set the intervals to be left-closed instead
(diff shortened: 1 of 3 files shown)


lib/explorer/series.ex
```diff
@@ -4824,11 +4824,19 @@ defmodule Explorer.Series do
...
 
     * `:category_label` - The name given to the category column.
       Defaults to `category`.
 
+    * `:left_closed` - Set the intervals to be left-closed instead
+      of right-closed. Defaults to `false`.
+
+    * `:include_breaks` - Include a column with the right endpoint
+      of the bin that each observation falls in.
+      Defaults to `true`.
```
billylanchantin, line 4838: I think we want default `false` here (and same for the other one). This option strikes me more as a way to get metadata about the binning, which probably isn't wanted most of the time.

WDYT?

billylanchantin (changes requested):

lib/explorer/series.ex
philss: @billylanchantin https://github.com/elixir-explorer/explorer/pull/1009/commits/baf75e680b604b94cfd1e3bb8e396009692be5d7

josevalim (approved):

billylanchantin (approved):

Issue #1014, opened 2024-11-12
Fix skips/todos on property tests


billylanchantin:
We discovered a number of bugs while developing #1012. This issue will keep track of known bugs and our progress on them.

* [ ] `DataFrame.new!/2` fails often with the `:decimal` dtype
* [ ] Can't print empty DataFrames
* [ ] Can't print DataFrames in certain scenarios
* [ ] Can't dump to CSV in certain scenarios
* [ ] Can't dump to IPC in certain scenarios
* [ ] Can't dump to NDJSON in certain scenarios
* [ ] Can't dump to PARQUET in certain scenarios

Feel free to edit this description with updates to the bug-list.

billylanchantin:
After #1028, we think the only thing left blocking this issue is #1011.

We're currently excluding all dataframes with the `:category` dtype. Once we address #1011, we can remove that exclusion. Unfortunately there may still be additional `:category`-related issues not related to #1011, but we won't know until it's fixed.

PR #1015, opened 2024-11-14
Fix decimal empty list bug

billylanchantin:
Fixes two things:

1. The `:decimal` generator was generating Elixir-side `#Decimal<>`s that the Rust-side couldn't represent.
2. There was a bug in `s_from_list_decimal` where empty lists failed to have the `Decimal(...)` dtype.

Now we can remove the decimal-related TODOs we added in https://github.com/elixir-explorer/explorer/pull/1012.

See also: https://github.com/elixir-explorer/explorer/issues/1014.

### DIFF
--- a/native/explorer/src/series/from_list.rs
+++ b/native/explorer/src/series/from_list.rs
@@ -242,7 +242,7 @@ pub fn s_from_list_decimal(
                 .map(|ex_decimal| AnyValue::Decimal(ex_decimal.signed_coef(), ex_decimal.scale()))
                 .map_err(|error| {
                     ExplorerError::Other(format!(
-                        "cannot decode a valid decimal from term. error: {error:?}"
+                        "cannot decode a valid decimal from term; check that `coef` fits into an `i128`. error: {error:?}"
                     ))
                 }),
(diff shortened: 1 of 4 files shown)


josevalim (approved):

billylanchantin:
Also worth noting:

We can't currently represent decimals on the Rust-side where the Elixir-side's `decimal.coef` is outside the `u64` range. This isn't really an issue in itself. But it makes the default precision of 38 over powered since the precision is the number of digits in base 10.

```elixir
iex(1)> max_u64 = 2 ** 64 - 1
18446744073709551615

iex(2)> :math.log10(max_u64)
19.265919722494797
```

I think the confusion came because:

```elixir
iex(3)> max_u128 = 2 ** 128 - 1
340282366920938463463374607431768211455

iex(4)> :math.log10(max_u128)
38.53183944498959
```

So if Rust were using an _unsigned_ 128-bit integer, we could push the precision up to 38/39. But because it's using a _signed_ 128-bit integer, we can really only go up to 19/20.

I think we can drop the default down to 20.

Reference: https://arrow.apache.org/docs/cpp/api/datatype.html#classarrow_1_1_decimal128_type

PR #1020, opened 2024-11-20
Fix datetime encoding overflow

billylanchantin:
Fixes one of the #1014 issues with printing dataframes.

This one stems from how we encode datetimes. Example:

```elixir
dtypes = [{"col_j", {:naive_datetime, :millisecond}}]
rows = [[{"col_j", ~N[8332-06-03 16:03:40.489446]}]]
df = DF.new(rows, dtypes: dtypes)
DF.print(df)
```

Panics with:

```
thread '<unnamed>' panicked at src/encoding.rs:167:35:
attempt to multiply with overflow
thread '<unnamed>' panicked at /Users/billy/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-arrow-0.43.1/src/temporal_conversions.rs:159:37:
invalid or out-of-range datetime


  1) test specific property failures 1 (Explorer.DataFrameTest)
     test/explorer/data_frame_test.exs:4850
     code: DF.print(df)
     stacktrace:
       (explorer 0.11.0-dev) Explorer.PolarsBackend.Native.s_to_list(#Explorer.PolarsBackend.Series<
  #Reference<0.2007251331.1252655135.131382>
>)
       (explorer 0.11.0-dev) lib/explorer/polars_backend/shared.ex:24: Explorer.PolarsBackend.Shared.apply_series/3
       (explorer 0.11.0-dev) lib/explorer/data_frame.ex:5981: anonymous fn/2 in Explorer.DataFrame.non_empty_table_string/2
       (elixir 1.16.0) lib/enum.ex:1700: Enum."-map/2-lists^map/1-1-"/2
       (explorer 0.11.0-dev) lib/explorer/data_frame.ex:5979: Explorer.DataFrame.non_empty_table_string/2
       (explorer 0.11.0-dev) lib/explorer/data_frame.ex:5945: Explorer.DataFrame.print/2
       test/explorer/data_frame_test.exs:4854: (test)
```

The culprit is how we encode datetimes: we always turn them into microseconds first. This is usually fine. But with certain operations like printing, especially when the datetime is far from the unix epoch, it can result in overflow.

My proposed fix is to build the `i64` representation directly from the `(%DateTime{}, time_unit)` pair using (mostly) Polars functions. This works -- and I added some properties to be confident -- but it has the drawback of us no longer being able to support `impl From<ExNaiveDateTime> for i64` since we need to know which time unit the user wants to represent the datetime with. We didn't use that `impl` much so I think the drawback is acceptable.

We're also handling the case where nanosecond precision datetimes can't be represented more explicitly.

### DIFF
--- a/native/explorer/src/datatypes.rs
+++ b/native/explorer/src/datatypes.rs
@@ -332,7 +332,7 @@ pub fn microseconds_six_digits(microseconds: u32) -> u32 {
     }
 }
 
-pub fn ex_naive_datetime_to_timestamp_res(
+pub fn ex_naive_datetime_to_timestamp(
     ex_naive_datetime: ExNaiveDateTime,
     time_unit: TimeUnit,
 ) -> Result<i64, ExplorerError> {
@@ -417,7 +417,7 @@ pub struct ExDateTime<'a> {
     pub zone_abbr: &'a str,
 }
 
-pub fn ex_datetime_to_timestamp_res(
+pub fn ex_datetime_to_timestamp(
     ex_datetime: ExDateTime,
     time_unit: TimeUnit,
 ) -> Result<i64, ExplorerError> {
--- a/native/explorer/src/datatypes.rs
+++ b/native/explorer/src/datatypes.rs
@@ -19,6 +19,15 @@ use polars::prelude::cloud::AmazonS3ConfigKey as S3Key;
 
 use chrono_tz::{OffsetComponents, OffsetName, Tz};
 
+pub use polars::export::arrow::datatypes::TimeUnit as ArrowTimeUnit;
+
+pub use polars::export::arrow::temporal_conversions::{
+    date32_to_date as days_to_date, timestamp_ms_to_datetime as timestamp_ms_to_naive_datetime,
+    timestamp_ns_to_datetime as timestamp_ns_to_naive_datetime,
+    timestamp_to_datetime as arrow_timestamp_to_datetime,
+    timestamp_us_to_datetime as timestamp_us_to_naive_datetime,
+};
+
 pub use ex_dtypes::*;
 
 pub struct ExDataFrameRef(pub DataFrame);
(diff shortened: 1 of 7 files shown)


native/explorer/src/datatypes.rs
```diff
@@ -407,28 +417,21 @@ pub struct ExDateTime<'a> {
...
-                    .unwrap(),
-            );
-
-        match duration.num_microseconds() {
-            Some(us) => us,
-            None => duration.num_milliseconds() * 1_000,
+pub fn ex_datetime_to_timestamp_res(
+    ex_datetime: ExDateTime,
+    time_unit: TimeUnit,
+) -> Result<i64, ExplorerError> {
```
billylanchantin, line 423: 1. I'm returning a `Result` here so we get better error messages elsewhere.
2. Is adding a `_res` suffix to a function that return `Result` weird? Do people do that?

native/explorer/src/encoding.rs
```diff
@@ -248,7 +234,7 @@ macro_rules! unsafe_encode_datetime {
                         $calendar_iso_module,
                         dt_tz.day().encode($env).as_c_arg(),
                         dt_tz.hour().encode($env).as_c_arg(),
-                        (limited_ms, 6).encode($env).as_c_arg(),
+                        (limited_microseconds, 6).encode($env).as_c_arg(),
```
billylanchantin, line 237: Since we now have `time_unit` in scope, we could be choosing between:

* `(limited_milliseconds, 3)`
* `(limited_microseconds, 6)`

if we wanted to. Should we?

billylanchantin:
Turns out this branch fixes more than I thought. `main` is not encoding `:nanosecond` datetimes correctly:

```bash
MIX_ENV=test iex -S mix # test env to get access to non-UTC timezones
```

```elixir
iex> dt = %DateTime{year: 2017, month: 11, day: 7, zone_abbr: "CET",
...>                hour: 11, minute: 45, second: 18, microsecond: {123456, 6},
...>                utc_offset: 3600, std_offset: 0, time_zone: "Europe/Paris"}
#DateTime<2017-11-07 11:45:18.123456+01:00 CET Europe/Paris>

# This branch (correct)
iex> [dt] |> Explorer.Series.from_list(dtype: {:datetime, :nanosecond, dt.time_zone})
#Explorer.Series<
  Polars[1]
  datetime[ns, Europe/Paris] [2017-11-07 11:45:18.123456+01:00 CET Europe/Paris]
>

# main (way off)
iex> [dt] |> Explorer.Series.from_list(dtype: {:datetime, :nanosecond, dt.time_zone})
#Explorer.Series<
  Polars[1]
  datetime[ns, Europe/Paris] [1970-01-18 11:27:35.118123+01:00 CET Europe/Paris]
>
```

native/explorer/src/datatypes.rs
billylanchantin: Reverted the `_res` thing: [e9cac12](https://github.com/elixir-explorer/explorer/pull/1020/commits/e9cac12ea984ca7027c7d6e7c505a54dc75f8688)

native/explorer/src/encoding.rs
josevalim: We probably should, the more precision, the better?

josevalim (approved):

native/explorer/src/encoding.rs
billylanchantin: Yeah. I think I'm gonna do it on a follow-on PR since it breaks a test.

billylanchantin: https://github.com/elixir-explorer/explorer/pull/1024

Issue #885, opened 2024-03-21
`{:datetime, :second}` dtype support


LostKobrakai:
Looking at the initial PR for the precision on datetimes I didn't find seconds as a unit being discussed. I'm dealing with a lit of unix timestamps and doing `{:datetime, :millisecond}` feels wrong given I initially don't have that much precision. Are there any reasons not to support second precision?

josevalim:
If it is supported in polars, I believe we can totally support it here too.

LostKobrakai:
Ah, that's where the limitation is. It only supports the 3 precisions we have it seems. I was looking at the arrow docs, which were linked on #867 and it showed second as a unit.

PR #1024, opened 2024-11-21
Encode millisecond precision

billylanchantin:
Follow-up to: https://github.com/elixir-explorer/explorer/pull/1020.

Before we were always encoding microseconds like:

* `microsecond: {us, 6}`

This PR changes that for when we have millisecond precision to:

* `microsecond: {us / 1_000 * 1_000, 3}`

It's still 6 digits, but rounded down to the nearest 1,000.

### DIFF
--- a/native/explorer/src/datatypes.rs
+++ b/native/explorer/src/datatypes.rs
@@ -321,14 +321,55 @@ fn convert_to_arrow_time_unit(time_unit: TimeUnit) -> ArrowTimeUnit {
     }
 }
 
-// Limit the number of digits in the microsecond part of a timestamp to 6.
-// This is necessary because the microsecond part of Elixir is only 6 digits.
-#[inline]
-pub fn microseconds_six_digits(microseconds: u32) -> u32 {
-    if microseconds > 999_999 {
-        999_999
-    } else {
-        microseconds
+/// Trait for the fractional part of seconds for Elixir's time-related structs.
+pub trait ExMicrosecondTuple {
+    /// Get the sub-second part.
+    fn subsec_micros(&self) -> u32;
+
+    /// Round the sub-second part to a max of 6 digits.
+    fn subsec_micros_limited(&self) -> u32 {
+        // In the event of a leap second, `subsec_micros()` may exceed 999,999.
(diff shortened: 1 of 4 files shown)


native/explorer/src/datatypes.rs
```diff
@@ -321,14 +321,63 @@ fn convert_to_arrow_time_unit(time_unit: TimeUnit) -> ArrowTimeUnit {
...
+        self.nanosecond() / 1_000
+    }
+    // fn microsecond_tuple(&self) -> (u32, u32) {
+    //     let us = self.subsec_micros_limited();
+    //     if us == 0 {
+    //         (0, 0)
+    //     } else {
+    //         (self.subsec_micros_limited(), 6)
+    //     }
+    // }
```
billylanchantin, line 363: There's currently a test that fails with:

```elixir
       iex> s = Explorer.Series.from_list([~T[01:55:00], ~T[15:35:00], ~T[23:00:00]])
       iex> Explorer.Series.quantile(s, 0.5)
       ~T[15:35:00]
     code:  Explorer.Series.quantile(s, 0.5) === ~T[15:35:00]
     left:  ~T[15:35:00.000000]
     right: ~T[15:35:00]
```

The commented code would fix this test. However, I think we should change the test instead.

The Rust-side can't tell the difference between a `chrono::NaiveTime` that was originally encoded from an Elixir `%Time{}` with `:second`-precision and one with higher precision that happens to have 0 microseconds. I think the best we can do is return what we know for sure.

Thoughts?

josevalim: We can change the test. All of the times given in the test have precision of 0, so ideally we would return precision 0, but if we have to give it higher precision, any of the precisions are fine.

billylanchantin: > ideally we would return precision 0

Right. Unfortunately there's no way to annotate this now because of the lack of `:second` precision in Polars: https://github.com/elixir-explorer/explorer/issues/885. Also we only support `:time` and not `{:time, :millisecond | :microsecond | :nanosecond}` (though we could if we wanted).

Issue #1011, opened 2024-10-29
NIF panic with list of struct


maennchen:
## Code

```elixir
Mix.install([{:explorer, "~> 0.10.0"}])

name_dtype = {"names",
{:list,
 {:struct,
  [
    {"language", :string},
    {"name", :string},
    {"transliteration", :category},
    {"type", :category}
  ]}}}

[
  %{names: []},
  %{names: [%{name: "CABK", type: "acronym", language: nil, transliteration: "none"}]}
]
|> Explorer.DataFrame.new(dtypes: [name_dtype])
|> dbg
```

## Expected

A working Dataframe

## Actual

```
thread '<unnamed>' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-arrow-0.43.1/src/array/binview/mod.rs:327:9:
assertion failed: i < self.len()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-arrow-0.43.1/src/array/binview/mod.rs:327:9:
assertion failed: i < self.len()
#Inspect.Error<
  got ErlangError with message:

      """
      Erlang error: :nif_panicked
      """

  while inspecting:

      %{
        data: %Explorer.PolarsBackend.DataFrame{
          resource: #Reference<0.3840168779.3188326401.44015>
        },
        remote: nil,
        names: ["names"],
        __struct__: Explorer.DataFrame,
        dtypes: %{
          "names" => {:list,
           {:struct,
            [
              {"language", :string},
              {"name", :string},
              {"transliteration", :category},
              {"type", :category}
            ]}}
        },
        groups: []
      }

  Stacktrace:

    (explorer 0.10.0) Explorer.PolarsBackend.Native.s_to_list(#Explorer.PolarsBackend.Series<
  #Reference<0.3840168779.3188326416.41949>
>)
    (explorer 0.10.0) lib/explorer/polars_backend/shared.ex:24: Explorer.PolarsBackend.Shared.apply_series/3
    (explorer 0.10.0) lib/explorer/backend/data_frame.ex:324: anonymous fn/3 in Explorer.Backend.DataFrame.build_cols_algebra/3
    (elixir 1.17.3) lib/enum.ex:1703: Enum."-map/2-lists^map/1-1-"/2
[...]

PR #1028, opened 2024-11-28
Unskip DataFrame serialization properties

billylanchantin:
Part of: #1014 

This removes all the `@tag :skip`s from the properties. If we `exclude: :category` to work around #1011, everything passes.

Note: #1014 is still a WIP because there may be other issues with `:category` dtypes, but we won't find them until #1011 is resolved.

### DIFF
--- a/test/explorer/data_frame_test.exs
+++ b/test/explorer/data_frame_test.exs
@@ -4729,8 +4729,7 @@ defmodule Explorer.DataFrameTest do
 
   # These property tests are a work in progress. They currently aim to cover
   # creation and serialization (including printing). Serialization in particular
-  # is causing lots of panics. The plan is to keep the properties that don't
-  # pass but with `@tag :skip` until we can fix them.
+  # is causing lots of panics. Issue #1011 currently blocks a few.
   #
   # Notes:
   #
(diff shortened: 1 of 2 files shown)


test/explorer/data_frame_test.exs
```diff
@@ -4804,10 +4807,12 @@ defmodule Explorer.DataFrameTest do
...
 
-    @tag :skip
-    property "can dump any DataFrame to NDJSON" do
+    property "can dump any non-binary, non-time DataFrame to NDJSON" do
+      not_yet_implemented = [:binary, :time]
+
       check all(
-              dtypes <- Explorer.Generator.dtypes(),
+              # TODO: Remove `exclude: :category` after #1011 is resolved.
+              dtypes <- Explorer.Generator.dtypes(exclude: [:category | not_yet_implemented]),
```
billylanchantin, line 4815: If we don't exclude e.g. `:time`, we get:

```
thread '<unnamed>' panicked at path/to/.cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-json-0.43.1/src/json/write/serialize.rs:517:18:
not yet implemented: Writing Time64(Nanosecond) to JSON
```

It may be better to check for the not-yet-implemented dtypes in Elixir and raise an `ArgumentError` if we find them. We could do this either in `Explorer.DataFrame` or `Explorer.PolarsBackend.DataFrame`.

Thoughts?

josevalim: It needs to be done in the backend I think because the decision is backend specific. I can see some of them happy in converting Time64 to an integer or a string to send as JSON.

billylanchantin: Yeah I think so.

I'd love to document it too, otherwise it'll be a bit of a surprise. I'll see if I can find a good place for it.

PR #1030, opened 2024-11-29
Make region truly optional

szajbus:
Omitting `region` option causes an exception down the line because it is expected to be a binary in https://github.com/aws-beam/aws_signature/blob/0429920baf6477a76378a8569238a3186d95a9db/src/aws_signature.erl#L77

With this change, unset region is converted to an empty string in the same way as body.


philss:
@szajbus thank you!

PR #1031, opened 2024-11-29
Update Polars to v0.44

philss:
## Changes introduced in this version

- `Series.json_decode/2` cannot decode date time strings. I believe this is an unintentional regression from Polars.
- `Series.json_decode/2` is going to raise error in case it cannot decode a single entry - there is an issue about this upstream: https://github.com/pola-rs/polars/issues/19847
- `Series.covariance/2` may return `nil` in case covariance cannot be computed.
- minor float precision differences.


test/explorer/series_test.exs
```diff
@@ -6578,9 +6578,14 @@ defmodule Explorer.SeriesTest do
 
     test "extracts primitive from json and nil for mismatch" do
```
josevalim, line 6579: should we change the name of this test?

josevalim (approved):

test/explorer/series_test.exs
philss: done: [065776f](https://github.com/elixir-explorer/explorer/pull/1031/commits/065776ff4a7664a0aa94d67bca83eab88074635f)

philss:
I will go ahead with this and tackle the regressions in another moment :)

Issue #1018, opened 2024-11-15
mismatched dtypes with decimal and summarise


dvic:
This 

```elixir
require Explorer.DataFrame

Explorer.DataFrame.new([
  {"field_a", Explorer.Series.from_list([], dtype: :s64)},
  {"field_b", Explorer.Series.from_list([], dtype: :s64)},
  {"field_c", Explorer.Series.from_list([], dtype: {:decimal, 18, 2})},
])
|> Explorer.DataFrame.summarise(summy: sum(field_a * field_b) * field_c)
```

gives

```
(ArgumentError) cannot invoke Explorer.Series.multiply/2 with mismatched dtypes: {:s, 64} and :null
    (explorer 0.10.0) lib/explorer/series.ex:6814: Explorer.Series.dtype_mismatch_error/3
```

But this works

```elixir
require Explorer.DataFrame

Explorer.DataFrame.new([
  {"field_a", Explorer.Series.from_list([], dtype: :s64)},
  {"field_b", Explorer.Series.from_list([], dtype: :s64)},
  {"field_c", Explorer.Series.from_list([], dtype: {:decimal, 18, 2})},
])
|> Explorer.DataFrame.summarise(summy: sum(field_a * field_b) * cast(field_c, :f64))
```

Is this expected? decimals are different from floats in this regard?

billylanchantin:
Thanks for the report!

No this is not expected. Both should work.

Curiously, we just fixed a bug where empty lists ended up with the `:null` dtype instead of the intended `:decimal` dtype. That's what I expected to be going on here. But on `main`, now _both_ of your examples fail like this:

```
** (RuntimeError) DataFrame mismatch.

expected:

    names: ["summy"]
    dtypes: %{"summy" => {:f, 64}}

got:

    names: ["summy"]
    dtypes: %{"summy" => {:list, :null}}

    (explorer 0.11.0-dev) lib/explorer/polars_backend/shared.ex:61: Explorer.PolarsBackend.Shared.apply_dataframe/4
    (explorer 0.11.0-dev) lib/explorer/polars_backend/data_frame.ex:895: Explorer.PolarsBackend.DataFrame.summarise_with/3
    iex:4: (file)
```

Not sure what's going on there yet. Will have to dig deeper.

dvic:
[...]

PR #1033, opened 2024-11-30
Document/test unstable decimal dtype

billylanchantin:
Discussed here: https://github.com/elixir-explorer/explorer/issues/1018#issuecomment-2508583403

Adds a test module and a warning to the moduledoc:

Screen Shot 2024-11-30 at 11 29 17 AM

### DIFF
--- a/test/explorer/polars_backend/decimal_unstable_test.exs
+++ b/test/explorer/polars_backend/decimal_unstable_test.exs
@@ -1,0 +1,65 @@
+defmodule Explorer.PolarsBackend.DecimalUnstableTest do
+  # This module tracks some oddities we've found with decimal dtypes. The Polars
+  # docs warn that decimals are unstable:
+  #
+  #   > This functionality is considered unstable. It is a work-in-progress
+  #   > feature and may not always work as expected. It may be changed at any
+  #   > point without it being considered a breaking change.
+  #
+  # https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Decimal.html
+  #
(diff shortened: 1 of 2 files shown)


test/explorer/polars_backend/decimal_unstable_test.exs
```diff
@@ -0,0 +1,65 @@
...
+          dtypes: %{"a" => {:f, 64}}
+
+      got:
+
+          names: ["a"]
+          dtypes: %{"a" => {:decimal, 38, 1}}
+      """,
+      fn -> DF.summarise(df, a: mean(a)) end
+    )
+  end
```
billylanchantin, line 43: This is actually a result of our expectations:

https://github.com/elixir-explorer/explorer/blob/020fb23e6db145a640b3f1fcb1fe95f404afb8ee/lib/explorer/backend/lazy_series.ex#L772-L783

I'm not entirely sure it's correct.

billylanchantin: @josevalim Sorry forgot about this. Are you saying we should remove this test and force our decimal operations to return float instead?

josevalim: I think we should return whatever Polars return for now and we will fix it naturally as Polars fixes it.

billylanchantin: Ok, then the test can stay. When Polars fixes their end, we will learn about it because the test will break. Then we can convert it to a test of what the behavior should actually be.

josevalim (approved):

Issue #1032, opened 2024-11-30
Estimating DataFrame memory use?


thbar:
I looked into the code of Explorer & Polars, and wondered if somehow we could expose or estimate the theoretical memory use of a given `DataFrame`, even with a bit of external estimate/computation.

For the context: I am building a "consolidation" of > 1.5k CSV files, a number that will grow steadily in the next few years (containing data for electrical vehicle charging stations of France).

I have decided to use Explorer for that use-case, for the following reasons:
- first to get a higher-level abstraction to manipulate tabular data (compared to list of maps)
- and as a side-effect, for reduced memory consumption (dates/boolean/integer/floats are stored on much fewer bytes, and strings can see considerably reduced memory use via [enums/categories](https://github.com/elixir-explorer/explorer/blob/020fb23e6db145a640b3f1fcb1fe95f404afb8ee/README.md?plain=1#L16), something that will reduce memory use by many MB in my case).

Happy to contribute my findings too, but if someone has pointers on how to estimate theoretical memory use, please let me know!

billylanchantin:
Hi @thbar!

It looks like Polars has an `estimated_size` method available for DataFrames:

* https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.estimated_size.html

That seems like what you need. I think we'd accept a PR to add that functionality.

thbar:
Hi @billylanchantin! Thanks for the pointer; I went ahead and implemented something that appears to work:
- https://github.com/elixir-explorer/explorer/pull/1034

thbar:
Closing, I will try that out on the field. Thanks for the quick review @billylanchantin!

halian-vilela:
Hey @billylanchantin, I'm commenting here because I think a have a much related issue...

Seeking to avoid unnecessary memory allocation, I would like to be able to read some metadata about a given Parquet **before** actually loading it into memory. I've not found any Explorer function that seems to be able to do that...

Do we have something? Actually, what I really need is the number of lines/rows (and I'm pretty sure that it's available as metadata once I have the parquet available).

Thanks!

thbar:
In a similar fashion, a useful optimisation will be (if anyone already knows how to do that) to only keep the columns we have interest for (or even achieve row-filtering before adding it to the in-memory frame). I am not sure if the "lazy" mode allows that or not.

billylanchantin:
[...]

PR #1034, opened 2024-11-30
Add method to estimate DataFrame memory use

thbar:
See:
- #1032 

This adds `Explorer.DataFrame.estimated_sized/1`, which leverages https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.estimated_size.html (thanks @billylanchantin for the pointer) to "return an estimation of the total (heap) allocated size of the DataFrame".


billylanchantin (approved):
Perfect, thanks! This seems like a handy function :)

PR #1044, opened 2024-12-28
Update Polars to v0.45

philss:
It also updates `object_store` to v0.11 and `thiserror` to v2.0.

The only important change was the deprecation of the `ddof` option for [`correlation/3`](https://hexdocs.pm/explorer/Explorer.Series.html#correlation/3) function.
A minor change is in that `window_standard_deviation` now propagates `nil` values.

### DIFF
--- a/lib/explorer/query.ex
+++ b/lib/explorer/query.ex
@@ -320,7 +320,7 @@ defmodule Explorer.Query do
     then: 2
   ]
 
-  @kernel_only kernel_only -- (kernel_only -- kernel_all)
+  @kernel_only kernel_only -- kernel_only -- kernel_all
 
   @doc """
   Returns a "query-backed" `Explorer.DataFrame` for use in queries.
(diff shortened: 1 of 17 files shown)


lib/explorer/query.ex
```diff
@@ -320,7 +320,7 @@ defmodule Explorer.Query do
     then: 2
   ]
 
-  @kernel_only kernel_only -- kernel_only -- kernel_all
+  @kernel_only kernel_only -- (kernel_only -- kernel_all)
```
philss, line 323: Going to revert this only for the CI. We can update back when Elixir 1.18 is adopted as default version.

```suggestion
  @kernel_only kernel_only -- kernel_only -- kernel_all
```

josevalim (approved):

MurdoqZodoq (commented):
Kuluwm

Issue #1045, opened 2025-01-01
Add Radian/Degree Conversion


sasikumar87:
It might be handy to have built-in Series functions for converting between radians and degrees.  Polars has `radians` and `degrees` for this purpose .  

Reg naming, should we choose polars like `Series.radians/1` / `Series.degrees/1` or `Series.to_radians/1` / `Series.to_degrees/1` to be more expressive?

josevalim:
I would say `radians`/`degrees`, as there is no conversion of types. PRs welcome!

PR #1046, opened 2025-01-01
Add radian/degree conversion functions

sasikumar87:
Closes https://github.com/elixir-explorer/explorer/issues/1045

### DIFF
--- a/lib/explorer/series.ex
+++ b/lib/explorer/series.ex
@@ -4068,6 +4068,58 @@ defmodule Explorer.Series do
   def atan(%Series{dtype: dtype}),
     do: dtype_error("atan/1", dtype, [{:f, 32}, {:f, 64}])
 
+  @doc """
+  Converts given radians to degrees.
+  The resultant series is going to be of dtype `{:f, 64}`.
+
+  ## Supported dtype
+
+    * `{:f, 32}`
+    * `{:f, 64}`
+
+  ## Examples
+
+      iex> s = [-12.566370614359172, -9.42477796076938, -6.283185307179586, -3.141592653589793, 0.0, 3.141592653589793, 6.283185307179586, 9.42477796076938, 12.566370614359172] |> Explorer.Series.from_list()
+      iex> Explorer.Series.degrees(s)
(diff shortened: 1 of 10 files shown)


josevalim (approved):

lib/explorer/series.ex
```diff
@@ -4068,6 +4068,58 @@ defmodule Explorer.Series do
...
+    do: dtype_error("degrees/1", dtype, [{:f, 32}, {:f, 64}])
+
+  @doc """
+  Converts given degrees to radians.
+  The resultant series is going to be of dtype `{:f, 64}`.
+
+  ## Supported dtype
+
+    * `{:f, 32}`
+    * `{:f, 64}`
```
billylanchantin, line 4104:
```elixir
iex> s = [-720.0, -540.0, -360.0, -180.0, 0.0, 180.0, 360.0, 540.0, 720.0] |> Explorer.Series.from_list(dtype: :f32)
#Explorer.Series<
  Polars[9]
  f32 [-720.0, -540.0, -360.0, -180.0, 0.0, 180.0, 360.0, 540.0, 720.0]
>

iex> Explorer.Series.radians(s)
** (RuntimeError) Polars Error: invalid series dtype: expected `Float64`, got `f32`
    (explorer 0.11.0-dev) lib/explorer/polars_backend/shared.ex:27: Explorer.PolarsBackend.Shared.apply_series/3
    iex:11: (file)
```

billylanchantin (changes requested):
Thanks @sasikumar87! Unfortunately I'm getting an error for `dtype: :f32`. I think we either want to restrict these functions to only `:f64` or do some hand-holding on the Rust-side types.

sasikumar87:
Thanks for the catch @billylanchantin . Found the same issue with other trigonometric functions like sin, cos, etc . How do you think we should proceed next?

billylanchantin:
@sasikumar87 _Ah_ ok this is a wider issue. I say we merge this PR then make a dedicated issue to track the `f32` problem.

Issue #1047, opened 2025-01-04
Trig functions raise on `f32` dtypes


billylanchantin:
Example:

```elixir
iex> import Explorer.Series
iex> pi = :math.pi
iex> [pi/2, pi, 3*pi/2] |> from_list(dtype: :f32) |> sin
```

Raises with:

```
** (RuntimeError) Polars Error: invalid series dtype: expected `Float64`, got `f32`
    (explorer 0.11.0-dev) lib/explorer/polars_backend/shared.ex:27: Explorer.PolarsBackend.Shared.apply_series/3
    iex:3: (file)
```

Although we document that this should work:

<img width="849" alt="Screen Shot 2025-01-04 at 10 21 25 AM" src="https://github.com/user-attachments/assets/a25f4bea-74a8-4fb0-9c72-ef2ae686dd9e" />

Found here by @sasikumar87:

* https://github.com/elixir-explorer/explorer/pull/1046

PR #1051, opened 2025-01-09
Fix trigonometric functions to not raise on f32 inputs

sasikumar87:
Fixes https://github.com/elixir-explorer/explorer/issues/1047

### DIFF
--- a/native/explorer/src/series.rs
+++ b/native/explorer/src/series.rs
@@ -1588,50 +1588,178 @@ pub fn s_clip_float(s: ExSeries, min: f64, max: f64) -> Result<ExSeries, Explore
 
 #[rustler::nif(schedule = "DirtyCpu")]
 pub fn s_sin(s: ExSeries) -> Result<ExSeries, ExplorerError> {
-    let s1 = s.f64()?.apply_values(|o| o.sin()).into();
-    Ok(ExSeries::new(s1))
+    match s.dtype() {
+        DataType::Float64 => {
+            let s1 = s.f64()?.apply_values(|o| o.sin()).into();
+            Ok(ExSeries::new(s1))
+        }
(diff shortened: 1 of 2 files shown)


billylanchantin (approved):
Approved pending the precision issues with the `tan` test.

It looked like they were off by a single decimal place 🙄. Let us know if they persist.

josevalim:
Precision issues will happen on different machines, OSes and architectures. So one option is to compare that they are within certain delta.

billylanchantin:
@josevalim Do we have a helper somewhere that acts like `Nx.all_close`? That'd be handy here.

josevalim:
We don't but it should be easy to implement. Subtract two series, compute the absolute, see if they are within delta and return true if all true.

billylanchantin:
Agreed. It'd be generally helpful too.

```elixir
pi = :math.pi()
radians_list = [-2 * pi, -pi, -pi / 2, 0, pi / 2, pi, 2 * pi]
degrees_list = [-360.0, -180.0, -90.0, 0.0, 90.0, 180.0, 360.0]

radians = Series.from_list(radians_list, dtype: :f32)
degrees = Series.from_list(degrees_list, dtype: f32)

result = Series.degrees(radians)
assert all_close?(result, degrees)
assert Series.dtype(result) == {:f, 64}

# ...

defp all_close?(%Series{} = a, %Series{} = b, tol \\ 1.0e-8) do
  Series.subtract(a, b)
  |> Series.abs()
  |> Series.less_equal(tol)
  |> Series.all?()
end
```

sasikumar87:
@billylanchantin fixed the test.

billylanchantin:
Thanks, @sasikumar87!

Issue #1023, opened 2024-11-20
Provide more details in error message around `concat_rows`


LostKobrakai:
Given the following dataframes:

```
[
  #Explorer.DataFrame<
    Polars[1134 x 4]
    gtin string […]
    series string […]
    program string […]
    program_color string […]
  >,
  #Explorer.DataFrame<
    Polars[1520 x 4]
    gtin string […]
    series string […]
    program null […]
    series_color string […]
  >
]
```

I got

```
[error] ** (ArgumentError) dataframes must have the same columns
    (explorer 0.10.0) lib/explorer/data_frame.ex:5436: anonymous fn/3 in Explorer.DataFrame.compute_changed_types_concat_rows/1
```

This lead me to believe that the null vs string column type to be the issue while it was the different *_color columns.

The error message could be better and `concat_rows` docs could call out that typecasting works between null and other column types

billylanchantin:
Small clarification (we chatted on slack):

The error message could be improved by calling out which columns specifically didn't match. Something like:

```
** (ArgumentError) dataframes must have the same columns

  * Left DataFrame has these columns not present in the right DataFrame:

      ["program_color"]

  * Right DataFrame has these columns not present in the left DataFrame:

      ["series_color"]

    (explorer 0.10.0) lib/explorer/data_frame.ex:5436: anonymous fn/3 in Explorer.DataFrame.compute_changed_types_concat_rows/1
```

where internally we'd do something like:

```elixir
left_cols = left_df |> names() |> MapSet.new()
right_cols = right_df |> names() |> MapSet.new()

mismatched_cols = MapSet.symmetric_difference(left_cols, right_cols)

in_left_only = left_cols |> MapSet.intersection(mismatched_cols) |> Enum.to_list()
in_right_only = right_cols |> MapSet.intersection(mismatched_cols) |> Enum.to_list()
```

viniciussbs:
> This lead me to believe that the null vs string column type to be the issue while it was the different *_color columns.

Just as a feedback, I had the same issue a few days ago and I came to the same wrong conclusion around null vs string. It was just singular vs plural column name, though. 😅

PR #1057, opened 2025-01-18
Better `concat_rows` error messages

billylanchantin:
Fixes: https://github.com/elixir-explorer/explorer/issues/1023

Example:

```elixir
df0 = DF.new([x: [1, 2, 3], y: ["g", "h", nil]])
df1 = DF.new([z: [4, 5, 6]])
DF.concat_rows(df0, df1)
```

```
dataframes must have the same columns

* dataframe 0 has these columns not present in dataframe 1:

    ["x", "y"]

* dataframe 1 has these columns not present in dataframe 0:

    ["z"]
```

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -5437,107 +5437,99 @@ defmodule Explorer.DataFrame do
   @doc type: :multi
   @spec concat_rows([DataFrame.t()]) :: DataFrame.t()
   def concat_rows([%DataFrame{} = head | _tail] = dfs) do
-    with :ok <- ensure_column_names_match(dfs),
-         {:ok, changed_types} <- compute_changed_types_concat_rows(dfs) do
-      out_df = out_df(head, head.names, Map.merge(head.dtypes, changed_types))
+    changed_types = compute_changed_types_concat_rows(dfs)
+    out_df = out_df(head, head.names, Map.merge(head.dtypes, changed_types))
 
-      dfs =
-        if changed_types == %{} do
-          dfs
-        else
-          for df <- dfs do
-            mutate_with(ungroup(df), fn ldf ->
-              for {column, target_type} <- changed_types,
-                  do: {column, Series.cast(ldf[column], target_type)}
-            end)
-          end
-        end
-
-      Shared.apply_dataframe(dfs, :concat_rows, [out_df])
-    else
-      {:error, message} -> raise ArgumentError, message
-    end
-  end
-
-  defp ensure_column_names_match([head | tail]) do
-    df_0_cols = head |> names() |> MapSet.new()
-
-    tail
-    |> Enum.with_index(1)
-    |> Enum.reduce_while(:ok, fn {df, index}, :ok ->
-      df_i_cols = df |> names() |> MapSet.new()
-
-      if MapSet.equal?(df_0_cols, df_i_cols) do
-        {:cont, :ok}
+    dfs =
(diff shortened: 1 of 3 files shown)


billylanchantin:
@josevalim I'm aiming to emulate the new type error messages in Elixir with the indentations and newlines. If you would, please review with that in mind.

We can maybe skip the ANSI colors though since that'd be a big departure from our other error messages.

lib/explorer/data_frame.ex
```diff
@@ -5437,47 +5437,106 @@ defmodule Explorer.DataFrame do
...
-    Shared.apply_dataframe(dfs, :concat_rows, [out_df])
+      Shared.apply_dataframe(dfs, :concat_rows, [out_df])
+    else
+      {:error, message} -> raise ArgumentError, message
+    end
   end
 
-  defp compute_changed_types_concat_rows([head | tail]) do
-    types = head.dtypes
+  defp ensure_column_names_match([head | tail]) do
```
josevalim, line 5462: Ideally we should do this in two passes. One is a quick check to see if the names match. This should be as quick and efficient as possible. If they don't, then we start doing all of the additional work. I think the previous version was the efficient one. Could we keep it around and then do the MapSet dance only in case of errors? :)

billylanchantin: Fair point about the big O!

> Ideally we should do this in two passes.

I actually took it from one pass to two.

Before:

```
for df in tail:                     -- O(m)
  if num_cols_df != num_cols_head:  -- O(1)
    raise
  for col in df:                    -- O(n)
    if col not in head:             -- O(log(n)) since we're looking in a map
      raise
    if col's dtype not compatible:  -- O(1)
      raise
    otherwise, continue building dtype map

Overall: O(m n log(n))
```

After:

```
for df in tail:                    -- O(m)
  if col_set_df != col_set_head:   -- O(n log(n))
    bail
for df in tail:                    -- O(m)
  for col in df:                   -- O(n)
    if col's dtype not compatible: -- O(1)
      bail
    otherwise, continue building dtype map

Overall: O(m n log(n))
```

The problem is that the columns have to be the same, but not necessarily in the same order. So the names pass has to have O(n log(n)) complexity.

I can add back in the count check though. Then we can wait to build the mapset of the column names until we're sure that they'll have the same size.

I can also go back to the one pass approach if you prefer! But I _believe_ they have the same big O.

josevalim: Sorry, I should clarify. We have two scenarios: the success scenario and the failure one. The failure one can be slow, the success one can be quick. The previous code was computing compatibility and the types at once, in the fastest way possible. This code builds map sets and compares them, which I am almost sure it will be slower, and then builds the types.

So ideally we would keep the previous code and then add your code with better reporting. Does that make sense? You can even use `throw/catch` if necessary. :)

billylanchantin: Ok, I've put the old approach back!

This is good for two reasons: 1) I fixed a bug! 2) Even if I was right about the new complexity being similar, this way we _definitely_ didn't introduce an inefficiency. So that makes it worth it.

Also, I think the time big O is the same but now I see the space big O is not: now we're only allocating memory for the MapSets in the event that there's a problem. Sorry if that's what you were saying and I misunderstood.

josevalim: No worries at all, reviewing the new approach now, thank you! :heart:

josevalim (approved):
Beautiful!

PR #1060, opened 2025-01-31
Update Polars to v0.46

philss:


philss:
It looks like a regression. I will investigate this another time. But feel free to explore this if you can.

Issue #1005, opened 2024-10-16
Accessing a DataFrame with an out of bounds range should raise


billylanchantin:
Today if a range extends past the end of a DataFrame, we filter those columns out.

```elixir
df = Explorer.DataFrame.new(a: [1, 2, 3])
df[1..1]
# #Explorer.DataFrame<
#   Polars[0 x 0]
# >
```

Per the discussion [here](https://github.com/elixir-explorer/explorer/pull/1004), we think we should raise instead.

PR #1061, opened 2025-02-02
Raise on out of bounds range access

billylanchantin:
Fixes: https://github.com/elixir-explorer/explorer/issues/1005

In the example from that issue we now get:

```elixir
iex> df[1..1]
# ** (ArgumentError) range 1..1 is out of bounds for a dataframe with 1 column(s)
#     (explorer 0.11.0-dev) lib/explorer/shared.ex:221: Explorer.Shared.to_existing_columns/3
#     (explorer 0.11.0-dev) lib/explorer/data_frame.ex:387: Explorer.DataFrame.fetch/2
#     (elixir 1.18.1) lib/access.ex:322: Access.get/3
#     iex:3: (file)
```

### DIFF
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -207,19 +207,15 @@ defmodule Explorer.Shared do
     names
   end
 
-  def to_existing_columns(%{names: names} = df, first..last//step = columns, raise?) do
+  def to_existing_columns(%{names: names} = df, first..last//_ = columns, raise?) do
     if raise? do
+      # We say a range is in bounds when both limits are valid column indices.
+      # In certain instances this will reject a valid input to `Enum.slice/2`,
+      # but there were some ambiguous cases we needed to cover. See
+      # https://github.com/elixir-explorer/explorer/pull/1061 for a discussion.
       n_cols = Explorer.DataFrame.n_columns(df)
 
-      # With `Enum.slice/2`, negative indices are counted from the end.
-      [slice_min, slice_pseudo_max] =
-        [first, last]
-        |> Enum.map(&if(&1 < 0, do: n_cols + &1, else: &1))
-        |> Enum.sort()
-
-      slice_max = slice_min + abs(step) * (Range.size(slice_min..slice_pseudo_max//abs(step)) - 1)
-
-      if slice_min < 0 or slice_max >= n_cols do
+      if max(abs(first), abs(last)) >= n_cols do
         raise ArgumentError,
               "range #{inspect(columns)} is out of bounds for a dataframe with #{n_cols} column(s)"
       end
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -210,9 +210,14 @@ defmodule Explorer.Shared do
   def to_existing_columns(%{names: names} = df, %Range{} = columns, raise?) do
     if raise? do
       n_cols = Explorer.DataFrame.n_columns(df)
-      existing_columns = 0..(n_cols - 1)
 
-      if Enum.any?(columns, &(&1 not in existing_columns)) do
+      # With `Enum.slice/2`, negative indexes are counted from the end.
+      [slice_min, slice_max] =
+        [columns.first, columns.last]
+        |> Enum.map(&if(&1 < 0, do: n_cols + &1, else: &1))
+        |> Enum.sort()
+
+      if slice_min < 0 or slice_max >= n_cols do
         raise ArgumentError,
               "range #{inspect(columns)} is out of bounds for a dataframe with #{n_cols} columns"
       end
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -207,15 +207,17 @@ defmodule Explorer.Shared do
     names
   end
 
-  def to_existing_columns(%{names: names} = df, %Range{} = columns, raise?) do
+  def to_existing_columns(%{names: names} = df, first..last//step = columns, raise?) do
     if raise? do
       n_cols = Explorer.DataFrame.n_columns(df)
 
       # With `Enum.slice/2`, negative indices are counted from the end.
-      [slice_min, slice_max] =
-        [columns.first, columns.last]
+      [slice_min, slice_pseudo_max] =
+        [first, last]
         |> Enum.map(&if(&1 < 0, do: n_cols + &1, else: &1))
         |> Enum.sort()
+
+      slice_max = slice_min + step * (Range.size(slice_min..slice_pseudo_max//step) - 1)
 
       if slice_min < 0 or slice_max >= n_cols do
         raise ArgumentError,
--- a/lib/explorer/shared.ex
+++ b/lib/explorer/shared.ex
@@ -207,7 +207,17 @@ defmodule Explorer.Shared do
     names
   end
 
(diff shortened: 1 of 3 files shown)


billylanchantin:
The bounds check on ranges was a little fiddly since `Enum.slice/2` accepts negative bounds. I think I got it but I'm not 100% sure.

lib/explorer/shared.ex
```diff
@@ -207,7 +207,24 @@ defmodule Explorer.Shared do
...
-  def to_existing_columns(%{names: names}, %Range{} = columns, _raise?) do
+  def to_existing_columns(%{names: names} = df, first..last//step = columns, raise?) do
+    if raise? do
+      n_cols = Explorer.DataFrame.n_columns(df)
+
+      # With `Enum.slice/2`, negative indices are counted from the end.
+      [slice_min, slice_pseudo_max] =
+        [first, last]
+        |> Enum.map(&if(&1 < 0, do: n_cols + &1, else: &1))
+        |> Enum.sort()
```
josevalim, line 218: Because ranges can be reverse ordered, I am not sure how much we want to rely on sort. My suggestion would be to validate first and last independently:

```elixir
for pos <- [first, last] do
  cond do
    pos >= 0 and pos < n_cols -> :ok
    pos < 0 and pos >= -n_cols -> :ok
    true -> raise "..."
  end
end
```

WDYT?

billylanchantin: `Enum.slice/2` says it doesn't support reverse ordered ranges yet. Do you want to support them here in anticipation of their future support in Elixir 2.0?

josevalim: Ah, good point. If the code above works (big if), then it should be simpler and more future proof. Worth giving it a try?

billylanchantin: Sure! Unfortunately though, I think your algorithm doesn't work for the following situation:

```elixir
df = DF.new(a: [1, 2, 3], b: ["a", "b", "c"], c: [4.0, 5.1, 6.2])

n_cols = DF.n_columns(df) #=> 3
slice = 1..10//10

assert slice.last > n_cols
assert DF.to_columns(df[slice]) == %{"b" => ["a", "b", "c"]}
```

Here `slice` is a valid subset of the columns because even though `slice.last > n_cols`, `Enum.to_list(slice) == [1]`. This only happens when `step > 1` of course.

We can still try and work out a future proof approach though.

josevalim: I think you can adjust the last for a range `start..finish//step`, the last element can be calculated using this formula:
`start + (div(finish - start, step) * step)`

billylanchantin: Note that we also need to account for negative indices like `0..-1//1`. Even though that's an empty range, passing it to `Enum.slice/2` results in the full set of columns since `-1` means "the last element" in that context.

But I think if we just use `abs(step)` in my current approach, we're set: [dc059a1](https://github.com/elixir-explorer/explorer/pull/1061/commits/dc059a1e878b78c0234485f12e99c3dfb26a2a37)

josevalim: Nice. I wish there was a way we had to implement less of the existing range machinery. :(

josevalim: Do we deal with empty ranges correctly? 10..0//1 is empty and there should not fail, even in a DF with 2 columns.

billylanchantin: > I wish there was a way we had to implement less of the existing range machinery. :(

Ha yeah I was thinking `Range` could use some helpers. I started to design some in my head before realizing I'd spend all day on that. 😅 

> Do we deal with empty ranges correctly?

Blerg. So apparently not:

```elixir
iex(4)> df[slice]
** (ArgumentError) range 10..0//1 is out of bounds for a dataframe with 3 column(s)
    (explorer 0.11.0-dev) lib/explorer/shared.ex:223: Explorer.Shared.to_existing_columns/3
    (explorer 0.11.0-dev) lib/explorer/data_frame.ex:387: Explorer.DataFrame.fetch/2
    (elixir 1.18.1) lib/access.ex:322: Access.get/3
    iex:4: (file)
iex(4)> df[..]
#Explorer.DataFrame<
  Polars[3 x 3]
  a s64 [1, 2, 3]
  b string ["a", "b", "c"]
  c f64 [4.0, 5.1, 6.2]
>
```

billylanchantin: Er, wait. Is that actually the behavior we want? Even though both `0..-1//1` and `10..0//1` are empty, `Enum.slice/2` treats them differently:

```elixir
iex(7)> Enum.slice(df.names, 10..0//1)
[]
iex(8)> Enum.slice(df.names, 0..-1//1)
["a", "b", "c"]
```

I think we actually want the current behavior?

josevalim: Negative indexes are treated as last. So it is an empty range but it gets normalized. The other one is an empty range altogether. And then there are ranges that raise when out of index.

it may be best to leave this to until we add a slice! to Elixir? 🤔

billylanchantin: > it may be best to leave this to until we add a slice! to Elixir? 🤔

I think I'd prefer to merge this if possible. The semantics of out-of-bounds are currently different between Nx and Explorer. This PR attempts to bring them in line.

> The other one is an empty range altogether. And then there are ranges that raise when out of index.

Sorry if I'm not understanding. To my eyes, `10..0//1` is both empty on its own and out of bounds for a 2-column dataframe and so should raise. Can you clarify what you think should happen with these two cases?

```elixir
df = DF.new(a: [1], b: [2]) # 2 columns

df[1..0//1] # empty, but within bounds
df[9..0//1] # empty, but out of bounds
```

I think `df[1..0//1]` should return an empty dataframe and `df[9..0//1]` should raise. The other option is that all empty ranges (post normalizing) should return an empty dataframe regardless of bounds.

I admit this is a weird case and I don't feel strongly.

josevalim: I think that, if 1..10//10 should not raise for being outbounds because 10 is not actually included, so should not 10..1//1 for the same reason?

I agree that it doesn’t really matter which, but I think for those cases it should be consistent. I think always checking first and last, regardless of step, is the simplest way to go about it?

billylanchantin: > I think that, if 1..10//10 should not raise for being outbounds because 10 is not actually included, so should not 10..1//1 for the same reason?

Ah great point. It's like whack-a-mole with these cases!

>  I think always checking first and last, regardless of step, is the simplest way to go about it?

I think you're right. I'll go ahead and do that.

The results will differ slightly from `Enum.slice/2`. But we need to do what makes sense for us.

josevalim: Probably best to think of slice as an implementation detail. I think my proposed implementation should work for checking indexes too!

billylanchantin: > Probably best to think of slice as an implementation detail.

That makes sense!

> I think my proposed implementation should work for checking indexes too!

It did! Hope it's ok that I played a little code golf with it. I realized yours was equivalent to:

```elixir
if max(abs(first), abs(last)) >= n_cols do
  raise ...
end
```

josevalim (approved):
Super clean!

PR #1065, opened 2025-02-10
Add `DF.table_string/2`

billylanchantin:
Adds `table_string/2` that formats the DataFrames for printing. This way we can include doctests which help with documentation.

It may also be useful for testing.

### DIFF
--- a/lib/explorer/data_frame.ex
+++ b/lib/explorer/data_frame.ex
@@ -6039,7 +6039,7 @@ defmodule Explorer.DataFrame do
       `:limit` rows above the `…` and half below. The top half comes from the
       head of the dataframe and the bottom from the tail. Defaults to `:split`.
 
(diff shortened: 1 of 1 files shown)


josevalim:
Do you think it is worth calling it `to_table_string` or `to_table_format`? To keep in line with conventions?

Also, we should probably document that the format may change :)

billylanchantin:
Sure! Of your options, I like `to_table_string` more since it hints at the return type.

> Also, we should probably document that the format may change :)

Can do. But clarification: do you mean since it just changed a bit? Or are we expecting more changes down the road?

josevalim (approved):

josevalim:
More changes on the road. We can basically guarantee to support options and similar, but the particular style of formatting may change.

PR #1069, opened 2025-02-26
chore: update flake.nix

mhanberg:
I believe the minimum rust version was incremented a few months ago, but
the flake inputs were not updated as well. I was getting an error about
a failure with cargo metadata and the lockfile version being 4.

Now it all works. It also updates the Erlang version 27, I assume that
is fine.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -8,11 +8,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1705299878,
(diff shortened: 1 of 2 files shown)


josevalim (approved):

billylanchantin:
Thank you!

billylanchantin:
@josevalim Is this something we should add to CI?

mhanberg:
Probably fine to let the nix users take the burden.

PR #1097, opened 2025-07-07
ci: update ubuntu and windows in precompilation matrix

mhanberg:
Ubuntu 20 is deprecated and no longer runs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,18 +34,18 @@ jobs:
         nif: ["2.15"]
         job:
           - { target: aarch64-apple-darwin, os: macos-13 }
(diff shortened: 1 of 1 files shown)


billylanchantin:
Ok the errors all look like this:

```
Error: Failed to get ID token: Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
```

This seems to happen if you don't have a certain permission set:

https://stackoverflow.com/a/74167257/5932228

But we do have that. I found a possibly relevant discussion here:

https://github.com/actions/attest-build-provenance/issues/99

Seems like the issue might be that this is coming from a fork. I think then that if I merge it, this won't happen on the merge commit.

PR #1102, opened 2025-07-14
Add 'stable' flag to group_by operation

petrkozorezov:
I set stable version by default to avoid API breaking, but python implementation uses non-stable version because it faster. I do not know if it is worth breaking the API in this case.


billylanchantin:
@petrkozorezov Thanks for giving this a whirl!

> I do not know if it is worth breaking the API in this case.

Given the potential performance improvements, I think a breaking change is worth it. We're pre-1.0 so it's not so big an issue.

---

Quick thought: I don't _love_ the extra `:stable_groups?` field directly on `%DataFrame{}`. What about something like this:

```elixir
%DataFrame{
  # ...
  groups: %{
    columns: []      # this is what dataframe.groups used to be
    stable?: boolean # this is the addition
  }
}
```

Downside to my suggestion: it's even more breaking. WDYT?

lib/explorer/data_frame.ex
```diff
@@ -5553,6 +5554,11 @@ defmodule Explorer.DataFrame do
   When the dataframe has grouping variables, operations are performed per group.
   `Explorer.DataFrame.ungroup/2` removes grouping.
 
+  ## Options
+
+    * `:stable` (`boolean`) — when `true`, preserve the order of the groups exactly as they appear in the original dataframe.
+      Groups may be reordered (for example, sorted) for performance. By default `false`.
```
josevalim, line 5560: What if we pass the stable option when aggregating instead?

josevalim: s/aggregate/summarize

petrkozorezov:
> @petrkozorezov Thanks for giving this a whirl!
> 
> > I do not know if it is worth breaking the API in this case.
> 
> Given the potential performance improvements, I think a breaking change is worth it. We're pre-1.0 so it's not so big an issue.
> 
> Quick thought: I don't _love_ the extra `:stable_groups?` field directly on `%DataFrame{}`. What about something like this:
> 
> ```elixir
> %DataFrame{
>   # ...
>   groups: %{
>     columns: []      # this is what dataframe.groups used to be
>     stable?: boolean # this is the addition
>   }
> }
> ```
> 
> Downside to my suggestion: it's even more breaking. WDYT?

I like it more too, but it will lead to much bigger diff. I can try.

lib/explorer/data_frame.ex
josevalim: Right, but they may be "unstable" in the future... this will give us the opportunity to deal with those later (otherwise doing stable when unstable is required is fine).

My thinking is that instability is a property of the dataframe? Usually if you want to keep it stable, you would keep it throughout all ops?

petrkozorezov:
The obvious candidate for the such flag is sorting, but I don't think we need to mix it up because in sorting it has a slightly different semantic.
And different grouping can be with different stable requirements.

josevalim:
If different groupings have different requirements then the solution here is not enough, as we always flatten it, right? So we would need to convert the groups into a stack...

petrkozorezov:
Yes, IMHO, stack grouping looks better, but the current grouping interface doesn't allow it.

Looks like current implementation is a good enough balance.

petrkozorezov:
> I like it more too, but it will lead to much bigger diff. I can try.

Looks not so bad.

lib/explorer/data_frame.ex
```diff
@@ -5615,14 +5620,50 @@ defmodule Explorer.DataFrame do
...
+      case Keyword.fetch(opts, :stable) do
+        :error ->
+          nil
+
+        {:ok, value} when is_boolean(value) ->
+          value
+
+        {:ok, value} ->
+          raise ArgumentError, "`:stable` must be `true` or `false`, found: #{inspect(value)}."
+      end
```
billylanchantin, line 5649: @petrkozorezov How are you feeling besides this TODO?

petrkozorezov: Forgot to remove the TODO.

billylanchantin:
Reading the comments above, I don't think there's a consensus yet. Here are the options as I see them:

1. Original: `%DF{groups: [], stable_groups?: boolean}`
2. Generic: `%DF{groups: [], stable?: boolean}`
3. Current: `%DF{groups: %{columns: [], stable?: boolean}}`
4. Aggregate: `%DF{groups: []} |> DF.summarise(..., stable?: boolean)`

I think all are workable solutions, so the decision will come down to minutia. I personally order the choices like this:

Generic < Original < Aggregate < Current

I favor Current the most because it seems like a nice balance between how Polars models things and what we already have. On reflection, I would actually favor Aggregate the most except I don't understand why Polars chose to make stability a property of grouping and not aggregation, so I'm hesitant to go against it.

So @petrkozorezov and @josevalim, how are we feeling? Can this go forward as is? Or do we favor one of the other options more (including maybe an option I missed)?

petrkozorezov:
> So @petrkozorezov and @josevalim, how are we feeling? Can this go forward as is? Or do we favor one of the other options more (including maybe an option I missed)?

For me, the current solution is the best tradeoff.
As I told earlier, variant 'Aggregate' is much more complicated due to need to add the flag to every command that can be used with grouping. I tried it first and found it unacceptable.
'Original' variant IMHO is obviously worse.
About 'Generic' I've told already before. And I tried to find any other libraries with such API and couldn't.

mhanberg:
@petrkozorezov @billylanchantin 

I pulled down main after this was merged and used it with some exist Explorer code with have, and encountered this error. And to clarify, this is without changing any of our code.

```
 ** (ArgumentError) `:stable` flag can't be changed after the first `group_by`
       (explorer 0.11.1-dev) lib/explorer/data_frame.ex:5772: Explorer.DataFrame.group_by/3
       (explorer 0.11.1-dev) lib/explorer/data_frame.ex:6577: Explorer.DataFrame.frequencies/3
       (lib_sgp 0.23.8) <redacted>.ex:38: anonymous fn/2 in <redacted>/1
       (elixir 1.16.2) lib/enum.ex:1700: Enum."-map/2-lists^map/1-1-"/2
       (lib_sgp 0.23.8) <redacted>.ex:31: <redacted>/1
       (lib_sgp 0.23.8) <redacted>.ex:16: anonymous fn/1 in <redacted>/2
       (telemetry 1.3.0) /Users/m.hanberg/src/<redacted>/deps/telemetry/src/telemetry.erl:324: :telemetry.span/3
       <redacted>:99: (test)
```

PR #1103, opened 2025-07-22
Add dump arrow schema and batchrecord

WolfDan:
This PR adds the capability to convert a DataFrame Schema into an IPC schema message or a chunked DataFrame message.

## Reasoning

While both `dump_ipc` and `dump_ipc_stream` functions serialize the DataFrame into an arrow IPC File or Stream binary, they encode both the schema and the data in a single binary. Because of that we don't have a way to only encode the schema or a chunked batch record as an IPC message, this PR adds that functionality.

This is helpful for APIs that consume this message format like BigQuery storage API https://cloud.google.com/python/docs/reference/bigquerystorage/latest/google.cloud.bigquery_storage_v1.types.AppendRowsRequest.ArrowData

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -33,6 +33,7 @@ defmodule Explorer.Backend.DataFrame do
   @type query_frame :: Explorer.Backend.QueryFrame.t()
   @type lazy_series :: Explorer.Backend.LazySeries.t()
 
+  @type compact_level :: option(:oldest | :newest)
   @type compression :: {algorithm :: option(atom()), level :: option(integer())}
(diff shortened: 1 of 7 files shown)


josevalim (approved):

WolfDan:
> Also, any theories on why the code in io.rs isn't part of the Polars API?

I'm not sure as to why, they tend to expose all functions they use for the library, this feels like a mistake from their part since the functions themselves are public, is just the module is not exported as public

billylanchantin:
@WolfDan Sorry, small gotcha! All our docs are required to have a type annotation:

```elixir
@doc type: :io
```

Adding this line under the docstring should fix the CI errors.

PR #1107, opened 2025-07-30
feat: cumulative_count

mhanberg:
Adds a binding to the `cum_sum` method in polars

### DIFF
--- a/lib/explorer/backend/lazy_series.ex
+++ b/lib/explorer/backend/lazy_series.ex
@@ -74,6 +74,7 @@ defmodule Explorer.Backend.LazySeries do
     cumulative_max: 2,
     cumulative_min: 2,
     cumulative_sum: 2,
+    cumulative_count: 2,
     cumulative_product: 2,
     window_max: 5,
     window_mean: 5,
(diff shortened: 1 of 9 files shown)


lib/explorer/backend/lazy_series.ex
```diff
@@ -524,6 +525,17 @@ defmodule Explorer.Backend.LazySeries do
...
 
+  @impl true
+  def cumulative_count(%Series{} = series, reverse) do
+    args = [lazy_series!(series), reverse]
+
+    if aggregations?(args), do: raise_agg_inside_window(:cumulative_count)
+
+    data = new(:cumulative_count, args, {:u, 32}, false)
+
+    Backend.Series.new(data, {:u, 32})
```
mhanberg, line 536: The `cum_sum` value seems to return a u32, which was causing some mischief when using the same dtype as the underlying series. So here we define the function explicitly to control the dtype, rather than creating it by looping through the module attribute that has the other window functions.

billylanchantin (approved):
Looks great! The jobs were fine after re-running.

PR #1118, opened 2025-08-17
Add Series.index_of

fizzlebert:
I wanted to use the [index_of](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.index_of.html)  in elixir so here's my implementation. My rust skills are lacking but I tried, so please let me know if anything needs fixing. I'm not super happy with the inconsistent error messages for mismatching types but if there's a better way I'm happy to change it.

### DIFF
--- a/lib/explorer/polars_backend/series.ex
+++ b/lib/explorer/polars_backend/series.ex
@@ -855,6 +855,8 @@ defmodule Explorer.PolarsBackend.Series do
     value_series =
       try do
         case {series.dtype, value} do
+          # cast value to duration of same type as series to ensure durations are correctly
+          # compared at the same precision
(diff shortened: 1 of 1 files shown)


lib/explorer/backend/series.ex
```diff
@@ -344,6 +344,8 @@ defmodule Explorer.Backend.Series do
   # Struct
   @callback field(s, String.t()) :: s
 
+  @callback index_of(s, valid_types()) :: integer()
```
fizzlebert, line 347: I just realised this should have a return type of integer() | nil not just integer().  I can make a new pr for it if you think it's necessary.

billylanchantin: No that's ok, I'll just change it. Thanks for letting us know!

PR #1121, opened 2025-09-13
Add the option to set the `quote_char` setting when opening a CSV file.

oscar345:
The current implementation of the `Explorer.DataFrame.load_csv/2` and `Explorer.DataFrame.from_csv/2` function don't allow users to pass the `:quote_char` option. The description of the option on the polar documentation is: 

Single byte character used for csv quoting, default = `"`. Set to None to turn off special handling and escaping of quotes.

I was working with a dataset where the `"` character was used as part of the value, not some kind of special case, which caused an error when loading the csv as a dataframe. 

I was not sure what name to use for the parameter in the elixir code, so I used the same name as polars.

### DIFF
--- a/lib/explorer/backend/data_frame.ex
+++ b/lib/explorer/backend/data_frame.ex
@@ -60,7 +60,8 @@ defmodule Explorer.Backend.DataFrame do
               columns :: columns_for_io(),
               infer_schema_length :: option(integer()),
               parse_dates :: boolean(),
-              eol_delimiter :: option(String.t())
+              eol_delimiter :: option(String.t()),
+              quote_char :: option(String.t())
             ) :: io_result(df)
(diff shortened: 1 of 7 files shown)


billylanchantin (changes requested):
This looks like a great addition, thanks!

One request: please include a test that captures the new behavior.

billylanchantin:
@oscar345 Yeah good catch, `quote_delimeter` may be preferable.

> Do I have to create the same tests for the Explorer.DataFrame.from_csv/2 function?

No that's OK, I think the tests already added are sufficient.

oscar345:
Okay I changed the parameter name for the option to `:quote_delimiter`

billylanchantin (approved):
❤️ Thank you! ❤️

PR #1151, opened 2026-05-15
Allow decimal ~> 3.0 (mitigates CVE)

stuartjohnpage:
### Overview

The goal of this PR is to relax the `decimal` version constraint to allow 3.x, which by default mitigates [CVE-2026-32686](https://www.cve.org/CVERecord?id=CVE-2026-32686). Briefly, this vulnerability springs from an exponent amplification DoS in `Decimal.parse`, `Decimal.cast`, and `Decimal.to_string`.

This was fixed in decimal 3.1.

I mirrored [jason's handling ](https://github.com/michalmuskala/jason/blob/master/mix.exs) of this and added this as an `or` so this wouldn't break existing users

### Compatibility

The major changes from 2.x to 3.x affected the parse, cast, and to_string functions, which aren't a problem in the library

Other lockfile changes are the minimum required for decimal 3.1.0 to resolve: adbc, jason (and added fine for adbc)


daveshah:
@stuartjohnpage looks like you and I were working on the same thing today (https://github.com/elixir-explorer/explorer/pull/1152).

akoutmos:
Hey there and thanks for the PR! I'm going to close this one in favor of https://github.com/elixir-explorer/explorer/pull/1153 as there is version 2.4 that also contains the security fix and I would prefer to have a requirement set to a 2.x version that isn't susceptible to DoS attacks. Thanks!

PR #1153, opened 2026-05-16
Updating Decimal version requirement

akoutmos:


billylanchantin (approved):

thbar (approved):
I've tested this upgrade against https://github.com/etalab/transport-site.

Things look fine here - but FYI I noticed 2 unrelated problems:
- Elixir 1.19 appears to bring in failures due to how series inspect is affected
- the upgrade to Polars 0.52 broke a bit of code

So LGTM (and thanks for the bump), will open issues for those 2 issues.

akoutmos:
@thbar I think I addressed both of those issues in https://github.com/elixir-explorer/explorer/tree/1141-cloud-integration-tests-are-broken-in-ci but I am still working through Linux tests seg faulting. That branch also upgrades to Polars 0.53. I’ll get back to work on that branch after I am back from vacation in a couple of weeks, unless you want to take a stab at it?

thbar:
> @thbar I think I addressed both of those issues in https://github.com/elixir-explorer/explorer/tree/1141-cloud-integration-tests-are-broken-in-ci but I am still working through Linux tests seg faulting. That branch also upgrades to Polars 0.53. I’ll get back to work on that branch after I am back from vacation in a couple of weeks, unless you want to take a stab at it?

I'll try to look into it before you're back (no hard promise, but possible).

One thing caught my eye: the Rust version is fairly old and out of sync with what Polars uses today:

https://github.com/elixir-explorer/explorer/blob/aea455c19acb9b7ab6d475927e10358edebec04d/native/explorer/rust-toolchain.toml#L4

Would you be OK with me bumping it to the latest non-nightly at least? I'm not familiar enough with the setup, but I wonder if that could actually fix the segfaults.

akoutmos:
Go for it! Hopefully that fixes it🤞

marcnnn:
There is not a release for the decimal update.
I use Ash, and this blocked my Ash dependency to update.

I do the {:decimal, "~> 3.0", override: true} workaround now.
How can I help to address this?

akoutmos:
Hey @marcnnn. I should be able to cut a new Explorer release by early next week so you don't have to override deps any more.

PR #1155, opened 2026-07-05
Fix warnings

greetingsfellowhumans:
Fixed a lot of compiler warnings.

Most of it comes down to functions being defined automatically inside macros, and so there were lots of duplicate definitions.

There were also several impossible clauses, matching on something that can never happen.


akoutmos:
Thanks for the fixes! They have been added via #1159.

PR #1158, opened 2026-07-17
fix typo

Devon-Olivier:


akoutmos:
Good find! It has been added via #1159!

PR #1161, opened 2026-09-15
chore: Allow Nx 1.0

georgeguimaraes:
Nx 1.0.0 is out and the optional `nx ~> 0.4` keeps it out of the resolver for anyone who has both Explorer and Nx 1.0 in a project. This widens it to `~> 0.4 or ~> 1.0`.

The Nx side of Explorer is `Nx.tensor`, `from_binary`, `to_binary` and the `Nx.LazyContainer` impl, all unchanged in 1.0. The lock moves to 1.0.0.


akoutmos (approved):
Thanks for the update!

PR #1163, opened 2026-09-17
Add DataFrame.partition_by/2

georgeguimaraes:
One dataframe per distinct combination of the given columns, in the order each combination first appears, backed by Polars' `partition_by`. Every piece keeps all the columns and a lazy frame is collected first.

Fitting one model per time series in a frame, or one per store or sensor, needs exactly this split, and the alternative today is `distinct` followed by one `filter_with` per key, which reads the frame once per group.


akoutmos:
Thank for another great contribution! I should have some bandwidth to review this next week and merge it in!

